PR #196, opened 2014-03-27
removeDimensions plugin

bennyschudel:
Hey

I've added a plugin that removes the width / height attribute in case there is a viewBox attribute specified. This helps with scaling svgs to its container.

Because I do not speak russia, I've included the english translation for the README.md.ru file. Could you fix that?

Many thanks!
Ben


baloneysandwiches:
@bennyschudel Nice, thanks. But I think there needs to be a check so that nested svgs do not lose their dimension attributes.

E.g.

````
<svg width="#" height="#" viewBox="# # # #"> // remove width and height here to make responsive
    <path d="#####" />
    <svg x="25%" y="25%" width="50%" height="50%"> // don't remove width and height here
    </svg>
</svg>
```

Nested svg is well-provided for in the spec and sometimes essential:

http://www.w3.org/TR/SVG/coords.html#EstablishingANewViewport

@GreLI how to check that an `<svg>` is rootmost within the syntax tree?

Lastly, I think we should consider renaming this to something like "responsive", to make it clear that we are not removing all dimensions. I would :heart: to see this land.

GreLI:
> @GreLI how to check that an <svg> is rootmost within the syntax tree?

For now it'd be something like ``item.parentNode.elem == '#document'``. May be it would be better to add a method to API.

baloneysandwiches:
Ok, are there other situations where you only want to operate on rootmost `<svg>`? If so, +1 to adding a `isRootmost()` or whatever method to the API. Otherwise I can submit a PR along the lines above.

PR #202, opened 2014-04-06
Update cleanupIDs.js

DanielMazurkiewicz:
added "prefix" parameter to solve problem of combining multiple svg into one svg or html file

### DIFF
--- a/plugins/cleanupIDs.js
+++ b/plugins/cleanupIDs.js
@@ -6,7 +6,8 @@ exports.active = true;
 
 exports.params = {
     remove: true,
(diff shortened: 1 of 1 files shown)


NeXTs:
@DanielMazurkiewicz @GreLI 

How to specify prefix using bash console?

DanielMazurkiewicz:
@NeXTs 
Check if this works:
--config {"plugins":[{"cleanupIDs":{"remove":true,"minify":true,"prefix":"yourPrefix_"}}]}

PR #235, opened 2014-08-21
Update coa.js

terg:
Update line : 379 from undefined 'config' when -f and not config output


GreLI:
I've added some checks in a2f77efa531e174af8ce86903831e51ec8661808.

PR #263, opened 2014-11-05
Make the passes to be independent from each other

dfilatov:


dfilatov:
It's time to release a new version, isn't it? :)

GreLI:
Almost. I'm hoping to fix the most annoying bug for now with merging overlapping paths.

YetiOr:
@GreLI how about (help wanted) Label in issues ?

PR #278, opened 2014-11-26
Simplify.js plugin

cbas:
Reduce file size drastically on complex polylines. Especially useful with maps.

Supports simplifying `<polygon>`, `<polyline>` point series, as well as simplifying sequential line instructions in path syntax on `<path>`, `<glyph>`, `<animate>`, and `<animateMotion>`.

Uses the Simplify.js library. Demo: http://mourner.github.io/simplify-js/

cc svg/svgo/issues/240 & svg/svgo/issues/9
cc @cheeaun

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -40,6 +40,7 @@ plugins:
   - removeEmptyAttrs
   - removeEmptyContainers
   - mergePaths
+  - simplifyPaths
   - cleanupIDs
   - removeUnusedNS
   - transformsWithOnePath
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "coa": "~0.4.1",
     "js-yaml": "~3.2.2",
     "colors": "~1.0.3",
+    "simplify-js": "^1.2.1",
     "whet.extend": "~0.9.9"
   },
   "devDependencies": {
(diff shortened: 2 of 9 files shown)


chrissound:
Seeing as this hasn't been merged in, is there a way I can get a copy of this branch? 
I'm guessing I would then just need to install this 'package' via npm?

cbas:
@GreLI @deepsweet Any reason why this hasn't been accepted yet?

GreLI:
Just lack of time and hands. Also it'll be good to mention a license of the embedded code and verify that it is compatible with svgo's MIT license. I'm not sure that code under BSD license can be embedded in a such way. Probably, you have to at least paste BSD license into the plugin.
Btw, I'm not sure that it'll give much effect.

shinnn:
I'm also disagree about such a license violation. Why doesn't this PR use [npm package](https://www.npmjs.com/package/simplify-js)?

sindresorhus:
The merge conflict needs to be fixed.

GreLI:
Closing, since it should be done in other way.
Also, it's not that useful. It's much more effective to simplify curves. E.g. like Paper.js does.

PR #304, opened 2015-01-26
Add reference to SVGOMG

jakearchibald:
…if you think it's useful. Also interested in any changes/suggestions you might have to the app, where's the best place to chat/email?


GreLI:
Ok

Issue #272, opened 2014-11-14
Error "Invalid character entity" for any input file


keithkml:
Sample input file: https://dl.dropboxusercontent.com/u/49423508/windowpiece-18.svg

This error happens always on every file I've tried.

```
% svgo windowpiece-18.svg
Invalid character entity
```

GreLI:
As a temporary solution I recommend to turn off «Preserve Illustrator Editing Capabilities» and some other advanced options, like including XMP, in Illustrator SVG saving dialog. These options make no sense—any extra will be stripped by svgo anyway.

Issue #225, opened 2014-07-09
Cannot parse when entity is used for namespace declaration


chulkilee:
### Problem

SVG file generated by Adobe Illustrator uses references to XML entities when declaring name spaces. See the following excerpt:

```xml
<?xml version="1.0" encoding="utf-8"?>
<!-- Generator: Adobe Illustrator 16.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd" [
<!ENTITY ns_extend "http://ns.adobe.com/Extensibility/1.0/">
<!ENTITY ns_ai "http://ns.adobe.com/AdobeIllustrator/10.0/">
<!ENTITY ns_graphs "http://ns.adobe.com/Graphs/1.0/">
]>
<svg xmlns:x="&ns_extend;" xmlns:i="&ns_ai;" xmlns:graph="&ns_graphs;" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" x="0px" y="0px" width="583.029px" height="45px" viewBox="0 0 583.029 45" enable-background="new 0 0 583.029 45" xml:space="preserve">
```

**This is due to the limitation of https://github.com/isaacs/sax-js** (not a bug).

### Solutions

- Turn off strict mode in svgo: it works for above cases, but it may allow other cases passed without any error
- Make sax-js to support DTD
- Use other parser library

### Resource

See related inkscape bugs: [1093433](https://bugs.launchpad.net/inkscape/+bug/1093433) and [166371](https://bugs.launchpad.net/inkscape/+bug/166371)

PR #305, opened 2015-01-27
Disable sax-js strict mode

mhoffmeyerDC:
This is disabled by default in sax-js, and [by their own admission](https://github.com/isaacs/sax-js) makes sax-js a jerk. Having it enabled also causes bugs like [#225](https://github.com/svg/svgo/issues/225) and [#272](https://github.com/svg/svgo/issues/272) - breaking svgo entirely for SVG exported from Illustrator with certain settings. Seems more prudent to let some things 'slip by' with strict mode disabled rather than fail entirely.

### DIFF
--- a/lib/svgo/svg2js.js
+++ b/lib/svgo/svg2js.js
@@ -4,7 +4,7 @@ var SAX = require('sax'),
     JSAPI = require('./jsAPI');
 
 var config = {
-    strict: true,
+    strict: false,
     trim: true,
     normalize: true,
     lowercase: true,


aidansteele:
Why was this closed?

mhoffmeyerDC:
Disabling strict mode broke many of the tests, I think it caused some rounding errors in the SVG.

aidansteele:
Thanks for such a quick response. Is there a solution on the horizon for the issues linked in the PR?

mhoffmeyerDC:
The original issue I was having was svgo failing over an SVG with the Adobe Illustrator "preserve editing capabilities" mode enabled . It had some weird namespace convention or character that made sax-js, in strict mode, fail. I guess another solution would be to use a regex to check for that string that sax doesn't like and replace it or delete it.

aidansteele:
I would suggest that such a regex-stripping solution could be implemented as a plugin, but plugins appear to be executed after sax-js rejects the file. 

Given the "hackiness" of a such a regex solution, it still seems like it should be a plugin rather than core functionality - maybe the plugin architecture could be enhanced to support plugins running earlier in the pipeline? I'd be happy to write a plugin to fix this issue, but changing the plugin architecture might be better suited to someone more familiar with the codebase.

eitanp461:
+1
I am trying to optimize SVGs created wihj Adobe illustrator and the entity references cause sax to fail, i.e. <path style="&st0;" .../>.
Is there any workaround?

GreLI:
The workaround is to save images in Illustrator with disabled options like “Preserve editing capabilities” (they don't make sense anyway if you are passing that images to svgo). Or it's possible to replace entities with their value manually, but it's suits better if you have only one or few images.

PR #328, opened 2015-03-21
Use console instead of util to avoid deprecation warnings

silverwind:
`util.puts` and `util.error` have been deprecated in io.js and node.js 0.12 and print ugly deprecation warnings on the CLI. This commit replaces them with console equivalents.

Also there were two cases where errors weren't printed to stderr, which I changed to do so.

### DIFF
--- a/lib/svgo/coa.js
+++ b/lib/svgo/coa.js
@@ -4,7 +4,6 @@ require('colors');
 
 var FS = require('fs'),
     PATH = require('path'),
-    UTIL = require('util'),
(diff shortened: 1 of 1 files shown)


GreLI:
Thanks, but I've made such a commit, actually. Just didn't pushed yet.

silverwind:
Damn. Well I hope you get this fixed soon :)

GreLI:
Yes, the new version is coming out! Just polishing it.

PR #335, opened 2015-03-31
JSAPI.prototype.clone - fixes #311

### LINKED ISSUE #311: Api for duplicating element with all it's content
Please add the api for duplicating the element with all it's content or simply add a solution to create the svg element from JSON:

exports.fn = function(data, params) {
      return JSON.parse(JSON.stringify(data)); //not working right now
}

jakearchibald:
I made `parentNode` enumerable again, breaking `JSON.stringify(node)`. If being able to convert a node to JSON is useful, a custom `.toJSON` would fix that, but as far as I can tell it was only used for cloning.

### DIFF
--- a/lib/svgo/jsAPI.js
+++ b/lib/svgo/jsAPI.js
@@ -3,15 +3,40 @@
 var EXTEND = require('whet.extend');
 
 var JSAPI = module.exports = function(data, parentNode) {
-
     EXTEND(this, data);
-    if (parentNode) {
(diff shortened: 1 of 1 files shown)


GreLI:
LGFM, but why you made it Non-enumerable again?
> If being able to convert a node to JSON is useful

It's somewhat useful for debugging.

jakearchibald:
Made it non-enumerable again

PR #350, opened 2015-04-28
new plugin: add classes to SVG element

AprilArcus:


plugins/addClassesToSVGElement.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+exports.type = 'full';
+
+exports.active = true;
```
GreLI: Should be disabled by default.

AprilArcus: Can you give me a hint about how to use this plugin from the javascript API after I disable it by default? Specifically, I'm using it in conjunction with [svgo-loader](https://github.com/pozadi/svgo-loader) in webpack:

```JavaScript
{
    test: /icon-(.*)\.svg$/,
    loaders: [
        'ngtemplate?' + querystring.stringify({
            relativeTo: 'svg/',
        }),
        'raw',
        'svgo-loader?' + JSON.stringify({
            plugins: [
                {addClassesToSVGElement: {className: 'patreon-icon-svg'}}
            ]
        })
    ]
}
```

When I change `exports.active = true;` in `addClassesToSVGElement.js`, I can no longer use it from webpack — I assume there is some other parameter I need to pass in the configurations object to re-enable it, but I can't figure out what it is.

AprilArcus:
Is this looking ok?

GreLI:
Yes, but I'm not sure that it should be in SVGO. It's an _optimizer_, not image editor.

AprilArcus:
I totally respect your desire to avoid scope creep. I do feel feel that this plugin falls within the domain of optimization - by applying a CSS class with appropriate styles, we are able to move style information that must be repetitively specified on a per-SVG basis into a stylesheet, reducing the aggregate size of our assets.

GreLI:
Well, guess it's ok since svgo doesn't handle classes yet.

Issue #296, opened 2015-01-06
Plugin to apply the styles from a <style> block as inline on the shape


ionutzp:
Is anybody interested in this?

GreLI:
Well, we'll be always glad to review a pull request. Actually, svgo now do nothing with styles, and optimizations suffer from that. Not sure though that the plugin like this should be enabled by default.

shooftie:
I'd be keen to see this. I'm struggling with the way in which Illustrator outputs SVGs. I can either have IDs on paths and inline styles or IDs and classes on paths with a `<style>` block... But not both. 

This means that I duplicate IDs across the document when the same SVG is used or the `<style>` blocks leak and end up affecting paths in other SVGs on the page that have the same class.

It would be great if SVGO could inline the styles and then remove the `<style>` element altogether. On reading your comment on another issue, #399, this might not be possible?

strarsis:
+1

Example file:
````
<svg id="test" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
  <style>.st0{fill:red;}</style>
  <rect width="100" height="100" class="st0" style="stroke-width:3;stroke:blue" />
</svg>
````

With best regards

**Comment by Joeynoh** [comment]
+1

**Comment by strarsis** [comment]
I made a svgo plugin for this, here the PR:
https://github.com/svg/svgo/pull/447

**Comment by strarsis** [comment]
How to handle styles and classes processed by this plugin? See PR discussion.

**Comment by chrisblakley** [comment]
Adding a +1 to this. I've got two SVGs from Illustrator that have conflicting classes/styles and would love an easy way to prevent this.

**Comment by strarsis** [comment]
Plugin available now: https://github.com/svg/svgo/pull/592

**Comment by GreLI** [comment]
SVGO v1.0.0 with the plugin has been released.

**Comment by wordofchristian** [comment]
Is the usage of this plugin documented anywhere?

**Comment by GreLI** [comment]
I guess only in the plugin itself.
https://github.com/svg/svgo/blob/ceccf1ffdcdcba0f422a49685c2f60520579e46a/plugins/inlineStyles.js#L20-L38

**Comment by strarsis** [comment]
Also https://github.com/svg/svgo/pull/592#issuecomment-251956973

**Comment by wordofchristian** [comment]
Thanks for the help. For future searchers this is the usage I was looking for:
```
[...]

PR #351, opened 2015-04-29
Don't remove `styles` in removeUselessDefs.js

### LINKED ISSUE #341: BUG: removeUselessDefs plugin removes \<style\>
Of course one can assign an id as a workaround but as that is rather unusual **it would be nice if the plugin kept the \<style\> section** or at least offered an option.

Thank you!

dennari:
Fixes #341.


A1bi:
Until #296 becomes a thing, this should really be merged.

PR #368, opened 2015-05-23
Circle handling for convertShapeToPath plugin

dnewcome:
Uses two arcs technique for representing a circle as a path.
http://stackoverflow.com/questions/5737975/circle-drawing-with-svgs-arc-path/10477334#10477334

### DIFF
--- a/plugins/convertShapeToPath.js
+++ b/plugins/convertShapeToPath.js
@@ -104,11 +104,14 @@ exports.fn = function(item) {
         var cx = item.attr('cx').value;
         var cy = item.attr('cy').value;
         var r = item.attr('r').value;
-        var pathData =
-            'M' + cx + ' ' + cy +
-            'm' + (-r) + ' 0' +
+		var pathData;
+
+        if (isNaN(cx-r)) return;
+
+        pathData =
+            'M' + (cx-r) + ' ' + cy +
             'a' + r + ' ' + r + ' 0 1 0 ' + (r*2) + ' 0' +
-            'a' + r + ' ' + r + ' 0 1 0 ' + (-r*2) + ' 0z';
+            ' ' + r + ' ' + r + ' 0 1 0 ' + (-r*2) + ' 0z';
 
         item.renameElem('path');
         item.removeAttr(['cx', 'cy', 'r']);
--- a/plugins/convertShapeToPath.js
+++ b/plugins/convertShapeToPath.js
@@ -99,6 +99,24 @@ exports.fn = function(item) {
 
         item.renameElem('path')
             .removeAttr('points');
-    }
 
+    } else if (item.isElem('circle')) {
+        var cx = item.attr('cx').value;
+        var cy = item.attr('cy').value;
+        var r = item.attr('r').value;
+        var pathData =
+            'M' + cx + ' ' + cy +
+            'm' + (-r) + ' 0' +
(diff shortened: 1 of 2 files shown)


GreLI:
As you can see in tests, these paths are longer than circles, that's why I didn't added it when I wrote the plugin. So, probably it's worth only under the option like `convertCircles` (or ellipses if you'd add ellipses as well).

plugins/convertShapeToPath.js
```diff
@@ -99,6 +99,27 @@ exports.fn = function(item) {
...
             .removeAttr('points');
-    }
 
-};
\ No newline at end of file
+    } else if (item.isElem('circle')) {
+        var cx = item.attr('cx').value;
+        var cy = item.attr('cy').value;
+        var r = item.attr('r').value;
+		var pathData;
```
GreLI: Also, it would be nice if you follow the codestyle: only one var…

plugins/convertShapeToPath.js
```diff
@@ -99,6 +99,27 @@ exports.fn = function(item) {
...
+
+        if (isNaN(cx-r)) return;
+
+        pathData =
+            'M' + (cx-r) + ' ' + cy +
+            'a' + r + ' ' + r + ' 0 1 0 ' + (r*2) + ' 0' +
+            ' ' + r + ' ' + r + ' 0 1 0 ' + (-r*2) + ' 0z';
```
GreLI: …and spaces around operators: `(cx - r)`, `(r * 2)`, `(-r * 2)`.

dnewcome:
Thanks for taking a look at this PR. I will fix the code styling. However I'm not understanding where the paths differ from the circle. If I look at a subset of my test case here with exaggerated semi-transparent strokes they appear coincident. Intuitively there should be some programmatic way of handling the difference in any case.

Here is what I'm looking at:
```
 	<circle cx="100" cy="100" r="20" stroke="rgba(255,0,0,0.5)" stroke-width="10"/> 
 	<path stroke-width="10" stroke="rgba(0,0,255,0.5)" d="M80 100a20 20 0 1 0 40 0 20 20 0 1 0 -40 0z"/> 
</svg>
```

GreLI:
Paths are ok, they are just written with more bytes which is opposite to purpose of optimization.

dnewcome:
Ah, I understand what you are saying now - I had interpreted "longer" as geometrically longer. 

In my use case I normalize a set of svg files for use as icons where I can only handle pure path data. So I use svgo as part of a build process to normalize everything to paths, not necessarily to optimize 100% for size (although svgo does a very good job of getting rid of all Adobe-related junk, which I also need as part of the process).

If you don't think that converting circles lies within the domain of this plugin or svgo in general I fully understand. If you would accept it as a separate plugin I will change the pull request.

src-code:
@GreLI @dnewcome I'd be interested in seeing this functionality supported, if not in the `convertShapeToPath` plugin then some other plugin.  (It seems to make sense in this plugin however, based on the name.)  Any chance we could resurrect this somewhere?

sparr:
@GreLI as someone starting to write a program with both "svg" and "optimize" in the name and considering attempting to implement it as a svgo plugin, I want to comment here that "file size" is not the only thing that might be optimized. "more bytes which is opposite to purpose of optimization" is a very narrow interpretation.

PR #375, opened 2015-06-02
Allow providing custom plugins

rlivsey:
When experimenting with a new plugin, or just doing custom things in my app, it's a pain to have to fork SVGO and add a plugin there then keep the fork up to date. It would be great to just be able to pass a custom plugin into the plugins list.

This pull request adds that functionality, so you can do something like:

```javascript
new SVGO({
    plugins: [
        myCustomPugin: {
            type: ‘perItem’,
            fn:   function(item) {
                /* do stuff */
            }
        }
    ]
});
```

### DIFF
--- a/lib/svgo/config.js
+++ b/lib/svgo/config.js
@@ -71,23 +71,31 @@ function preparePluginsArray(plugins) {
         if (typeof item === 'object') {
 
             key = Object.keys(item)[0];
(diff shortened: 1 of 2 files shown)


GreLI:
Tests seem to be little excessive (you're actually testing how the config works in the second part), but ok.

PR #380, opened 2015-06-07
Add --show-plugins option

kt3k:
Hi,

I tried to add `--show-plugins` option to show the available plugins in cli. It would be convenient when the user forget the plugin names and he's offline.

I took the description of each plugin from `README.md` (Some are missing in README, then I took the first line of documentation in source code in those cases.)

Thanks,


GreLI:
Looks nice!

PR #382, opened 2015-06-09
Fix lint

kt3k:
Hi,

I refactored to pass jshint check. I believe nothing's changed by this change.

I also set `jshint` target as a part of travis ci process, which should notify collabolators if lint is broken.

Thanks,

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -11,6 +11,7 @@
     "curly": false,
     "camelcase": true,
     "eqeqeq": false,
+    "-W041": false,
     "forin": false,
     "immed": true,
     "multistr": true,
--- a/.jshintrc
+++ b/.jshintrc
@@ -12,6 +12,7 @@
     "camelcase": true,
     "eqeqeq": false,
     "-W041": false,
+    "-W008": false,
     "forin": false,
     "immed": true,
     "multistr": true,
@@ -21,5 +22,6 @@
     "undef": true,
     "unused": true,
     "loopfunc": true,
-    "trailing": true
+    "trailing": true,
+    "lastsemic": true
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,7 @@ branches:
   only:
     - master
 
+before_install:
+  - npm install -g jshint
+
 script: make travis
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ coveralls: lib-cov
 	@cat lcov.info | ./node_modules/.bin/coveralls
 	@rm -rf lib-cov lcov.info
 
-travis: test coveralls
+travis: jshint test coveralls
 
 jshint:
 	@jshint --show-non-errors .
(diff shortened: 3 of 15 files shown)


GreLI:
jshint is just a hint. And actually replacing `"'"` with `'\''` makes code just worse. Robots for humans not vice versa.

kt3k:
But enforcing the consistent style to every contributer is a good way to keep the quality of codebase, especially in open collaborative environment. Nobody wants to see code in broken style and discusions over style tend to be fruitless.
Anyway I reverted the quotmark modification and turned off quotmark checking in `coa.js`.

GreLI:
Well, there are other points too. Like triple equality is totally dumb (I have encountered only bugs caused by it, not double). Actually, I've [corrected](/svg/svgo/commit/5e7caa49dcfcf08b4a2f284c4f3a85a53dd709f1) some real errors in code with jshint some time ago, and remaining has been left intentionally.

kt3k:
I reverted changes of equality operators and set `-W041` option to false in  `.jshintrc` (this is necessary to stop warnings of use of `==`).
But in general `==` operator is not a good thing because it invokes implicit coercion and could be a bug prone...

http://www.codeproject.com/Articles/182416/A-Collection-of-JavaScript-Gotchas#double-eq

GreLI:
> But in general == operator is not a good thing because it invokes implicit coercion and could be a bug prone...

If you are using odd comparisons, yes it can, but nobody says you to write weird code.
“[You Don't Know JS](https://github.com/getify/You-Dont-Know-JS/blob/master/types%20&%20grammar/ch4.md#loose-equals-vs-strict-equals)” excellent describes the topic.

GreLI:
About the remaining. See, there is semicolon before closing “}” which is not necessary–it only adds noise to `if`s with array methods like `arr.some(function(){ foo; })`. Removing var in `for` loop can blow if the another `for`, where the `var` is declared, is gone. Absence of `0` before a dot can be confusing only for a very novice.

Rounding function in `convertPathData.js` is actually broken–it's intentionally replaces itself with the corresponding function in the first execution (we know precision only from passed parameter).

Probably, I'll drop `a2c` function in the future–it's been borrowed from Snap.svg, but I've made direct arcs transformation since then.

After that there is actually nothing left.

kt3k:
I turned on `lastsemic` option, which suppresses the warnings of omitted semicolon only in the one-liner functions, and turned `W008` option off, which suppresses the warnings about the omission of leading decimal 0.

> Removing var in for loop can blow if the another for, where the var is declared, is gone.

I turned off warnings of double definition of vars in convexHull. (`-W004`)

> Rounding function in convertPathData.js is actually broken–it's intentionally replaces itself with the corresponding function in the first execution (we know precision only from passed parameter).

OK, I reverted this part. jshint doesn't seem to have the option to suppress this warning, so I turned off lint for this line.

And finally I reverted a2c's change as well (and made jshint ignore it.)

jshint is not only for styles but also can detect simple errors like spelling mistakes or scope errors, and basically it reduces human efforts around writing source code, and therefore I recommend keeping .jshintrc in this repo.

PR #396, opened 2015-07-06
Allow config to specify use of self closing tags

bradbarrow:
Allow config to specify `useShortTags` (default true).

When `useShortTags` is true, empty elements are terminated with />

When `useShortTags` is false, empty elements are closed with a full closing tag
e.g. < / path >

### DIFF
--- a/lib/svgo/js2svg.js
+++ b/lib/svgo/js2svg.js
@@ -26,7 +26,8 @@ var defaults = {
     regEntities: /[&'"<>]/g,
     regValEntities: /[&"<>]/g,
     encodeEntity: encodeEntity,
-    pretty: false
+    pretty: false,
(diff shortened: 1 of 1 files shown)


GreLI:
Nice job!

bradbarrow:
Thanks @GreLI - I noticed the build has broken on this change - I was intending to fix that error this morning. Is that something you can fix easily?

GreLI:
No, it's my previos commit. Don't worry.

GreLI:
FIxed it today.

bradbarrow:
Hey @GreLI what's the plan looking like for this change?

bradbarrow:
Ah, I see a release just recently! Thanks :+1:

Issue #361, opened 2015-05-18
How to provide options to plugins via command line?


dhardtke:
Hey,

I'm trying to use the "prefix" option of the cleanupIDs plugin:
https://github.com/svg/svgo/blob/master/plugins/cleanupIDs.js#L10

Apparently it's not stated anywhere on how to provide options to plugins like this.

How do I do it using a config file?
How do I do it using the command line?

Thanks.

GreLI:
See about [config](/svg/svgo/blob/master/docs/how-it-works/en.md#1-config).

GreLI:
So you need to provide your config in .yml file with content like this:
```(yml)
plugins:
  - cleanupIDs:
      prefix: foo
```
and run svgo with it like
```
svgo --config=yourconfig.yml
```
You can also provide a config in a command line like json:
```
svgo --config='{ "plugins": [ { "cleanupIDs": { "prefix": "foo" } } ] }'
```
Note: ``cleanupIDs`` doesn't work with SVG files containing ``<style>`` or ``<script>`` since svgo can't parse them.

LukeStorry:
This seems to no longer work, has passing in config via command-line strings been deprecated?

TrySound:
Was removed 1.5 years ago

PR #412, opened 2015-08-07
Add description for removeAttrs plugin

cy6erskunk:
Add description to eliminate ugly `undefined` in `--show-plugins` output


GreLI:
Ah, forgot that part.

evenfrost:
Is it possible to pass a pattern while running this plugin in cli? Something like `svgo ... --enable removeAttrs=attr`?

evenfrost:
Found it in #361, sorry for bothering.

PR #423, opened 2015-09-02
Add setDimensions plugin

allypalanzi:
Adds a setDimensions plugin that allows users to specify a width and height for their svg.

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -51,3 +51,4 @@ plugins:
   - removeDimensions
   - removeAttrs
   - addClassesToSVGElement
+  - setDimensions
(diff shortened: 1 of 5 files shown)


GreLI:
Since the plugin needs only an outer `<svg>` it's better not to run it against every element, but use `var svg = data.content[0]` with `exports.type = 'full'` like it's done in [plugins/addClassesToSVGElement.js](/svg/svgo/blob/master/plugins/addClassesToSVGElement.js) (XML-prolog and doctype should be normally removed by corresponding plugins executed before).

Also, the value “50” of `width` and `height` seems to be too arbitrarily. It's better to set to common icon size like 16×16.

allypalanzi:
Good point! I'll make that update. 50 was totally arbitrary – the idea was that it could be configured.

htahmasebpour:
http://track.mlsend3.com/link/c/YT0yNjcyNjExMTg4NzgwNTg0MTImYz10Mmc1JmU9MTkyOCZiPTI4MzI5NzQ3JmQ9YjVwNmg2aw==.V8aWHpHHP7PwRNRaECVMwtfn-LawxA5x2n_D2scVVK8

GreLI:
The plugin is trivial, so I'm closing it until a more elaborated version will come.

PR #428, opened 2015-09-10
Use null prototype for maps

blopker:
I was looking at the cleanupID plugin to add functionality and noticed it could be cleaned up a little.

The main change is using `Object.create(null)` instead of `{}` to create the ID maps. This lets us remove the ID prefixing everywhere. The change mostly improves readability, but making less strings will result in faster garbage collection and runtime.

Other changes include cleaning up whitespace and inverting `if` statements to decrease nesting and improve readability.

### DIFF
--- a/plugins/cleanupIDs.js
+++ b/plugins/cleanupIDs.js
@@ -36,8 +36,8 @@ exports.fn = function(data, params) {
 
     var currentID,
         currentIDstring,
-        IDs = {},
-        referencesIDs = {},
+        IDs = Object.create(null),
+        referencesIDs = Object.create(null),
         idPrefix = 'id-', // prefix IDs so that values like '__proto__' don't break the work
         hasStyleOrScript = false;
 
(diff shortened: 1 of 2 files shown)


GreLI:
Yes, `Object.create(null)` is nice, but there is a problem with `id="__proto__"` that it doesn't resolve. See the [failing test](https://travis-ci.org/svg/svgo/jobs/79736146).

blopker:
Yeah, it looks like it works on newer JS engines (nodejs 0.11+, but not 0.10). Are you against using a polyfill to get es6 `Map` support in older engines? If not I'll just toss this change.

GreLI:
Not that I'm opposite to it, but svgo is a performance tool, so the less dependencies and starting time the better it would be.

blopker:
I hear you. I'll give it a try, but I'll probably just submit a PR with the other changes.

GreLI:
Yeah, thanks! You're welcome. Actually, there are plans about building svgo on the new architecture. So, the `cleanupIDs` will just clean unused ids. References will be collected in the parsing time, since they are needed for some other optimisations as well (e.g. can we transform a gradient or it's being used elsewhere). So, the plugins will be likely rewritten at some point.

Also, there are some optimizations that can be done already and are waiting for implemetation. So, minor code improvements aren't in top of priorities. Unfortunately, there is not so much time to implement all these plans. But pull requests are welcome.

blopker:
Ah, good to know about the re-architecture. I've been using SVGs and by extension svgo a lot recently. I mostly just wanted to get familiar with the codebase so I could eventually add features.

I modified the PR to add back the ID prefixes. I think the other changes are still good though and will help with refactoring (note the added test for __proto__).

Looks like CI passes now, so I'll leave it at this. Cheers! :beers:

GreLI:
What's up? Why did you closed it?

blopker:
Oh, I just don't like having PRs sitting around for too long (OCD maybe :) ) . If you'd like to merge it then cool, but if you want to wait for the rearchitecture that's cool too.

GreLI:
Yeah, merged it manually (176d0329a475b70ad8b593f3c9f8e7a62c345c14).

PR #435, opened 2015-10-06
fix TypeError("Can't get property contents of undefined") 

thiakil:
Plugin removeUnknownsAndDefaults.

If an element is encountered that doesn't exist in collections.elems, unhandled exception occurs as there is no existence check.

Encountered with this xml block (Illustrator export)
```
<metadata>
	<sfw  xmlns="&ns_sfw;">
		<slices></slices>
		<sliceSourceBounds  height="67.3" width="85.9" y="-40.8" x="-42.5" bottomLeftOrigin="true"></sliceSourceBounds>
	</sfw>
</metadata>
```

### DIFF
--- a/plugins/removeUnknownsAndDefaults.js
+++ b/plugins/removeUnknownsAndDefaults.js
@@ -66,7 +66,7 @@ for (var elem in elems) {
 exports.fn = function(item, params) {
 
     // elems w/o namespace prefix
-    if (item.isElem() && !item.prefix && elems[item.elem]) {
+    if (item.isElem() && !item.prefix) {
 
         var elem = item.elem;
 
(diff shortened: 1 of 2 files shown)


GreLI:
It would be nice to add a test in test/plugins.

thiakil:
Added test. NB: End result is that the unknown elements are NOT removed.

Should this be changed to filter elements not in the Collection?

GreLI:
I guess it should, that's what the plugin do.

thiakil:
Updated to enable checking children of elements without a content property in the collections file. If the child itself is not listed in the elems collection, we delete it.

plugins/removeUnknownsAndDefaults.js
```diff
@@ -74,13 +74,23 @@ exports.fn = function(item, params) {
...
-                    elems[elem].content.indexOf(content.elem) === -1
+                    !content.prefix && 
+                    (
+                        (
+                            elems[elem].content && // Do we have a record of its permitted content?
+                            elems[elem].content.indexOf(content.elem) === -1
+                        ) || 
+                        (
+                            !elems[elem].content && // we dont know about its permitted content
+                            !elems[content.elem] // check that we know about the element at all
```
GreLI: This check seems to be redundant. It'll be performed on this element by the plugin itself in a next iteration.

thiakil: That was the original issue; it causes an error if the element is not in elems[], as was the case with the <sfw> element. Just adding a check for existence of elems[elem] leaves the element intact.

This seems to right place due to params.unknownContent

GreLI:
Ok, thanks for the good job!

GreLI:
[v0.6.0](https://github.com/svg/svgo/releases/tag/v0.6.0) with the fix have been just released.

PR #447, opened 2015-10-28
Add localStyles plugin, tests and dependencies.

strarsis:
This PR adds the localStyles plugin which copies styles from style element to style attribute of the selected elements.

- run this plugin before the removeStyleElement plugin
- this plugin won't remove the style element, use removeStyleElement after this plugin
- run convertStyleToAttrs after this plugin
- minify styles in style element if necessary before this plugin (minified = "computed" styles)
- this plugin currently works only with class and id selectors, advanced css selectors (e.g. :nth-child) aren't currently supported
- inline styles will be written after the styles from style element
- classes are inlined by the order of their names in class element attribute

### DIFF
--- a/plugins/localStyles.js
+++ b/plugins/localStyles.js
@@ -112,6 +112,8 @@ var generateLookup = function(item) {
      lookups              = [];
 
   styleCssRules.forEach(function(styleCssRule) {
+    if(styleCssRule.type != 'rule') { return; } // skip anything nested like mediaqueries
+
     styleCssDeclarations = getCssDeclarations(styleCssRule);
     styleCssRule.selectors.forEach(function(styleSelector) {
       lookups.push({ selector: styleSelector, declarations: styleCssDeclarations });
--- a/plugins/localStyles.js
+++ b/plugins/localStyles.js
@@ -7,12 +7,13 @@ exports.active = true;
 exports.description = 'copies styles from <style> to element styles';
 
 
-var cssParser   = require('css'),
-        uniq        = require('uniq'),
-        removeValue = require('remove-value'),
-        lookupRules = [],
-        svgElem     = {},
-        styleCssAst = {};
+var cssParser             = require('css'),
+        uniq              = require('uniq'),
+        removeValue       = require('remove-value'),
+	inlineStylesParse = require('inline-styles-parse'),
+        lookupRules       = [],
+        svgElem           = {},
+        styleCssAst       = {};
 
 
 // declarations (property-value paris) from rule
@@ -75,7 +76,7 @@ var cleanupRulesAst = function(rulesAst) {
 // parses css of a css rule (no selector)
 var parseRulesCss = function(str) {
     return cssParser
-                  .parse('.dummy { ' + str + ' }')
+                  .parse(inlineStylesParse.declarationsToRule(str))
                   .stylesheet.rules[0];
 };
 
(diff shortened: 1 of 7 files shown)


Joeynoh:
+1!

GreLI:
Why there are styles both in attribute and `<style>` tag?
What about media queries?
Also, code style check fails.

strarsis:
Right, no code style check on the code yet, I ran npm test which doesn't invoke make travis, so jshint was ommited, hence the build errors, I will test with make travis now.

As mediaqueries and other nested structures cannot be added to style attribute, they would be preserved in style element in any case for further possible usage.

Initially the styles element was intended to be preserved by this plugin and removed by removeStyleElement plugin instead. However, there may be mediaqueries and orphaned selectors left in style element. What kind of modi would make sense?
  - Preserve style element completely (current mode).
  - Remove matching selectors, preserve everything else.
  - Remove all selectors except nested structures that cannot be copied to style attribute at all (e.g. mediaqueries).

What should happen to the class attributes/classe names inside that matched?
cleanupAttrs plugin afterwards?

With best regards

Joeynoh:
Just my two cents: It makes sense that this plugin does not remove the style element, considering there is already a plugin that does. You can simply run them in sequence. Optionally, this plugin can perhaps call removeStyleElement, but should be set to false by default.

Calling cleanupAttrs afterwards makes sense as well.

strarsis:
When there are mediaquery styles, the svgo end user would disable removeStyleElement anyway.
Maybe an extra plugin that just removes styles that aren't used by the classes/ids inside before this one?

strarsis:
I added style element clean up which will strip any selectors/rules that matched the classes (and which declarations were copied to style attribute).

The following rules apply now:
- Preserve nested structures (e.g. media queries) in any case (cannot be inlined).
- Remove selectors that matched.
- Remove rules where all its selectors matched.

With best regards

P.S.: Maybe later, when it is stable enough, it could be worth moving the clean up part to an extra plugin (cleanupStyles.js)?

strarsis:
Ready to merge?

GreLI:
I haven't looked into it closely yet.
Also, we're using indentation of 4 spaces mostly.

strarsis:
There seems to be more issues posted concerning styles in svgo.
- Remove styles with display: none; (hidden layers)
- Remove styles element when not needed (similar to this plugin)
- Minify styles element (and inline styles?)

To improve selector support, using a XML parser like cheerio would help 
because it offers $(...) which can evaluate more complex selectors.

plugins/localStyles.js
```diff
@@ -0,0 +1,243 @@
...
+};
+
+// generates an ast declaration per property-value pair
+var prepareCssDeclarationsAst = function(declarations) {
+    var declarationsAst = [],
+            declarationAst  = {};
+
+    declarations.forEach(function(declaration) {
+        declarationAst = { type    : 'declaration',
+                                              property: declaration.property,
```
GreLI: uh-oh

GreLI:
I'm sorry, still don't have much time to look into it. Quite a big pull-request.

plugins/localStyles.js
strarsis: I can improve/add comments if necessary. :)

strarsis:
I did a full rewrite and rebranch and will create a new PR instead.

PR #454, opened 2015-11-13
Local styles plugin

strarsis:
This PR adds the local styles plugin.
It is a complete rewrite of previous PR, to make the code simpler and more efficient.

With best regards

### DIFF
--- a/plugins/localStyles.js
+++ b/plugins/localStyles.js
@@ -326,8 +326,7 @@ exports.fn = function(data) {
             }
 
             // recurse
-            if(item.content && item.content.length > 0 &&
-               recurseFn(item)              ) {
+            if(!item.isEmpty() && recurseFn(item)) {
               monkeys(item, callFn, selectFn, recurseFn);
             }
             i++;
--- a/plugins/localStyles.js
+++ b/plugins/localStyles.js
@@ -307,10 +307,7 @@ exports.fn = function(data) {
     function monkeys(items, callFn, selectFn) {
       if(selectFn  === undefined) { selectFn  = selectAll;     }
 
-        var      i = 0,
-            length = items.content.length;
-
-        while(i < length) {
+        for (var i = 0; i < items.content.length; i++) {
 
             var item = items.content[i];
 
--- a/plugins/localStyles.js
+++ b/plugins/localStyles.js
@@ -295,10 +295,6 @@ exports.fn = function(data) {
     function selectAll() {
       return true;
     }
-    // Default: Always further recurse an item
-    function alwaysRecurse() {
-      return true;
-    }
 
     /**
      * Recursively traverses the HTML AST
@@ -306,13 +302,11 @@ exports.fn = function(data) {
      * @param {Array}    items input items
      * @param {Function} fn for processing an item
      * @param {Function} fn for selecting an item
-     * @param {Function} fn for recursion check
      *
      * @return {Array} output items
      */
-    function monkeys(items, callFn, selectFn, recurseFn) {
+    function monkeys(items, callFn, selectFn) {
       if(selectFn  === undefined) { selectFn  = selectAll;     }
-      if(recurseFn === undefined) { recurseFn = alwaysRecurse; }
 
         var      i = 0,
             length = items.content.length;
@@ -326,8 +320,8 @@ exports.fn = function(data) {
             }
 
             // recurse
-            if(!item.isEmpty() && recurseFn(item)) {
-              monkeys(item, callFn, selectFn, recurseFn);
+            if(!item.isEmpty()) {
+              monkeys(item, callFn, selectFn);
             }
             i++;
         }
--- a/plugins/localStyles.js
+++ b/plugins/localStyles.js
@@ -1,0 +1,359 @@
+'use strict';
+
+exports.type = 'full';
+
+exports.active = true;
+
+exports.description = 'copies styles from <style> to element styles';
+
+
+var cssParser         = require('css'),
+    inlineStylesParse = require('inline-styles-parse'),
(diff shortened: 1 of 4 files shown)


strarsis:
Hm, jshint fails with build using nodejs release v0.10.40.
Should I base off from a different commit than current master?

With best regards

plugins/localStyles.js
```diff
@@ -0,0 +1,359 @@
...
+     * Recursively traverses the HTML AST
+     *
+     * @param {Array}    items input items
+     * @param {Function} fn for processing an item
+     * @param {Function} fn for selecting an item
+     * @param {Function} fn for recursion check
+     *
+     * @return {Array} output items
+     */
+    function monkeys(items, callFn, selectFn, recurseFn) {
```
GreLI: It seems that `recurseFn` isn't actually used.

plugins/localStyles.js
```diff
@@ -0,0 +1,359 @@
...
+        while(i < length) {
+
+            var item = items.content[i];
+
+            if ( selectFn(item) ) {
+              callFn(item);
+            }
+
+            // recurse
+            if(item.content && item.content.length > 0 &&
```
GreLI: You can use 'isEmpty()' method of API. See [lib/svgo/jsAPI.js#L87](https://github.com/svg/svgo/blob/master/lib/svgo/jsAPI.js#L87).

plugins/localStyles.js
```diff
@@ -0,0 +1,359 @@
...
+      // Additionally parse inline CSS from the element (if there is any inline CSS)
+      var inlineAstRule = { declarations: [] };
+      if(elemHasStyles(matchedElement)) {
+        inlineAstRule = getInlineCss(matchedElement).stylesheet.rules[0];
+      }
+
+      // Merge inline CSS and looked up CSS
+      var newAstDeclarations =
+                 styleAstRule.declarations   // inline styles last for precedence
+        .concat(inlineAstRule.declarations);
```
GreLI: Jumping indentation.

plugins/localStyles.js
```diff
@@ -0,0 +1,359 @@
...
+
+      // Store the new (inlined + existing inline) CSS
+      // We don't stringify at match because this would be unnecessarily costly,
+      // when there are more selectors that match.
+      if(!matchedElement.newAstDeclarations)  matchedElement.newAstDeclarations = [];
+          matchedElement.newAstDeclarations = matchedElement.newAstDeclarations.concat(newAstDeclarations);
```
GreLI: Line is indented as it's under `if` condition but it's not.

GreLI:
Also, I'd like if styles were transfered just if there is only one match, not if two or more (non-adjanced, adjanced can be grouped). It seems it can be done easy enough in the current implementation.

strarsis:
"Also, I'd like if styles were transfered just if there is only one match, not if two or more (non-adjanced, adjanced can be grouped). It seems it can be done easy enough in the current implementation."

So this means that when a CSS rule with multiple selectors 
is selected because _one_ selector matched /not all of them), 
then the whole CSS rule should be cleaned?

Like for example using this rule in input:
````
.test1, test2 {
    color: blue;
}
.test3 {
    background: red;
}
[...]
<rect class="test1" [...]
````
The rule is cleaned up completely because .test1 matched?:
````
.test3 {
    background: red;
}
[...]
<rect class="test1" [...]
````

With best regards

**Comment by strarsis** [comment]
The jshint fails with other files for this PR, should I base it off a different commit that passes the tests?
https://travis-ci.org/svg/svgo/jobs/91121264#L176

With best regards

**Comment by GreLI** [comment]
No, I don't think it's required. Do it if you want.

It's odd that other builds including master branch passed it.

**Comment by strarsis** [comment]
OK, so the last thing that needs to be clarified is how matched CSS rules are handled.
So a CSS rule should be cleaned up when any of its selectors matched, right?

With best regards

**Comment by GreLI** [comment]
I mean remove a rule only if there only one corresponding class in document. E.g.
```xml
<svg>
<style>
  .thick { stroke-width: 10px }
</style>
<path class="thick" d="..."/>
</svg>
```
but not then there are several classes here and there:
```xml
<svg>
<style>
  .thick { stroke-width: 10px }
</style>
...
<circle class="thick" .../>
...
<path class="thick" d="..."/>
...etc...
</svg>
```

plugins/localStyles.js
```diff
@@ -0,0 +1,351 @@
...
+     *
+     * @return {Array} output items
+     */
+    function monkeys(items, callFn, selectFn) {
+      if(selectFn  === undefined) { selectFn  = selectAll;     }
+
+        var      i = 0,
+            length = items.content.length;
+
+        while(i < length) {
```
GreLI: Why not just simple `for` loop?
```js
for (var i = 0; i < items.content.length; i++) {
```
Modern engines optimizes it, so there is no need in caching `length = items.content.length`.

strarsis:
Yes, I changed it.
I also try to get rid of the global variables and add some further tests for edge cases.

With best regards

GreLI:
Don't forget to check with «CDATA», empty `<style/>` tag, complex selectors, etc. Just fixed a couple of issues: fc174aa9f0412061a427ab592b7784f63c6ed170.

LeZuse:
Do you guys have any update on this?

strarsis:
Yes, I rewrote the plugin so that it can use the svgo ast as cheerio ast and juice for cleaning up the styles.
cheerio ast back to svgo ast is not fully working yet, hence no new PR has been created yet.

strarsis:
OK, the rewrite has been finished: https://github.com/svg/svgo/pull/592

PR #475, opened 2015-12-29
Added tests for svg2js behavior if passed in malformed svg

yandavid:


GreLI:
I don't know, actually what this test for. I've reworked some issues with errors and parsing errors now just show error instead of throwing (not released yet). However, `error` attribute is still being passed to callback.

PR #476, opened 2016-01-04
Fix "Invalid character entity" with illustrator svgs #272

kaij:
Makes external entities declared in doctype known to the sax parser in svgo. This allows the source file to be saved in illustrator with preserved editing capabilities. Added a very basic test to check and sample file to check whether the parser runs through. Shouldn't have too many side effects. 

Note: there are activities around entity parsing going on in the sax parser itself, see https://github.com/isaacs/sax-js/pull/162. But it doesn't seem to have high priority. I think putting a (maybe temporary) fix in svgo is useful because the workflow Illustrator -> SVG -> SVGO -> Grunt Webfont is quite common.


GreLI:
Thanks for contribution! I was simultaneously thinking in the same direction.

PR #480, opened 2016-01-06
Fixed convertPathData plugin crashing when invoked several times with different precision values.

DanReyLop:
I added the fix, and a test to reproduce this issue.

When you call SVGO with a precision value different than 0, and then (in the same script) you call it with precision 0, an error like this is thrown:

```
Uncaught RangeError: toFixed() digits argument must be between 0 and 20
      at Number.toFixed (native)
      at strongRound (plugins/convertPathData.js:767:36)
      at plugins/convertPathData.js:416:17
      at Array.filter (native)
      at filters (plugins/convertPathData.js:270:17)
      at Object.exports.fn (plugins/convertPathData.js:77:20)
      at lib/svgo/plugins.js:59:45
      ...
```

This little thing might seem pointless (why anyone would do that, right?), but is preventing SVGOMG to use `"floatPrecision=0"` in its UI.

Please tell me if I missed anything.

Cheers!


DanReyLop:
On a side note, Travis is configured to run against Node `0.10` and `0.12`. Is there any specific reason for that? Can we bump it to `0.12`, `4` and `stable`?

plugins/convertPathData.js
```diff
@@ -772,10 +772,7 @@ function roundData(data) {
         return data;
     }
 
-    roundData = precision > 0 ? strongRound : round; // jshint ignore: line
-
-    return roundData(data);
-
+    return (precision > 0 ? strongRound : round)(data);
```
GreLI: In a such case it will every time checking precision which is excessive.

GreLI:
Probably, it's again some incorrectly preserved settings like I fixed in 614bc00d77eee6c0ebdc4b118c9c77ccbae1d901. Have you user recent version of svgo?

DanReyLop:
I've cloned the last commit on `master`, it's still happening.

plugins/convertPathData.js
DanReyLop: This method will round a set of values, iterating them and invoking `Number.toFixed` in every case (and sometimes more logic like `Math.abs`). Compared to that, a single integer comparison more or less is just a micro-optimization. In my opinion, overwriting the `roundData` function reference for the sake of gaining a couple of CPU cycles is just not worth it.

GreLI:
Thanks for the test case! Fixed it a bit differently.

DanReyLop:
Fine for me.
I still think that the difference in performance is negligible, but your solution is also cleaner :)

PR #497, opened 2016-02-12
Added plugin that converts fill values to "currentColor" so you can s…

doot0:
…tyle the compiled SVGs with CSS color property

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -55,6 +55,7 @@ plugins:
   - removeAttrs
   - addClassesToSVGElement
   - removeStyleElement
+  - convertFillsToCurrentColor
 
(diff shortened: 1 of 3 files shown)


CodeTheory:
:+1: This would be very useful for me. I currently do this manually.

GreLI:
Not sure, this plugin should be in SVGO. Besides, it has quite odd parsing and doesn't follow the code-style. (You can somewhat check it with `make jshint` command.)

However, there is a plugin [convertColors.js](https://github.com/svg/svgo/blob/master/plugins/convertColors.js). It already has color parsing (while I'm not sure you're really need it) and you can just add an option to convert `fill` to `currentColor`. I think it's the best way.

GreLI:
#521 done it.

PR #521, opened 2016-03-21
Add currentColor param to convertColors plugin

scurker:
This is related to #497, but instead of creating a new plugin it adds an additional param to the existing `convertColors` plugin. 

Part of the issue that this addresses, if a color is defined within an SVG you can't override the color easily in CSS without manually editing the SVG. With this new param, you can automatically convert colors to `currentColor` allowing CSS overrides.

The only downside of it being a param within `convertColors`, is that it will override any other params when set to true. That's why initially `currentColor` defaults to `false`.

The only other potential change I could see is to take into account `transparent` values. I'm not sure if transparent values should or should not be converted. For my use case, if a color value is transparent we don't need it changed, but I could see it being different for someone else.

### DIFF
--- a/plugins/convertColors.js
+++ b/plugins/convertColors.js
@@ -19,7 +19,7 @@ var collections = require('./_collections'),
     rComma = '\\s*,\\s*',
     regRGB = new RegExp('^rgb\\(\\s*' + rNumber + rComma + rNumber + rComma + rNumber + '\\s*\\)$'),
     regHEX = /^\#(([a-fA-F0-9])\2){3}$/,
-    none = /none/i;
+    none = /\bnone\b/i;
 
 /**
  * Convert different colors formats in element attributes to hex.
--- a/plugins/convertColors.js
+++ b/plugins/convertColors.js
@@ -7,6 +7,7 @@ exports.active = true;
 exports.description = 'converts colors: rgb() to #rrggbb and #rrggbb to #rgb';
 
 exports.params = {
+    currentColor: false,
     names2hex: true,
     rgb2hex: true,
     shorthex: true,
@@ -17,7 +18,8 @@ var collections = require('./_collections'),
     rNumber = '([+-]?(?:\\d*\\.\\d+|\\d+\\.?)%?)',
     rComma = '\\s*,\\s*',
     regRGB = new RegExp('^rgb\\(\\s*' + rNumber + rComma + rNumber + rComma + rNumber + '\\s*\\)$'),
-    regHEX = /^\#(([a-fA-F0-9])\2){3}$/;
+    regHEX = /^\#(([a-fA-F0-9])\2){3}$/,
+    none = /none/i;
 
 /**
  * Convert different colors formats in element attributes to hex.
@@ -55,6 +57,11 @@ exports.fn = function(item, params) {
 
                 var val = attr.value,
                     match;
+
+                // Convert colors to currentColor
+                if (params.currentColor && (match = !val.match(none))) {
+                    val = 'currentColor';
+                }
 
                 // Convert color name keyword to long hex
                 if (params.names2hex && val.toLowerCase() in collections.colorsNames) {
--- a/test/plugins/convertColors.04.svg
+++ b/test/plugins/convertColors.04.svg
@@ -1,0 +1,25 @@
+<svg xmlns="http://www.w3.org/2000/svg">
+    <g color="black"/>
+    <g color="BLACK"/>
+    <g color="none"/>
+    <path fill="rgb(64, 64, 64)"/>
+    <path fill="rgb(86.27451%,86.666667%,87.058824%)"/>
+    <path fill="rgb(-255,100,500)"/>
+    <path fill="none"/>
+</svg>
+
+@@@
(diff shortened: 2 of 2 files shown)


scurker:
@doot0 @CodeTheory Would this still help resolve your issue with currentColor in #497?

plugins/convertColors.js
```diff
@@ -17,7 +18,8 @@ var collections = require('./_collections'),
     rNumber = '([+-]?(?:\\d*\\.\\d+|\\d+\\.?)%?)',
     rComma = '\\s*,\\s*',
     regRGB = new RegExp('^rgb\\(\\s*' + rNumber + rComma + rNumber + rComma + rNumber + '\\s*\\)$'),
-    regHEX = /^\#(([a-fA-F0-9])\2){3}$/;
+    regHEX = /^\#(([a-fA-F0-9])\2){3}$/,
+    none = /none/i;
```
GreLI: Probably, it's good idea to make it a bit more specific like at least `/\bnone\b/`.

GreLI:
That's a nice one! However, it will convert not only `fill` but `stroke` as well. May be there is need in more control over the properties.

doot0:
@scurker Seems like it would, thank you! Hope this gets merged in.

scurker:
@GreLI Correct, it'll convert both allowing you to convert colors for everything within css:

```css
svg {
  stroke: #000;
  fill: #000;
}
```

However, I could see potentially wanting more granular control over paths vs strokes. Are there any other examples of plugins that allow you to pass in an array of options?

GreLI:
I don't have strong opinion, personally. If it's ok for @doot0, I guess it's fine. From the other point view, one can have full control through the image itself by setting certaing properties

 Anyway, it's not the main svgo purpose, just a sideeffect, that can be achievied almost for nothing.

kevinSuttle:
Nice!

kevinSuttle:
So, how is it used? I can't find the docs. Sorry.

Fish2:
any chance you could add this in with it ? https://github.com/svg/svgo/issues/884

mnpenner:
I can't seem to get this to work.

I've got this `svgo.yml`:

```yml
plugins:
  - convertColors:
      currentColor: true
```
I'm running svgo with:

```sh
npx svgo --config=svgo.yml src/icons/fa/solid
```

And the source for one of my icons is like:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 512"><path d="M177 159.7l136 136c9.4 9.4 9.4 24.6 0 33.9l-22.6 22.6c-9.4 9.4-24.6 9.4-33.9 0L160 255.9l-96.4 96.4c-9.4 9.4-24.6 9.4-33.9 0L7 329.7c-9.4-9.4-9.4-24.6 0-33.9l136-136c9.4-9.5 24.6-9.5 34-.1z"/></svg>
```

I was hoping it would add `fill="currentColor"` to the `<path>` but that doesn't seem to be happening.

What am I missing?

kevinSuttle:
Don't you need to have a `fill` present for it to convert?

GreLI:
Yeah, it changes the value of `fill`, but doesn't add the attribute.

scurker:
If you look at the [test for currentColor](https://raw.githubusercontent.com/scurker/svgo/8e5706485e993267f1be7e72048e69649317c5ff/test/plugins/convertColors.04.svg), this plugin converts color values (such as named colors, hex values, rgb, etc.) to `currentColor`. You need to set a color attribute to replace with your path.

`<path fill="black" ... />`

mnpenner:
Oh. Well, I want to add it then. I just found another plugin, but it's breaking my SVGs:

```yml
plugins:
  - convertColors:
      currentColor: true
  - addAttributesToSVGElement:
      attributes:
        - fill: currentColor
```

Is making my SVG come out like:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512" [object Object]><path d="..." /></svg>
```

But the [comments](https://github.com/svg/svgo/blob/master/plugins/addAttributesToSVGElement.js) make it look like this should work...

kevinSuttle:
I think it might be the order in which you have them. Try swapping them?

mnpenner:
@kevinSuttle Tried swapping the order, same result. Tried deleting `convertColors` altogether, still doesn't work. Always comes out `[object Object]`

mnpenner:
Very strange.

I dropped some debugging in there, it looks like it's adding the attribute correctly:

```json
{ elem: 'svg',
  prefix: '',
  local: 'svg',
  attrs: 
   { xmlns: 
      { name: 'xmlns',
        value: 'http://www.w3.org/2000/svg',
        prefix: 'xmlns',
        local: '' },
     viewBox: 
      { name: 'viewBox',
        value: '0 0 512 512',
        prefix: '',
        local: 'viewBox' },
     fill: { name: 'fill', value: 'currentColor', prefix: '', local: 'fill' } },
```

Okay, and now I removed my debugging and they're coming out fine. What the heck.

I must have been doing something stupid. Thanks for your help anyway :smile:

PR #539, opened 2016-04-28
Precision Scaling Plugin

keithclark:
This PR adds a new precision scaling plugin to SVGO. This plugin scales decimal values by powers of ten to remove the decimal point from the SVG source.

The plugin accepts a single boolean parameter, `applyViewBox`. If set to `true` (and the image has with/height attributes), a viewBox will be added to the root SVG element and the original width/height values will not be scaled. This allows a client to render the scaled image at the original size. If `false` the viewBox will be omitted and SVG clients will render the image at the scaled size.

Depending on the SVG, this plugin can shave up to 10% off the compressed output, resulting in an average 2-5% reduction of SVG when gzip'd. In some files, scaling precision can lead to a larger compressed image so this plugin is deactivated by default.

| Image                     |Original|Compressed|Compressed (scalePrecision)|Gzip  |Gzip (scalePrecision)|
|---------------------------|-------:|---------:|--------------------------:|-----:|--------------------:|
| [gallardo.svg][gallardo]  |340873  |130918    |__124607__                 |32071 |__31510__            |
| [svgo.svg][svgo]          |25791   |21679     |__21534__                  |3657  |__3573__             |
| [car.svg][car]            |527014  |200945    |__186465__                 |49849 |__48909__            |

### General
- [ ] Handle non-pixel values better in `getPrecision`/`scalePrecision`
- [ ] Finalise the list of attributes that can be scaled

### Enhancements
- [ ] Only add scaled `stroke-width` to the root element if shape/text elements use the default width

[gallardo]:https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/gallardo.svg 
[svgo]:https://github.com/svg/svgo/blob/master/logo.svg
[car]:https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/car.svg

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -42,6 +42,7 @@ plugins:
   - moveGroupAttrsToElems
   - collapseGroups
   - convertPathData
+  - scalePrecision
   - convertTransform
   - removeEmptyAttrs
   - removeEmptyContainers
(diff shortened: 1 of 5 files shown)


keithclark:
Ooops! I raised this to the wrong repo. It's not ready for the main project yet.

PR #540, opened 2016-05-01
Created illustratorLayerIdsToClasses plugin

rally25rs:
Addition of a plugin that scans elements for `data-name` or `id` attributes and copies the value to the `class` attribute.

This is intended to be used with the Adobe Illustrator export option "Object IDs: Layer Names" to make it easier to get meaningful class names exported into SVGs from Illustrator.

I have a blog post that explains the intended use case: https://codingwithspike.wordpress.com/2016/05/01/css-styling-illustrator-svgs/

Since this is intended to be used with Illustrator exported SVGs, the plugin is disabled by default.

I updated the README.ru.md file with a Google translation from my english. Feel free to re-translate.

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -24,6 +24,7 @@ plugins:
   - cleanupAttrs
   - minifyStyles
   - convertStyleToAttrs
+  - illustratorLayerIdsToClasses
   - cleanupIDs
   - removeRasterImages
   - removeUselessDefs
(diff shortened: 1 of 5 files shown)


rally25rs:
Can anyone comment on why some PRs (like this one) are being left open for months without any comments, but other PRs are being accepted/merged?

GreLI:
Yes, sorry for the delay. Unfortunately, your PR is out of scope of SVGO, and thus I cannot accept it. It just have nothing with optimization which is the purpose of SVGO.

derhuerst:
> Unfortunately, your PR is out of scope of SVGO, and thus I cannot accept it.

I think that's up for discussion if it's out of scope. (Personally, I don't really have an opinion on this.)

But since this is a plugin that can be disabled by default, I don't see a reason why it shouldn't be included if it's a reasonably common use case.

rally25rs:
I doubt its a common use case, but it let me clean up my illustrator exported SVGs in one pass through SVGO. I can certainly find another more general SVG manipulating package and move my plugin there and pass images through 2 applications.

derhuerst:
I think the SVGO plugin setup is a bit more complicated than necessary, but as far as I could see it, bundling SVGO for in-browser won't include unsused plugins. Therefore, I don't see a reason not to include this plugin.

GreLI:
Closing it since it doesn't have any relation to optimization.

Issue #543, opened 2016-05-05
add valueless attributes


gjjones:
I'm currently outputting SVGO to a generated JSX file for our React icons library. We're looking to use the JSX hooks as part of the base element's attributes.

That would look like the following:

``` xml
<svg className={className} {...props} />
```

Since I need the attribute value to not have wrapping quotes, I decided to make a plugin to add `className={className}` as the attribute key, and provide an empty value. This does not work, as SVGO expects attributes to have a value, and assigns an empty string when value is empty. 

The following logic is what sets that:

``` js
//svgo/js2svg.js
    elem.eachAttr(function(attr) {

        attrs +=    ' ' +
                    attr.name +
                    this.config.attrStart +
                    String(attr.value).replace(this.config.regValEntities, this.config.encodeEntity) +
                    this.config.attrEnd;

    }, this);
```

What are you thoughts on allowing attributes to be valueless? I can create a PR to allow undefined in the addAttrs helper. And here we could check that value before adding the start/value/end series to the attrs string.

I understand that allowing users to specify an attribute without a valid could result in invalid xml and would love to hear alternatives if you have them.

GreLI:
Well, while SVG has not such attributes (at least I'm aware of), some content in `<foreignObject>` may have it, so it's worth to consider.

Note that [JSAPI.prototype.addAttr](https://github.com/svg/svgo/blob/a7624a0c438cc6ee9a1d29b4ecba0bd1ec510f0a/lib/svgo/jsAPI.js#L247) currently doesn't support adding valueless attributes either.

gjjones:
I'm open to making a PR if it'd be useful to see

GreLI:
You're welcome.

PR #546, opened 2016-05-12
adds addAttributes plugin

gjjones:
Allows adding valueless attributes to the svg element.


gjjones:
This PR is meant to address #543

mattyod:
+1

UsulPro:
Hi!
Am I right that it convert svg code to one jsx-sutable ?
if it so could you, please, give a short example how to use it finally?

gjjones:
@UsulPro, here is an example of how I used it for a plugin to convert SVGs to React components:
https://gitlab.com/gjjones/svg-to-react/blob/master/src/utilities.js#L6

PR #565, opened 2016-07-06
Add `removeElements` plugin

elidupuis:
This is a simple plugin that allows you to remove certain elements based on their `id` attributes.

We need this for a specific workflow in one of our projects where SVGs are exported from Photoshop with a `rect` to set the specifically to set the `viewBox` dimensions—but that `rect` needs to be removed before the SVGs are rendered in our app.

Definitely open to thoughts and improvements!


GreLI:
No doubts in content, but the name `removeElements` is too vague. It's hard to understand that the plugin removes elements with certain ids by the name.
Also, is some extensibility being supposed? E.g. removing by classes or other attributes.

elidupuis:
Yeah, ideally this would accept more than just `id` attributes but I haven't wrapped my head around how to make it work yet 😉

elidupuis:
@GreLI Any suggestions on a configuration format for allowing more flexible element selectors? Are there other plugins that use a CSS-like selector format?

elidupuis:
How about `removeElementsByAttr` with config similar to `removeAttrs`?

```
removeElementsByAttr:
  attrs:
    - 'id:foo'
    - 'class:bar'
```

This plugin config would remove elements with `id="foo"` or `class="bar"`.

GreLI:
> Any suggestions on a configuration format for allowing more flexible element selectors? Are there other plugins that use a CSS-like selector format?

No, there is no such plugins. And I don't think that it's really needed. It's somewhat out of SVGOs scope. However, since this plugin is removing stuff, it could be useful. Could you add removing elements by classes as well? Then it'll be worth enough.

> How about `removeElementsByAttr` with config similar to `removeAttrs`?

Yeah, something like that would be good. I'd like to suggest to have separate `ids`, `classes`, and `attrs` properties.

Optionally, for classes it worth to allow inner arrays for class intersections like `.foo.bar`. But it's not required.

Perhaps, in a future a separate property for set of requierements like “with class `figure` without `fill` (`attrs: fill: false`). But that's a whole new world.

GreLI:
:+1:

PR #576, opened 2016-08-19
When minify:false cleanupIds deletes too many ids

tombigel:
when running cleanupIds with minify:false option it does not save a reference to all ids and removes ids that shouldn't be removed.
This fix adds a var that keeps a ref to the original id even if minify is not defined


GreLI:
Nice! Thanks!

PR #582, opened 2016-08-26
Adds addAttributesToSVGElement to default plugins list.

gjjones:
Allows use of the addAttributesToSVGElement by adding to defaults plugins list.


gjjones:
Addresses #583

PR #587, opened 2016-08-31
Added recursive bypass feature

vkhv:


lib/svgo/coa.js
```diff
@@ -131,6 +139,7 @@ module.exports = require('coa').Cmd()
         .end()
     .act(function(opts, args) {
 
+	    console.log(`opts - ${opts.folder}, args - ${args}`)
```
GreLI: Has it left from debugging? SVGO works on node 0.12 and higher, where is no support for template strings.

lib/svgo/coa.js
```diff
@@ -53,6 +54,13 @@ module.exports = require('coa').Cmd()
         })
         .end()
     .opt()
+        .name('recursive').title('Input folder, optimize and rewrite all *.svg files')
+        .short('rf').long('recursive')
```
GreLI: Why `rf`?

vkhv: Now - r.
25d3f017a178f34429135771342006ac26f1248e

lib/svgo/coa.js
vkhv: Yes, just debug, fix - 25d3f017a178f34429135771342006ac26f1248e

GreLI:
Many things are wrong here.

PR #592, opened 2016-09-06
Add inlineStyles plugin (rewrite of localStyles plugin)

strarsis:
This pull request adds the inlineStyles plugin for inlining styles, 
also correctly handling selector specificity, cascading, 
advanced selectors, etc and intelligently cleaning up matched selectors.

Additionally, direct CSS selector support (powered by [css-select](https://github.com/fb55/css-select), [csstree](https://github.com/csstree/csstree) and [csso](https://github.com/css/csso)) is added to svgo which 
is used by this plugin ([`querySelectorAll(selector)`](https://developer.mozilla.org/en/docs/Web/API/Document/querySelectorAll), [`querySelector(selector)`](https://developer.mozilla.org/en/docs/Web/API/Document/querySelector), [`matches(selector)`](https://developer.mozilla.org/en/docs/Web/API/Element/matches)).
Furthermore, the [CSSStyleDeclaration interface](https://developer.mozilla.org/en/docs/Web/API/CSSStyleDeclaration) is added to svgo jsAPI for handling element styles.
As latest addition, [Element.classList property](https://developer.mozilla.org/en/docs/Web/API/Element/classList) is added to svgo JSAPI for proper cleaning up class attributes.
As little extra, the last unnecessary semicolon of inline styles declarationList is cleaned up.


strarsis:
Some cleanup necessary...

strarsis:
So tests are passing and I guess it could be merged now.
There may be some controversy concerning the svgo-cheerio-svgo AST conversion though.

package.json
```diff
@@ -48,13 +48,16 @@
     "jshint": "jshint --show-non-errors ."
   },
   "dependencies": {
-    "sax": "~1.2.1",
+    "cheerio": "^0.22.0",
     "coa": "~1.0.1",
-    "js-yaml": "~3.6.1",
     "colors": "~1.1.2",
-    "whet.extend": "~0.9.9",
+    "css": "^2.2.1",
```
GreLI: CSSO also has a parser (but probably with a bit different AST), so it could be reused without adding dependency.

strarsis:
Coming back to the discussion in https://github.com/svg/svgo/pull/447.

Should this plugin now remove all classes that 
a) matched some elements and have been inlined and 
b) aren't used multiple times among elements?

Or should this go into a separate plugin just for this task, 
which means that this plugin only inlines, 
but doesn't change the style elements content and element class attributes?

test/plugins/inlineStyles.02.svg
```diff
@@ -0,0 +1,17 @@
...
+        .st0,.st2{fill:red;}
+    </style>
+    <rect width="100" height="100" class="st0" style="stroke-width:3;stroke:blue;"/>
+</svg>
+
+@@@
+
+<svg id="test" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
+    <style>
+        .st2 {
```
GreLI: Shouldn't this class be removed?

GreLI:
I think the pugin should remove inlined classes (as well as emptied `<style>` elements). There is no reason to keep them. However, if a class is used many times, probably it's better to leave it as is without inlining.

If the plugin will allways inline classes, I don't think it's a good idea to make it run by default.

strarsis:
OK, so I will adjust the plugin to use csso ast instead and modify the css class removal logic.

strarsis:
So the plugin will now remove selectors that matched once, 
but preserve selectors that matched multiple times in styles elements.

strarsis:
So tests are passing, but there may be one thing left: 
Selectors (including classes) of classes that match multiple times are spared by this plugin - but still inlined.
I assume that it makes sense that multiple times matching selectors aren't only spared but also not inlined.
For this I would have to add an issue/PR for [juice](https://github.com/Automattic/juice/issues) that is used for the inlining.

zakness:
Not sure the status of this PR, I came across it because I want to avoid the class/id style leakage mentioned here: https://github.com/svg/svgo/issues/296#issuecomment-127957035.

It seems this plugin as currently designed would not solve the leakage problem because it does not inline selectors that match multiple times. To solve leakage it would have to always inline and always remove the `<style>` blocks. Do I understand this correctly?

Thanks for putting the effort into this plugin!

strarsis:
@zakness: The current PR code would inline everything, so you could use it from this fork/branch.
Additionally to inlining everything, an option/feature is requested that would only inline selectors that match once.

zakness:
Ah okay, thanks for the quick response @strarsis :)

GreLI:
@strarsis there is one caveat with the PR to juice. Since version 3 it requires Node.js >= 4, while svgo works on 0.10 and higher (and there was complains recently, when it was bumped to 0.12, so I've reverted changes). If you update dependency on juice you will not able to use the plugin on lower Node.js versions.

strarsis:
@GreLI The plugin has now been modified to use a "light" inliner alternative on top of cheerio.
When the version issues with juice have been fixed, juice can be used instead then.

plugins/inlineStyles.js
```diff
@@ -0,0 +1,124 @@
...
+      }
+    });
+  });
+
+  // stable sort selectors by specificity
+  var selectorItemsSorted = stable(selectorItems, function(item1, item2) {
+    return SPECIFICITY.compare(item1.selectorStr, item2.selectorStr);
+  });
+
+  for(var selectorItem of selectorItemsSorted) {
```
GreLI: Unfortunatetly, `for-of` loops aren't supported in Node 0.10 which is still supported. However, probably not for a long time.

strarsis: OK, I replaced them with for-in and an index variable loops.

diedsmiling:
Are there any updates on this plugin?

strarsis:
Juice has been replaced now with css-select which can use the svgo AST directly and gives SVGO CSS selector capabilities.
Currently mocking has to be used for injecting domutils for svgo AST into css-select, 
but passing it directly should be possible soon, see https://github.com/fb55/css-select/pull/66.

martinjuhasz:
Whats the current status on this? I'm in the need for moving styles declared in a <style> tag into every corresponding <path> tag, which this plugin seems to handle, right?

Will this be merged into the main repo?

strarsis:
@martinjuhasz You could use this plugin now, but for it into master, css-select has to be released to npm first.

For trying it out from source:
````
$ npm install github:strarsis/svgo#inlineStyles
````

**Comment by strarsis** [comment]
Currently this plugin can also filter for mediaqueries (at-rules) (it uses 'screen' and plain styles by default).
However, I am not sure whether it wouldn't be more elegant to have a second svgo plugin that selectively selectively moves styles out of their at-rules so that subsequent plugins use the same set of styles.

**Comment by strarsis** [comment]
Travis seems to have issues with npm github references: https://travis-ci.org/svg/svgo/jobs/182635289#L164

**Comment by alexjlockwood** [comment]
Hi there,

I'm not certain if this is still a work in progress but just wanted to let you know in case you weren't aware... I tried forking your branch and ran your plugin using svgo and saw unexpected results with your plugin enabled vs. disabled.

Here is a gist containing the SVGs I used/generated: https://gist.github.com/alexjlockwood/f5d65651f2357f2fc9b4781c6e124778

I ran the following commands:

```
svgo --disable=inlineStyles --pretty -i kids-source.svg -o kids-disabled.svg
svgo --enable=inlineStyles --pretty -i kids-source.svg -o kids-enabled.svg
```

As you can see in the gist, the third kid's hat is missing its fill color when the plugin is enabled.

The style tag for the source SVG looks like:

```
.cls-1,
.cls-7{fill:none}
.cls-2{opacity:.08}
.cls-3{clip-path:url(#clip-path)}
.cls-4{fill:#ffcc80}
.cls-5{fill:#5c6bc0}
.cls-6{fill:#212121}
.cls-7,
.cls-8{stroke:#000;stroke-linecap:round;stroke-miterlimit:10;stroke-width:.5px}
.cls-9{fill:#ff9800}
.cls-10{fill:#fb8c00}
.cls-11{fill:#00838f}
.cls-12{fill:#ffeb3b}
.cls-13{fill:#8e593c}
.cls-14{fill:#9575cd}
.cls-15{fill:#4e342e}
.cls-16{fill:#fff;opacity:.2}
```

Interestingly, if I change the lines beginning with `cls-1` and `cls-7` using the code below, the plugin works fine:

```
.cls-1{fill:none}
...
.cls-7{stroke:#000;stroke-linecap:round;stroke-miterlimit:10;stroke-width:.5px}
...
```

Let me know if you are seeing the same thing!

**Comment by strarsis** [comment]
@alexjlockwood:
I have the same issue. I isolated the causing part and created a test for this.
Apparently changes leak into previously referenced declarations when invoking csso.translate at the end.

**Comment by alexjlockwood** [comment]
that's great! thanks again for this... it's going to make my life a lot easier when dealing with UXers giving me SVGs that were exported by adobe illustrator. :D 

btw, do you have a ball park estimate of when this will be ready to be merged?

**Comment by strarsis** [comment]
@alexjlockwood: The recent css-select has to be [released to npm](https://github.com/fb55/css-select/issues/72).
The plugin should be complete now, so I guess that it is merged very soon after css-select has been released.

**Comment by alexjlockwood** [comment]
Any updates on this?

**Comment by strarsis** [comment]
@alexjlockwood: css-select has to be released.

**Comment by laurentvd** [comment]
Great work people! Really looking forward to this becoming a default svgo plugin. What's the usual procedure and timeline for these kinds of PRs to become merged and published?

**Comment by strarsis** [comment]
@laurentvd: You can already use it by installing svgo from this PR branch (see above).
Currently not the latest [css-select](https://github.com/fb55/css-select/issues/72) release is used but a newer commit because of features/API that has been added to it.

**Comment by laurentvd** [comment]
@strarsis thanks and yes I read those instructions (which work fine). The thing is - however - that I'm using a package which uses a package which uses `svgo` ;-) You probably see my problem there. I'll wait for it to be merged, was just curious on the timeline for this.

**Comment by martinjuhasz** [comment]
There is no timeline. Just when @fb55 decides to release a new version of `css-select`

**Comment by fb55** [comment]
@martinjuhasz I've released the current state as `1.3.0-rc0`. Does that work for you?

**Comment by strarsis** [comment]
@fb55: Thanks! Now a normal npm package can be used for css-select.

@GreLI: Ready to merge now?

**Comment by strarsis** [comment]
Support for pseudo classes added (previously issues with styles that use pseudo classes).
Styles in pseudo classes can now be optionally selectively inlined for special purposes (e.g. ':link' + ':hover').

**Comment by strarsis** [comment]
@GreLI : Now this could be merged, any changes required?

**Comment by GreLI** [comment]
It's a complicated topic. Thorough review is needed.

**Comment by strarsis** [comment]
@GreLI: I could separate the code for adding css selection capabilities to svgo and the inline styles plugin (that uses them) into two PRs.

**Comment by Nejik** [comment]
strarsis, thank you very much for your work, your plugin is indeed what we need! 
Hope GreLI will merge it soon

**Comment by satazor** [comment]
I'm in desperate need of this to be merged so that I can fully automate my webpack config in conjuction with an external svg sprite.

👍

**Comment by alexjlockwood** [comment]
Agree, this would be super awesome to have in svgo. Software like Adobe Illustrator outputs huge/spammy SVG files, and this plugin is necessary in order to reduce the size.

**Comment by alexjlockwood** [comment]
Hey, just wanted to let you know that I got an error when I tried to process the following svg using your plugin:

https://gist.github.com/alexjlockwood/0db6f64f81eb1eb54434636155463615

The SVG has a ton of unused stuff in the `<style>` tag so I assume it has something to do with that.

**Comment by strarsis** [comment]
@alexjlockwood: The issue is caused by the [deep combinator selectors](https://www.w3.org/TR/css-scoping-1/#deep-combinator) from the CSS Scoping Module used in the styles of that SVG.
css-select/css-what apparently don't support this yet.

Graceful handling of errors during selection has been added, the selectors are just skipped then.

**Comment by alexjlockwood** [comment]
Ahh interesting... and I figured I would also ask... does this affect the minifyStyles SVGO plugin in any way?

One thing I noticed was that the inlineStyles plugin took a *very* long time to finish. This does not surprise me at all considering how many selectors are in the `<style>` tag. But I also wonder if the minifyStyles plugin could have saved a bunch of time by removing the un-used selectors in there. Is that possible? (I noticed you wrote that plugin too :P).

**Comment by strarsis** [comment]
I just made a little benchmark, take it with a grain of salt - compare the numbers relatively to each other.
Running svgo only with the inlineStyles plugin against that particular SVG (https://gist.github.com/alexjlockwood/0db6f64f81eb1eb54434636155463615) takes about 1.3 seconds.
minifyStyles before inlineStyles actually increases the required time to about 3.5 seconds.
First inlineStyles, then minifyStyles (as in .svgo.yml for this PR) requires also about 3.5 seconds.
Running minifyStyles before and after inlineStyles increases required time to about 5 seconds.
Without inlineStyles and minifyStyles plugins it takes only about 150 ms.

So using minifyStyles in its current implementation before inlineStyles (and afterwards, because one wants everything minified) doesn't speed up but slows down the process.

But besides this, inlineStyles could be of course also further optimized in some places like in the loop where external style declarations are merged with already inlined styles-

File: plugins/inlineStyles.js [hunk 4]
```diff
@@ -0,0 +1,260 @@
...
+  *
+  *   useMqs (default: ['screen'])
+  *     what mediaqueries to be used,
+  *     non-mediaquery styles are always used
+  *
+  *   usePseudoClasses (default: [])
+  *     what pseudo-classes to be used,
+  *     non-pseudo-class styles are always used
+  *
+  * @param {Object} root (document)
```
alexjlockwood: FYI, these two argument names are out of date.

strarsis:
@alexjlockwood: A script can normally run only on one logical cpu core in node, so using a suitable task runner one could make it utilise all available logical cores. For a system with 8 logical cores that could result in 8x speed up.

The minifyStyles plugin has been modified recently and extra logic added that cleans up classes that aren't used in the SVG. This increases required time and it may be a good idea restoring the previous behaviour and moving the classes cleanup to a separate plugin that is usually not used after inlining all styles.

package.json
```diff
@@ -48,13 +48,17 @@
...
   },
   "dependencies": {
-    "sax": "~1.2.1",
     "coa": "~1.0.1",
-    "js-yaml": "~3.7.0",
     "colors": "~1.1.2",
-    "whet.extend": "~0.9.9",
+    "css-select": "^1.3.0-rc0",
+    "css-select-base-adapter": "^0.1.0",
+    "csso": "~2.3.1",
```
alexjlockwood: Just so you know, stuff seems to break when you upgrade to the most recent version of csso (v3.0.0). I get the following error:

ERROR in ./src/app/scripts/svgo/plugins/inlineStyles.ts
Module not found: Error: Can't resolve 'csso/lib/compressor/restructure/prepare/specificity' in '/Users/alockwood/ShapeShifter/src/app/scripts/svgo/plugins'
 @ ./src/app/scripts/svgo/plugins/inlineStyles.ts 6:20-82

strarsis:
This is nice, latest csso supports also latest css-tree which allows to use new AST walking methods.

package.json
strarsis: Yes, csso separated the tree library (css-tree).

lahmatiy: CSSO 3.0 has been changed a lot, moved to use `css-tree` and had changes in AST format and API. I think SVGO should migrate to 3.0 first (I suppose no changes in sources is needed beside a version change in `package.json`).
New path for `specificity` is `/lib/restructure/prepare/specificity.js`. But that's unsafe to use it that way.
Anyway, feel free if you need help in migration to new AST format and API.

strarsis:
inlineStyles plugin has been rewritten for using latest csstree for handling the AST.
Furthermore, CSSStyleDeclaration interface has been implemented for svgo jsAPI, 
it is possible to set/get styles on each node now with intelligent AST translation.

lib/css-tools.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    var pseudoSelectorsStr = csstree.translate(pseudoSelectorsAst);
+
+    return usePseudos.indexOf(pseudoSelectorsStr) > -1;
+  });
+}
+
+function cleanPseudos(selectors) {
+  selectors.map(function(selector) {
```
lahmatiy: Alternative:
```js
selectors.forEach(function(selector) {
    selector.pseudos.forEach(function(pseudo) {
        pseudo.list.remove(pseudo.item);
    });
});
```

lib/css-tools.js
```diff
@@ -0,0 +1,130 @@
...
+      pseudo.list.remove(pseudo.item);
+    }
+  });
+}
+
+
+// extracted from https://github.com/keeganstreet/specificity/blob/master/specificity.js#L211
+function compareSpecificity(aSpecificity, bSpecificity) {
+  for (var i = 0; i < 4; i += 1) {
+			if (aSpecificity[i] < bSpecificity[i]) {
```
lahmatiy: Something wrong with indentation.

lib/css-tools.js
```diff
@@ -0,0 +1,130 @@
...
+  return stable(selectors, _bySelectorSpecificity);
+}
+
+
+// csso declaration to style declaration
+function cssoToStyleDeclaration(declaration) {
+  var propertyName     = declaration.property,
+      propertyValue    = csstree.translate(declaration.value),
+      propertyPriority = (declaration.important ? 'important' : '');
+  return { name: propertyName, property: { value: propertyValue, priority: propertyPriority } };
```
lahmatiy: It can be simpler `{ name: ..., value: ..., priority: ... }`?

lib/svgo/css-style-declaration.js
```diff
@@ -0,0 +1,79 @@
...
+};
+
+
+CSSStyleDeclaration.prototype.update = function() {
+  var self = this;
+  if(!this.parentNode.hasAttr('style')) {
+    return;
+  }
+  var inlineCssStr = this.parentNode.attr('style').value;
+  var cssoDeclarations = csstree.parse(inlineCssStr, {context: 'declarationList'});
```
lahmatiy: Since value details is not needed here, we can use `parseValue: false` option for `parse()`. It can improve performance a little.

lib/svgo/css-style-declaration.js
```diff
@@ -0,0 +1,79 @@
...
+  cssoDeclarations.children.each(function(cssoDeclaration) {
+    var styleDeclaration = csstools.cssoToStyleDeclaration(cssoDeclaration);
+    self.setProperty(styleDeclaration.name, styleDeclaration.property.value, styleDeclaration.property.priority);
+  });
+};
+
+
+CSSStyleDeclaration.prototype.getCssText = function() {
+  var declarationList = csstree.parse('', {context: 'declarationList'});
```
lahmatiy: It's not clear why parse/translate needed here. Looks like this solution is good enough:
```js
CSSStyleDeclaration.prototype.getCssText = function() {
  return this.properties.map(function(property, propertyName) {
    var strImportant = property.priority === 'important' ? '!important' : '';
    return propertyName + ':' + property.value + strImportant;
  }).join(';');
};
```

plugins/inlineStyles.js
```diff
@@ -0,0 +1,181 @@
...
+
+  for(var styleEl of styleEls) {
+    if(styleEl.isEmpty()) {
+      // skip empty <style/>s
+      continue;
+    }
+    var cssStr = styleEl.content[0].text || styleEl.content[0].cdata || [];
+
+    // collect <style/>s and their css ast
+    var cssAst = csstree.parse(cssStr, {context: 'stylesheet'});
```
lahmatiy: `stylesheet` context is by default, no need to specify it explicitly.

plugins/inlineStyles.js
```diff
@@ -0,0 +1,181 @@
...
+
+    if(opts.removeMatchedSelectors && selectedEls !== null && selectedEls.length > 0) {
+      // clean up matching simple selectors if option removeMatchedSelectors is enabled
+      selector.rule.selector.children.remove(selector.item);
+    }
+  }
+
+
+  for(var style of styles) {
+    csstree.walkRules(style.cssAst, function(node, item, list) {
```
lahmatiy: CSSO removes all at-rules and rules with empty blocks. Do we need it here too?

lib/css-tools.js
```diff
@@ -0,0 +1,130 @@
...
+
+
+function filterByMqs(selectors, useMqs) {
+  return selectors.filter(function(selector) {
+    if(selector.atrule === null) {
+      return useMqs.indexOf('') > -1;
+    }
+
+    var mqName = selector.atrule.name;
+    var mqExpr = csstree.translate(selector.atrule.expression);
```
lahmatiy: It makes sense only when `selector.atrule.expression`'s type is `MediaQueryList`

plugins/inlineStyles.js
strarsis: This was considered in https://github.com/svg/svgo/pull/592#issuecomment-244979133.
For cleaning up empty <style/>s, the empty Atrules have to be cleaned up first, too.

lib/css-tools.js
strarsis: As a CSS AST node only belongs to one parent node (and its list), it would be nice being able to access the list of an item directly (e.g. `item.list`) - currently with the item its containing list has to be kept for removal later.

lib/css-tools.js
```diff
@@ -0,0 +1,131 @@
...
+
+function filterByPseudos(selectors, usePseudos) {
+  return selectors.filter(function(selector) {
+    var pseudoNodes = [];
+    for(var pseudo of selector.pseudos) {
+      pseudoNodes.push(pseudo.item.data);
+    }
+    var pseudoSelectorsStr = csstree.translate({
+      type: 'Selector',
+      children: new List().fromArray(pseudoNodes)
```
lahmatiy: Maybe?
```js
children: new List().fromArray(selector.pseudos.map(function(pseudo) {
  return pseudo.item.data;
})
```

test/plugins/inlineStyles.13.svg
```diff
@@ -0,0 +1,68 @@
...
+  </defs>
+  <rect width="100" height="100" class="blue"/>
+</svg>
+
+@@@
+
+<svg id="test" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 81.285 81.285">
+    <defs>
+        <style>
+            @charset &apos;UTF-8&apos;;@namespace svg url(http://www.w3.org/2000/svg);@import url(&apos;https://fonts.googleapis.com/css?family=Roboto&apos;);@font-face{font-family: SomeFont;src: local(&quot;Some Font&quot;), local(&quot;SomeFont&quot;), url(SomeFont.ttf);font-weight: bold}@viewport{zoom: 0.8;min-zoom: 0.4;max-zoom: 0.9}@keyframes identifier{0%{top: 0}50%{top: 30px;left: 20px}50%{top: 10px}100%{top: 0}}@page :first{margin: 1in}@supports (display: flex){.module{display: flex}}@document url(&apos;http://example.com/test.html&apos;){rect{stroke: red}}
```
lahmatiy: Do you know why space is going after `:` in declarations? Is it because of `parseValue: false` setting?

strarsis: Yes, this setting seems to subtly change how whitespaces are treated. With parseValue: false any custom/extra white space is preserved.

lib/css-tools.js
strarsis: Ideally it would be possible to get the containing list of an item so just the items need to be passed, currently the list of an item has to be stored with it together for removing itself from the list.

lib/css-tools.js
```diff
@@ -0,0 +1,129 @@
...
+    node.selector.children.each(function(selectorNode, selectorItem) {
+        var selector = {
+            item:    selectorItem,
+            atrule:  atrule,
+            rule:    rule,
+            pseudos: []
+        };
+
+        selectorNode.children.each(function(selectorChildNode, selectorChildItem, selectorChildList) {
+            if(selectorChildNode.type == 'PseudoClassSelector' ||
```
lahmatiy: `==` -> `===`

test/plugins/inlineStyles.13.svg
lahmatiy: Hm, probably we should fix it in csstree. For now, you can trim values on your side

strarsis: All style property fields are trimmed now in PR.

strarsis:
All tests are passing!

alexjlockwood:
Awesome to hear about the performance improvements! Going to try this out for myself right now. :D

alexjlockwood:
Hmm... first of all, I should mention that I'm working with a fork of this plugin right now, so this issue may or may not affect svgo. But I figured I'd point it out since I've forked the code and dependencies exactly and am still having trouble building in my local setup with the following error:

```
ERROR in ./~/css-tree/lib/syntax/default.js
Module not found: Error: Can't resolve '../../data' in '/Users/alockwood/ShapeShifter/node_modules/css-tree/lib/syntax'
 @ ./~/css-tree/lib/syntax/default.js 1:11-32
 @ ./~/css-tree/lib/index.js
 @ ./src/app/scripts/svgo/plugins/inlineStyles.ts
```

Any idea what might be going on here?

lahmatiy:
@alexjlockwood I suggest some kind of transpiler is used and it doesn't support for `require('./file.json')`.

strarsis:
I closed+reopened the PR for triggering a re-build, there was an issue with the coveralls web service.

strarsis:
@lahmatiy: I am not sure whether this is really worth the performance improvements, but one could try to pass the css-tree parsed selector tokens directly to css-select instead first translating it to a string and letting css-select parse it back to its own tokens (https://github.com/fb55/css-select/blob/e109d4b3668354542c806a6b652410f135d679ae/lib/compile.js#L33 and https://www.npmjs.com/package/css-what).

lahmatiy:
@strarsis It's not clear to me what you're asking for. Question about the performance gain? I don't think it's a bottleneck. And converting to `css-what` format is overengineering.

strarsis:
How can I further adjust plugin in order to facilitate merging it in?

GreLI:
So, is it about to complete? I'm ready to merge it, but need to have a look into it. (And probably to play a bit.)

It'd be nice if you adapt code style from the rest of SVGO code (like 4 spaces indentation, space between statement and condition, no table-like alignment and so on). That's not blocker though.

strarsis:
@lahmatiy: Would it be possible to release the current state of css-tree to npm so it can be used as package dependency?

lahmatiy:
@strarsis Done [1.0.0-alpha18](https://github.com/csstree/csstree/releases/tag/v1.0.0-alpha18)

lib/css-tools.js
```diff
@@ -0,0 +1,188 @@
+'use strict';
+
+var csstree = require('css-tree'),
+    stable = require('stable'),
+    specificity = require('csso/lib/restructure/prepare/specificity'),
+    List = require('css-tree/lib/utils/list');
```
strarsis: Changed here: https://github.com/svg/svgo/pull/592/commits/baa1ffe45b7fd5618ed1fc5a0ee1899b3879375d

package.json
```diff
@@ -52,7 +52,7 @@
     "colors": "~1.1.2",
     "css-select": "^1.3.0-rc0",
     "css-select-base-adapter": "^0.1.0",
-    "css-tree": "github:csstree/csstree#e9c024b02a38d98c9eb886526709e54655eee2e1",
+    "css-tree": "^1.0.0-alpha18",
```
lahmatiy: It's better to pin the version until `css-tree` in alpha stage.

lib/svgo/css-style-declaration.js
```diff
@@ -0,0 +1,241 @@
...
+        return;
+    }
+    this.hasSynced = true; // must be set here to prevent loop in setProperty(...)
+
+    if (!this.styleValue || this.styleValue.length === 0) {
+        return;
+    }
+    var inlineCssStr = this.styleValue;
+
+    var cssoDeclarations = {};
```
lahmatiy: I think `csso` prefix is redundant here and in other places.

package.json
strarsis: A version range for css-tree releases from -alpha1 to -alpha8?

lahmatiy: I meant to use a certain version, in this case:
```json
"css-tree": "1.0.0-alpha18"
```

strarsis:
There is a strange issue with the Travis test - the semicolon is removed in the expected test output in this PR branch, but Travis doesn't seem to pick it up, see https://github.com/travis-ci/travis-ci/issues/7563.

lahmatiy:
This is no `removeScriptElement.01` test in your brunch. It was added [few days ago](https://github.com/svg/svgo/commit/e4be4bbc100e2badc1d65d4ef1f30852aea13c7f). You need to merge upstream into your brunch.

strarsis:
@GreLI: Code style has been adjusted.

strarsis:
@GreLI: Merge?

Munter:
I'll chime in here as being very interested in getting this released. It seems IE11 in Win10 doesn't apply `<style>` in data-uri image svg's at all, leaving everthing mostly black. https://connect.microsoft.com/IE/feedback/details/1635483/ie-11-svgs-referenced-as-base64-text-on-imgs-dont-apply-embedded-style

Inlining the styles as individual svg attributes and as a `style`-attribute fixes it

alexjlockwood:
Any updates on this?

MattyBalaam:
@Munter It's slightly more complicated, if an SVG is under around 4k, then the style declarations will work in IE11, so to test a larger SVG is required.

strarsis:
@MattyBalaam: An extra plugin could be added that is able to enable/disable other svgo plugins, like the file size/length of input SVG. Or the inlineStyles plugin can have an option added for conditional style inlining.

package.json
```diff
@@ -48,12 +48,17 @@
...
+    "css-select": "^1.3.0-rc0",
+    "css-select-base-adapter": "^0.1.0",
+    "css-tree": "1.0.0-alpha18",
+    "csso": "^3.0.1",
+    "js-yaml": "~3.7.0",
     "mkdirp": "~0.5.1",
-    "csso": "~2.3.1"
+    "sax": "~1.2.1",
+    "stable": "^0.1.5",
+    "whet.extend": "~0.9.9"
```
alexjlockwood: what.extend was removed here: https://github.com/svg/svgo/commit/4495a34f55560b2bc05cf837ba9dd01ca20cd479

test/plugins/inlineStyles.15.svg
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" standalone="no"?>
```
alexjlockwood: FYI, I get some annoying warnings when I run this test. Not sure if it can be avoided.

```
WARN: 'Warning: Syntax error when trying to select 

html /deep/ [hidden]

WARN: 'Warning: Syntax error when trying to select 

html /deep/ [block]

, skipped. Error details: SyntaxError: Unmatched selector: /deep/ [block]'
```

test/plugins/inlineStyles.07.svg
```diff
@@ -0,0 +1,17 @@
...
+            fill: blue;
+        }
+    </style>
+    <rect width="100" height="100" class="blue red"/>
+</svg>
+
+@@@
+
+<svg id="test" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
+    <rect width="100" height="100" class="blue red" style="fill:red!important"/>
```
alexjlockwood: When I run the `convertStyleToAttrs` plugin on this SVG, I get:

```
<svg id="test" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
  <path class="blue red" fill="red!important" d="M0 0h100v100H0z"/>
</svg>
```

Maybe this is a bug that should be fixed in a separate pull request? You probably know better than me. WDYT?

test/plugins/inlineStyles.15.svg
strarsis: Yes, this is intentional, it skips selectors that cause problems instead failing.
I could add an option to the plugin for not showing these warnings.

package.json
strarsis: Should I merge this change into this PR branch?
Or will it be automatically merged when this PR branch is merged to master?

test/plugins/inlineStyles.07.svg
strarsis: Indeed, the [convertStyleToAttrs](https://github.com/svg/svgo/blob/master/plugins/convertStyleToAttrs.js) plugin.

elmpp:
about to wet myself...

strarsis:
@elmpp: In anticipation of this being merged? 😄

Goldenson:
Any progress on this?

richardadjogah:
Is this still planning on being launched? I would love to see this plugin become available.

nchase:
much like the last few commenters, I'd find this really useful...is there anything we can do to help get it into master?

GreLI:
It will be there, I just need some things to settle.

justrhysism:
> It will be there, I just need some things to settle.

Is there a rough timeline on this? If it's going to be a week or two, I can wait. Otherwise unfortunately I'm going to have to find an alternative solution somehow.

strarsis:
@justrhysism: You can already use this plugin though by installing svgo from this PR branch (instructions above).

My6UoT9:
Is there an option, to remove the "class" attribute? As the styles are inline or defined as attribute, the class is not required anymore. I tried removeAttrs plugin - but that didnt remove it.

strarsis:
@My6UoT9: Yes, by default this plugin should already remove any classes that matched, the class names in class attribute value and in the styles. For non matching class attributes and left over styles (that didn't match) you would need an extra plugin because this is out of the scope of this one.

My6UoT9:
@starsis I had misconfigured the removeAttrs plugin, now it did work. But most likely I have misconfigured your plugin also - as it did not remove the class attribute, while the styles in <def> did match the classes.
This are the options I used (directly modified in src file):

exports.params = {
    onlyMatchedOnce: false,
    removeMatchedSelectors: true,
    useMqs: ['', 'screen'],
    usePseudos: ['']
};
And maybe it helps I add the svg I tried to optimize.
[example.zip](https://github.com/svg/svgo/files/1287105/example.zip)

strarsis:
@My6UoT9: I made a test from it and added this feature to plugin.

alexjlockwood:
FYI, updating to csstree alpha 25 seems to cause some crashes now.

strarsis:
@alexjlockwood: What node version? It would be great if you could add an issue for this crash in https://github.com/csstree/csstree/issues.

theKashey:
Any chance to ship this plugin this year?
There are 2 plugins, which will just double profit of SVGO – this one, and #700, both unmerged :(

alexjlockwood:
@strarsis You know what, ignore me. I think I was using a cached version of csstree alpha 24. The crash no longer occurs with alpha 25. :)

alexjlockwood:
I am also wondering what the timeline is for this pull request. If there is anything I can do to help get this merged, please let me know. :)

strarsis:
👍

Goldenson:
Thank you 🙌 ❤️

alexjlockwood:
@strarsis @lahmatiy How should the csstree dep be updated in svgo in the future given that it is still in alpha? Any thoughts?

strarsis:
@alexjlockwood: Maybe a Pull Request (which triggers the built-in integration test so it catches possible issues before merging) from time to time? https://github.com/fb55/css-select does this with its dependencies using some automation tool.

test/plugins/inlineStyles.15.svg
strarsis: @alexjlockwood: The console warnings have been commented out in recent master commits.

GreLI: Yeah, I thought they're more related to dev then user.

GreLI:

moltar:
Not working correctly for the following SVG:

https://gist.githubusercontent.com/moltar/d4c09168515d4a539aedc105274481ae/raw/3fa7869356c1666089b1690e8f0887d2d169105f/test.svg

Running with:

```
svgo --multipass --pretty --enable=inlineStyles test.svg
```

```
$> svgo -v
1.0.5
```

strarsis:
@moltar: I just tried this input svg with svgo version 1.0.5 and found no issues, neither in svgo usage nor in visual differences, see https://gist.github.com/strarsis/2bc9a80bc0ce52dee53d41150a8252c2 .

moltar:
@strarsis But it had no effect... The `fill` property did not move from `<style>` block to the attribute on the `path`.

From `test.min.svg`:

```
<path class="st2" d="M62 40h-7v-3c0-1.1-.9-2-2-2h-6c-1.1 0-2 .9-2 2v3h-7c-1.7 0-3 1.3-3 3v16c0 1.7 1.3 3 3 3h24c1.7 0 3-1.3 3-3V43c0-1.7-1.3-3-3-3zm-15-3h6v3h-6v-3zm-9 5h24c.6 0 1 .4 1 1v5H37v-5c0-.6.4-1 1-1zm24 18H38c-.6 0-1-.4-1-1v-9h26v9c0 .6-.4 1-1 1z"/>
```

strarsis:
@moltar: You need to set the `onlyMatchedOnce` option of `inlineStyles` plugin to `false` (as shown in https://github.com/svg/svgo/pull/592#issuecomment-251956973).

moltar:
@strarsis I see, thank you for pointing that out!

brutzel:
I followed the discussion here for a while and have been using the plugin a few times already. Great work! But from my point of view the default value should be `false` for that option. I too had to look up how to set plugin options from the command line for the first time and I think this default value is not very intuitive, I would it expect it to be the default behaviour. I am aware this results in much larger files.

strarsis:
@brutzel: I agree.
@GreLI, @caub: Would it be an improvement to code the default for `onlyMatchedOnce` option to `true`?
IMHO inlining everything (and isolating the styles) is the more expected behaviour with this plugin.

caub:
you can't always inline everything without loss, example with `:hover`: https://jsfiddle.net/qwx02hgg/

strarsis:
@caub: Correct, non-inlineable styles aren't inlined anyway by the plugin.
Therefor the prefixIds plugin is suitable for further isolating these styles.

GreLI:
@strarsis Did you mean setting it to `false`? We've been there. I think, it's likely will be opposite to optimization. May be better naming or docs will help.

alexjlockwood:
I guess it depends whether the flag will result in an optimized file or not. On one hand, not inlining everything can save bytes by keeping common code in the style block. On the other hand, not inlining the style can have a big impact on plugins that will be run later on (i.e. you won't be able to merge paths or collapse groups if the elements declare classes).

strarsis:
I agree with @alexjlockwood. The other plugins use the styles of the element instead using something like [`Window.getComputedStyle()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) (which could be implemented using the new styles API (caching/performance will probably be the biggest issue here).

GreLI:
Yeah, I dreaming about that for years, but it's far from the current state. Massive rewriting is needed.

brutzel:
@alexjlockwood right, I think it depends on what you see as _optimization_. My use case for this plugin is to prepare SVG files for embedding in HTML source. If you do that, CSS classes on multiple SVG elements override each other if they have the same class names. For example Illustrator always uses the same CSS class names st0, st1 and so on for each file. This plugin solves that problem.
Maybe I had misunderstood the purpose of this plugin, but I think this would be a common use case - although it is not really _optimization_ in the sense of minifying SVG files. But maybe there is a more common use case for this plugin that justifies that default value.

strarsis:
@GreLI: Hm, could it make sense to rewrite svgo to use an existing DOM/selector library like `jsdom` as the fork [svgz](https://github.com/caub/svgz) is trying?

GreLI:
I think @caub has intended to do something like that.

caub:
Yes, I may try to start a branch on svgo for it, and we could tag it as @next on npm, for a future major version.

It takes times because most plugins must be rewritten

https://github.com/jsdom/jsdom has improved a lot its SVG support, but getComputedStyle is not working as well as browsers yet https://repl.it/@caub/DeterminedScalyCommands, we would need to fix it for svgo

strarsis:
TODO: `css-select` now also supports pseudo elements/states (https://github.com/fb55/css-select/pull/139).
This can make two steps in the `inlineStyles` plugin obsolete: https://github.com/svg/svgo/blob/e7b8a6ce7d4467dee981b43cf25fd01468b5ab03/plugins/inlineStyles.js#L91

PR #601, opened 2016-09-20
Fix compatibility issue with node v0.10.29

cravesoft:
for...of was first defined in ECMAScript 2015. This pull request replaces calls to for...of with calls to Map.prototype.forEach().

### DIFF
--- a/plugins/cleanupIDs.js
+++ b/plugins/cleanupIDs.js
@@ -120,10 +120,7 @@ exports.fn = function(data, params) {
         return data;
     }
 
-    for (var ID of referencesIDs) {
-        var key = ID[0],
-            references = ID[1];
-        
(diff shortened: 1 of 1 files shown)


GreLI:
The main reason of bumping dependencies to 0.12 is Map/Set, which are also not supported in 0.10. And I don't want to add polyfills useless for almost all.

cravesoft:
OK, thanks for your answer. We now ask our users to install the latest nodejs from the nodesource binary repository.

GreLI:
It's been reverted (but I managed keep the fix the original issue) in 8abdef7742e67d45ed39f677f8e808062563eb25.

cravesoft:
Thank you very much for keeping me informed. It does indeed work with node v0.10.29 now.

PR #608, opened 2016-09-30
feat/plugin/convertStyleTags

JonathanStoye:
So that is a plugin that removes style tags and adds their content to the elements via their classes according to the parameters as inline style or as attributes.

In addition I updated the testing part in the docs.

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -25,6 +25,7 @@ plugins:
   - cleanupAttrs
   - minifyStyles
   - convertStyleToAttrs
+  - convertStyleTags
   - cleanupIDs
   - removeRasterImages
   - removeUselessDefs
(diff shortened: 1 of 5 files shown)


GreLI:
Seems like it duplicates #592

JonathanStoye:
@GreLI: didn't looked into the pull requests first. We needed it, so I wrote it and thought I should wrap things up and make a PR here. 

So if you don't need the plugin, should I create a separate PR for the docs only?

katspaugh:
@strarsis are you still working on #592?

strarsis:
@katspaugh:
The plugin should also inline only when the selector matches once, otherwise not.

This part is not completely implemented yet in the [localStyles](https://github.com/svg/svgo/pull/592) plugin.
[juice](https://github.com/Automattic/juice/) is used internally because it already solved the inlining complexities and is well tested for this.
Adding an inline-only-if-selects-once options is discussed [here](https://github.com/Automattic/juice/issues/232).

In the current state, the plugin can be used for inlining everything by setting option "onlyOnceMatching" to false.

GreLI:
Certainly this is not fully-fledged plugin, so it can be accepted.

PR #610, opened 2016-10-08
Allow to force clean svg with style or script, excepting custom ids (in cleanupIDs)

Velenir:
When processing svg files that make use of [svg-stacks](http://xn--dahlstrm-t4a.net/svg/structure/svgstack-usage.html) or [view elements](https://betravis.github.io/icon-methods/svg-sprite-sheets#with-a-custom-view) cleanupIDs plugin would either leave the file untouched (when `<style>` is present) or delete all ids from `<view>` elements.

I have added **force** (false by default) parameter to **cleanupIDs** config that allows to ignore the presence of `<style>` and `<script>` and clean up anyway. In case of **force = true** the onus would be on the user not to reference elements by id in those tags. Or to include ids not to be cleaned in **except** parameter.

**except**, which can be a single id or an Array of ids, allows to keep user-specified ids from being removed or minified. This would make possible referencing them externally as part of `<view id="#id">` or an **svg-stack**. For example,

``

**except**ing ids in `<symbol>` would also allow to use **cleanupIDs** with such svg files (still need to disable **removeUselessDefs** though).

### DIFF
--- a/plugins/cleanupIDs.js
+++ b/plugins/cleanupIDs.js
@@ -9,7 +9,9 @@ exports.description = 'removes unused IDs and minifies used';
 exports.params = {
     remove: true,
     minify: true,
-    prefix: ''
+    prefix: '',
+    force: false,
+    except: []
 };
 
 var referencesProps = require('./_collections').referencesProps,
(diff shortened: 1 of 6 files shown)


GreLI:

PR #611, opened 2016-10-10
updating the docs on the testing section

JonathanStoye:
I tried to test my plugin with the old docs and it was not working so after I had a look into some other plugins I decided to update the docs here a little bit.


GreLI:
Nice gotcha! Thanks!

PR #612, opened 2016-10-12
feat: add attribute that can add title and description and the according aria attributes

joscha:
This adds a plugin that transforms a given SVG to make it accessible. It allows passing `title` and/or `desc` and references them via `aria-labelledby` in the SVG element.

``` svg
<svg>
...
</svg>
```

with the plugin and a setting of

``` js
{title: 'my-title', desc: 'my-desc'}
```

becomes:

``` svg
<svg aria-labelledby="desc-abc title-xyz">
    <title id="title-xyz">my-title</title>
    <desc id="desc-abc">my-desc</desc>
...
</svg>
```

More info, see:
- https://www.paciellogroup.com/blog/2013/12/using-aria-enhance-svg-accessibility/
- https://www.sitepoint.com/tips-accessible-svg/
- https://css-tricks.com/accessible-svgs/

### DIFF
--- a/test/plugins/_index.js
+++ b/test/plugins/_index.js
@@ -36,7 +36,9 @@ describe('plugins tests', function() {
                         plugins = {},
                         svgo;
 
-                    plugins[name] = (params) ? JSON.parse(params) : true;
+                    /* jshint evil:true */
+                    plugins[name] = (params) ? eval('var x = ' + params + '; x;') : true;
+                    /* jshint evil:false */
 
                     svgo = new SVGO({
                         full    : true,
--- a/plugins/addAriaLabels.js
+++ b/plugins/addAriaLabels.js
@@ -1,0 +1,68 @@
+'use strict';
+
+exports.type = 'perItem';
+
+exports.active = false;
+
+exports.description = 'Adds an aria-labelledby and a referenced <title> and/or <description> section';
+
+exports.params = {
+    title: '',
+    desc: ''
+};
+
+exports.fn = function addAriaLabels(item, params) {
+    if (item.isElem('svg')) {
+        var ids = [];
+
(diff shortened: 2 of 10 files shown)


test/plugins/_index.js
```diff
@@ -36,7 +36,7 @@ describe('plugins tests', function() {
                         plugins = {},
                         svgo;
 
-                    plugins[name] = (params) ? JSON.parse(params) : true;
+                    plugins[name] = (params) ? eval('var x = ' + params + '; x;') : true;
```
joscha: This change was necessary to allow function definitions in the passed attributes. The addAriaLabels plugin usually auto-generates a random ID to link title/desc to `aria-labelledby`, but with static fixtures this does not work unless the random part is made static as well.

joscha:
@GreLI is there anything I can do to get this merged?

joscha:
Hmm, I'd really like to get this in - is there anyone here that is willing to start a conversation about it?

GreLI:
Sorry, I'm rather busy lately. However, I have to say that your PR doesn't have anything related to the purpose of optimization SVGs. People mostly removing titles, etc. There is already too many plugins polluting the scope.

Fell free to use it on your own I just can't accept it since it's off the way.

joscha:
It does a very special optimization which is not related to size, correct,
but makes the SVG itself accessible when used within a web context (e.g. as
an inline icon for example), but I understand what you mean - too bad -
anyway this plugin and a few more related to using SVGs in an inline
context can be found here then:
https://bitbucket.org/atlassian/atlaskit/src/3eaaa649f0b644a36f451ceed82409dab9639abe/packages/ak-icon/bin/svgo/plugins/?at=master

Cheers,
Joscha

PR #614, opened 2016-10-14
Fixing #504. Jspm compatibility and static declaration of dependencies

### LINKED ISSUE #504: Please consider using static declaration of plugin dependencies.
Hey there,

We're working on getting JSPM's CSS loader all hooked up into PostCSS plugins. We would like to (easily) use CSSNano as our CSS minification plugin. CSSNano has a dependency on svgo.

Currently, JSPM does not have support for dynamic require statements (see here: https://github.com/jspm/registry/issues/221). As such, [this bit of code](https://github.com/svg/svgo/blob/master/lib/svgo/config.js#L99) is causing us some headache (discussed here: https://github.com/MeoMix/jspm-loader-css/pull/2)

It's possible to workaround this issue in a couple of ways. More than likely I'll end up forking svgo, creating static requires, and overriding the dependency, but that's not ideal nor very maintainable.

I can appreciate the appeal of using dynamic require statements, but I'd be very happy if you'd consider replacing it with the alternative.

Thanks

joeldenning:
This fixes #504. There are definitely pros and cons to doing this, and am interested to hear your thoughts on what is best for the svgo project. My opinion, of course, is that this the pros outweigh the cons, but I can see the other side too.

**Pros:**
1. This makes svgo compatible with JSPM. Up until now, anybody who uses jspm and depends on svgo (directly or indirectly) has to create a configuration file in the jspm registry for _each version of svgo_. @MeoMix did this for svgo@0.6.1, and now I [have a pull request to do so for 0.7.1](https://github.com/jspm/registry/pull/987). Merging this svgo pull request will make it so that there will be no need anymore to create a configuration file in the jspm registry for each version of svgo.
2. This will make upgrading svgo to es6 modules (instead of commonjs) easier. I don't know if there are plans to do that or not, but once you switch to `import`, `System.import()`, or `import()`, you can no longer synchronously load dynamic plugin code like you can with `require`. Instead, you can either import all of the plugins upfront (similar to what I've done in this pull request) or you can change the [preparePluginsArray](https://github.com/svg/svgo/compare/master...joeldenning:issue-504?expand=1#diff-8f58d3f133adabc0050efa4a28cc5457R104) function to be an asynchronous function that returns a System.import() promise.

**Cons:**
1. The `pluginMap` is sort of an annoying manifesty thing that introduces one more step to the process of writing a new svgo plugin.

### DIFF
--- a/lib/svgo/config.js
+++ b/lib/svgo/config.js
@@ -119,7 +119,7 @@ function preparePluginsArray(plugins) {
 
             } else {
               if (!pluginMap[key]) {
-                  throw new Error("Unknown plugin '" + key + "'");
+                  throw new Error('Unknown plugin "' + key + '"');
               }
 
               plugin = EXTEND({}, pluginMap[key]);
--- a/lib/svgo/config.js
+++ b/lib/svgo/config.js
@@ -5,6 +5,51 @@ var yaml = require('js-yaml');
 
 var EXTEND = require('whet.extend');
 
+var pluginMap = {
+    cleanupAttrs: require('../../plugins/cleanupAttrs.js'),
+    removeDoctype: require('../../plugins/removeDoctype.js'),
+    removeXMLProcInst: require('../../plugins/removeXMLProcInst.js'),
+    removeComments: require('../../plugins/removeComments.js'),
+    removeMetadata: require('../../plugins/removeMetadata.js'),
+    removeTitle: require('../../plugins/removeTitle.js'),
+    removeDesc: require('../../plugins/removeDesc.js'),
+    removeUselessDefs: require('../../plugins/removeUselessDefs.js'),
+    removeXMLNS: require('../../plugins/removeXMLNS.js'),
+    removeEditorsNSData: require('../../plugins/removeEditorsNSData.js'),
+    removeEmptyAttrs: require('../../plugins/removeEmptyAttrs.js'),
+    removeHiddenElems: require('../../plugins/removeHiddenElems.js'),
+    removeEmptyText: require('../../plugins/removeEmptyText.js'),
+    removeEmptyContainers: require('../../plugins/removeEmptyContainers.js'),
+    removeViewBox: require('../../plugins/removeViewBox.js'),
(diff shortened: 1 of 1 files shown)


lib/svgo/config.js
```diff
@@ -73,8 +118,11 @@ function preparePluginsArray(plugins) {
                 plugin = setupCustomPlugin(key, item[key]);
 
             } else {
+              if (!pluginMap[key]) {
+                  throw new Error("Unknown plugin '" + key + "'");
```
joeldenning: This is to maintain a similar behavior to what happens when you try to `require` a plugin that doesn't exist. Now that the plugins are not being `require`d in this function, this Error maintains backwards compatibility of what should happen for non-existent plugins.

lib/svgo/config.js
```diff
@@ -73,8 +118,11 @@ function preparePluginsArray(plugins) {
                 plugin = setupCustomPlugin(key, item[key]);
 
             } else {
+              if (!pluginMap[key]) {
+                  throw new Error('Unknown plugin "' + key + '"');
```
joeldenning: This is to maintain a similar behavior to what happens when you try to require a plugin that doesn't exist. Now that the plugins are not being required in this function, this Error maintains backwards compatibility of what should happen for non-existent plugins.

joeldenning:
Looks like the [build is failing](https://travis-ci.org/svg/svgo/jobs/167721725#L647) only in node 6. The tests passed but after the tests coverall failed and I'm not sure exactly why. Any ideas on why I'm getting `Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}` from the node 6 build?

GreLI:
That happens sometime. Dont't worry, it should be ok.

GreLI:
I don't like the implementation since it loads all plugins including non-used one. It'll certainly slow down svgo execution for all users which I cannot accept.

Also I cannot verify that it solves the issue. Is there a less kludgy solution?

meteorplus:
I really need this to load all the plugins so I can browserify this asap!

otherwise I will have to fork this and load all of them then publish it to npm and use it!

thanks

PR #625, opened 2016-11-23
Fix typo in README

s0meone:
Adding the plugin with this spelling to the config file caused a case sensitive file system to trip.

Took a while to figure this out ;)


GreLI:
Nice catch!

PR #633, opened 2016-12-13
browser compatible

caub:
- browser bundle
- promise api:

```js
new SVGO(svgoConfig).optim(svgStr)
.then(svgjs => {
	console.log('result:', svgjs.data)
})
.catch(console.error);
```


caub:
@GreLI could you check my PR, the travis tests fails, I'll try to fix, but I've added interesting things I think

[x] browser bundle
[x] promise api
[x] return svg and parsed svg: {svg, data}, it can be useful
[x] cleanupFontFamily plugin, some tools generate additional useless families like Pacifico-Regular,Pacifico,..

caub:
added here a commit to round viewBox to floatPrecision too, simplified removeViewBox plugin

Also we should use this to Math.round to round values:
```js
var p=3; console.time(1); var factor=Math.pow(10,p); for(var i=0;i<1e6;i++){ var x=Math.random(); Math.round(x*factor)/factor } console.timeEnd(1);
var p=3, re=/0+$/; console.time(1); for(var i=0;i<1e6;i++){ var x=Math.random(); x.toFixed(p).replace(re, '') } console.timeEnd(1);
```
it's faster than toFixed, and I can't think of a reason to keep leading zeros?

GreLI:
I'm afraid it's too much change to accept. I don't want to stick to webpack or any compiler to keep code base as simple, compact and effective as possible. Lazy plugins requiring when plugins are loaded only when needed also follows this goal.

Unfortunately, I don't know a graceful way to make it compatible with non-Node.js environment. Hopefully, there'll be a solution in future like new module system.

caub:
Do you accept if I clean it? this PR was really dirty

but for webpack, I don't especially love it myself too, but I'm sure many users would benefit if SVGO can run in browser without effort, like https://github.com/jakearchibald/svgomg too.
But webpack is just a dev dependency, I would just need to cange the .svgo.yml to .js or .json because yml doesn't work well with webpack

Would it be ok? and thanks for all the hard work

PR #640, opened 2017-01-09
Remove unused styles based on usage data

lahmatiy:
It's initial implementation of `removeUnusedStyles` plugin ([feature request](https://twitter.com/SaraSoueidan/status/817820467067293696)). Plugin collects usage data and removes styles from `<style>` having selectors that contain id, tag or class name never used in document. [CSSO usage data](https://github.com/css/csso#usage-data) filtering feature is using.

Plugin can be disabled through settings:

```json
{
    "removeUnusedStyles":false
}
```

or disable a part of filtering

```json
{
    "removeUnusedStyles": {
        "ids": false,
        "tags": false,
        "classes": true
    }
}
```

Issues for discussion:
- should this plugin to be merged with `minifyStyles` or separate plugin is better?
- default settings:
  - should be active by default?
  - should we enable or disable (like now) filter options
- should plugin detects dangerous situations when filtering might to be unsafe and warn about it (i.e. `<script>` elements or `on*` attributes in document)?
- naming


GreLI:
That's a very nice proposal!

Regarding your questions:
* Yes, both plugins are quite similar and using CSSO. This plugin will also minify styles, as I far as I see, even if the other one is disabled, so no sense in making them separate.
* I don't see a reason to not make it enabled by default, so yes, I think it should be enabled as well, as all current options.
* The only point is the presence of the scripts. Probably it worth to check if there are some `<script>` and disable in a such case. Also, add an option that can force removing anyway. Not sure about `on*` attributes. I think they could be skipped just to make things simpler.
* I don't see any problems with naming except `walk` functions.

lahmatiy:
@GreLI 
- Plugins are merged
- Enabled by default
- Doesn't remove unused styles when document has a <script> element or on* attribute. `usage.force` option to force removal
- Now it's ok. Feel free to make up a code style and related after merge ;)

lahmatiy:
Any comments? Why not to merge?

GreLI:

Issue #636, opened 2016-12-19
How to pass param to plugin through CLI


jkelstrup:
I can't seem to figure out how to pass a param to a plugin when using the command line.

Specifically I want to tell `convertColors` to set `currentColor` to `true`.

The only way I found to do this is to create a config-file. But I'm hoping that there's a simpler inline-way of doing it?

elrumordelaluz:
@jkelstrup looking at the `coa` code seems that there's a way to pass a _config_ String `JSON` formatted. Something like (tested and it works for me):
```
svgo -i test.svg -o test-out.svg --config '{ "plugins": [ { "convertColors": { "currentColor": true } }] }'
````

I think could be added into CLI helps.

jkelstrup:
Thanks a lot @elrumordelaluz ! This might not be the prettiest of CLI-interactions, but it does exactly what I was looking for!

I'm gonna close this issue. If others could find it useful to see this improved, I think a new issue should be opened for that.

elrumordelaluz:
yes @jkelstrup, absolutely agree about
> might not be the prettiest

and

> new issue should be opened for that.

evilPaprika:
@elrumordelaluz 
Inlined json config is no longer suppored in svgo@2. What should i do?

TrySound:
@evilPaprika Use svgo.config.js

Issue #598, opened 2016-09-14
Request: disable all plugins, and enabling only specific ones


waldyrious:
This could be implemented with an option (one that says that we only want to use the explicitly enabled plugins), but I think it would make more sense to be a preset (#312).

For now, the workaround that I've found is to create an empty config file, and use it in conjunction with `--enable`:

`echo "plugins:" > c && svgo --config=c --enable=<plugin_name> test.svg`

kevva:
I would also like this. Sometimes I don't want *all* the optimizations.

GreLI:
There is config with `full: true` option exists for such a purpose.

waldyrious:
I see, that info can be found in [docs/how-it-works/en.md#1-config](https://github.com/svg/svgo/blob/master/docs/how-it-works/en.md#1-config). However, that still requires the creation of a config file, so it doesn't change the current situation much:

`echo "full: true" > c && svgo --config=c --enable=<plugin_name> test.svg`

It would be nice to be able to do this in a single command.

(by the way, `full: true` is very unintuitive; I'd never guess it did that if I saw a config file with it, and hadn't read the "how it works" document previously)

GreLI:
@waldyrious, you can pass config in a form of JSON in cli:
```bash
svgo --config='{"full":true}' --enable=<plugin_name> test.svg
```
but yeah, I agree it's clumsy. Suggestions are welcome!

waldyrious:
Uh. I was pretty sure I had tried passing a config string, but maybe the syntax was wrong and the error message led me to think that the problem was that it expected a file. I'll do some more experiments and will get back to you.

waldyrious:
I guess I just took the documentation's word and didn't even try a string xD

I would suggest the documentation to be changed from

    --config=CONFIG : Config file to extend or replace default

to

    --config=CONFIG : Configuration file (or string) to extend or replace the default

And maybe renaming the "full:true" feature to something else, since if you hadn't told me, I'd have assumed it did the opposite...

alinnert:
[...]

PR #641, opened 2017-01-12
Update CLI help

elrumordelaluz:
Add the missed possibility to pass a `JSON` string to `--config` argument

solves #636


waldyrious:
Nicely done, @elrumordelaluz! This would also address #598 :)

Issue #494, opened 2016-02-08
Map IDs to Classes


infn8:
is it possible to add a plugin that would remove the id attr from an element and map it to a class with an optional prefix and / or postfix?
#### For example

```
- mapIDsToClasses:
    prefix: 'svg-'
    postfix: ''
```

would map:

```
<path id="arrow" d="M50 514.26l-25 25 25 25z"/>
```

to:

```
<path class="svg-arrow" d="M50 514.26l-25 25 25 25z"/>
```
#### Additionally:

This:

```
<rect id="cool-rect" class="cool-things" x="13" fill="#FFFFFF" width="4" height="30"/>
```

would map to:

```
<rect class="cool-things svg-cool-rect" x="13" fill="#FFFFFF" width="4" height="30"/>
```

My reason is that Group and path names in Adobe Illustrator come out as IDs when you save as an SVG but if you insert the resulting SVG into an HTML doc more than once it will invalidate the document because it will have more than one element with the same ID 

Thoughts?

GreLI:
Well, it's totally possilble. Couple of caveats though:

1. `cleanupIDs` will remove unused IDs. So, the plugin should run before it, or `cleanupIDs` needs to be disabled, if these IDs aren't being referenced. If they are, references should also be updated of course.
2. SVGO cannot parse CSS for now. So, replacing IDs there requires some additional managing.

May be it's better just to update `cleanupIDs` to work in a such manner through option.

infn8:
It would for sure need to run before ```cleanupIDs``` but after it does run ```cleanupIDs``` would have no IDs to cleanup.  

Re: CSS, I was thinking a little more brute force in that each element would only care about itself.
```
class="
{any pre existing value of class attribute}
{optional space if class attribute was not already empty}
{prefix}
{value of id attribute}
{postfix}
"
``` 
Then remove id attribute and that's it.  No need to parse CSS.  But if we can add elegance or intelligently extend another function that would be boss.  Adding it as a param to ```cleanupIDs``` would ensure it ran first!  

Thoughts?

GreLI:
No doubts about `class` content. I was talking about selectors in possible `<style>` declarations that can use IDs.

infn8:
Great point!  I had not thought of that.  It could just be a documented shortfall of that option.  Are there any other features you're considering that might need, or benefit from, CSS parsing?  

This doesn't seem like the cost benefit is there with this one feature on its own.

GreLI:
[...]

Issue #564, opened 2016-07-06
How to include custom, local, plugin


elidupuis:
Is there a way to add custom plugins to SVGO? I have a small plugin I'd like to include as a build step in a project.

For example, I'd like to run `svgo` with a command line option pointing to a directory that contains custom plugins. Or point to said directory in my `.svgo.yml` configuration file. Is this possible?

GreLI:
No, it's not possible for now.

gampleman:
@GreLI would you be happy for a contribution adding this ability?

GreLI:
Yes

gampleman:
So looking at the source I think something like this is already supported. When I pass in a configuration looking like this:

```
{ 
  plugins: [
    { 
     custom: { // this could be any name except the already registered ones
        type: 'full', // full, perItem or perItemReverse
        description: 'My custom plugin',
        params: {}, // some arbitrary data
        fn: function(data, params) {
          // custom plugin code goes here
          return data;
        }
     }
  ]
}```

I get to execute my code in a custom manner. 

It seems that the only limitation is:

- Only supports inline code, so you have to invoke SVGO via the API rather than using command line or YAML.

elidupuis:
Oh sweet! Thanks for looking into this @gampleman!

In my opinion it would be ideal to have a config (or command line) option that points to a local directory of custom plugins that following the conventions of the plugins bundled inside SVGO. This directory would essentially be merged with bundled plugins. 

I peeked at the source when I created this issue but didn't have time to investigate further.

PR #648, opened 2017-01-19
Convert Ids to Classes

baldwmic:
Initial implementation of #494 

Please let me know if there are improvements to make. Perhaps there are more edge cases to handle and/or tests to write.

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -58,6 +58,7 @@ plugins:
   - addClassesToSVGElement
   - removeStyleElement
   - addAttributesToSVGElement
+  - convertIdsToClasses
 
 # configure the indent (default 4 spaces) used by `--pretty` here:
 #
--- a/plugins/convertIdsToClasses.js
+++ b/plugins/convertIdsToClasses.js
@@ -1,0 +1,34 @@
+'use strict';
+
+exports.type = 'perItem';
+
+exports.active = false;
+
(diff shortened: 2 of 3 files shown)


GreLI:
There are many considerations, but the main question is: what does it have with optimization? I can't accept it, if it's out of scope of SVGO's purpose. You can freely use it by yourself, of course.

baldwmic:
Thanks for taking a look @GreLI 

## Motivation
The purpose of submitting this PR is to clean up id attributes in SVG files by converting those id attributes to class attributes. If you include an SVG file with an id attribute more than once into an HTML file, that file has become invalidated because an id attribute needs to be unique. 

## Application
Design tools like [Sketch](https://www.sketchapp.com/) and [Adobe Illustrator](www.adobe.com/illustrator) automatically generate SVG files when exporting the designs. When a designer exports these SVG files, the design layers in the application append an id attribute of that layer name to the SVG. Therefore, the idea is to use `svgo` to clean up or optimize SVG files automatically generated by design tools for use in web applications. #494 is an example of this.

## Future Work
Let me know if this is something you are interested in pursuing. I understand if it does not align with the goals of `svgo` and we can close the PR.

## Gratitude
Thank you for your work on `svgo`! We are using this fork at my company, and it was easy to get started with `svgo` thanks to the good documentation. My only advice is that it would be nice to be able to write a command line plugin without needing to fork the repo as described in #564

Issue #406, opened 2015-07-20
convert shape to path - add circle, ellipse


onlymega:
See working example here: https://github.com/mailzwj/sstp/blob/master/svg-shape-to-path.js

GreLI:
Circle and ellipse hasn't been added because they're have less bytes in a form of `circle` and `ellipse` tags instead of `path`.

onlymega:
But when there are a lot of circles it is better to convert them to path and merge. You can add ellipse and shapes conversion and and add a flag for those who want convert them.

GreLI:
It may be so, but there is no such a logic for now. It requires sophisticated algo that doesn't fit in current design. May be in future, when there'll be much better architecture (it's in early planning stage).

qbaty:
Agree to @onlymega , I try to use js to draw svg instead of use svg as image. 
I want my svg is all made by Paths, not circle or ellipse.

SnooHD:
@onlymega would also like to see this implemented.
Maybe as an optional argument which is off by default?

GreLI:
Such an option was added in v1.0.0.

Issue #362, opened 2015-05-19
can`t convert Circle and Ellipse


purplebamboo:
[ > ] convert some basic shapes to path
https://github.com/svg/svgo/blob/master/plugins/convertShapeToPath.js

i can not find the code to trans shapes like  Circle and Ellipse.

why?

GreLI:
Because svgo it optimizer and generally circles and ellipses are shorter as their own tags. I know the only exception with the transformed (rotated) ellipse

Issue #65, opened 2012-12-07
Replace circular paths with <circle>'s


premasagar:
When exported as SVG, Illustrator expresses circles as full-blown paths. These paths could be more succinctly expressed as simple `<circle>` elements.

konclave:
+1

GreLI:
Actually, you can note that the path segment like ``a5,5 0 1,0 10,0 5,5 0 1,0-10,0`` has shorter length then ``<circle cx="10" cy="10" r="5"/>``. So, there is no profit in making it circle. Moreover, you can make it even shorter without notable difference in a way like ``a5,5 0 1,0 .1,0``

johan:
Strange apples and oranges comparison.

```
<path d="a5,5 0 1,0 10,0 5,5 0 1,0-10,0"/>
<circle cx="10" cy="10" r="5"/>
<path d="a5,5 0 1,0 .1,0"/>
```

Size isn't the only metric by which this cleanup is useful, though; if you want to hand edit the file to tweak things, a `<circle>` is orders of magnitude more hackable than the arc syntax.

GreLI:
Well, it's unlikely that you'll find ‘arcto’ (‘a’) command in the wild. More likely that you find arcs within complex paths (that's why I compare only the command) expressed with bezier curves which are written with 6 coordinates pairs. Converting these curves to arcs gives much more, especially when one arc is expressed with few curves.
I have a [prototype](https://github.com/GreLI/svgo/tree/arcs) of such an optimization, but it's imperfect (rounding errors make effect) and a little bit outdated. (Also, beware of force pushes.) Currently, I'm buzy with fixing main bugs of svgo to ensure its correct work, and I've almost finished with it.

waldyrious:
> if you want to hand edit the file to tweak things, a <circle> is orders of magnitude more hackable
:+1: I'd prefer the circle markup too, or at least have that as an option.

GreLI:
FYI version 0.6 of svgo recognizes arcs in curves and converts them correspondingly. Arcs path commands like `a5.07 5.07 0 1 0-2.8-9.72 5.07 5.07 0 0 0 2.8 9.72` (with precision=2) in your example are shorter as an part of an path then a separate `<circle>`.

PR #652, opened 2017-01-24
added convertCircleToPath plugin

gopeter:
This is basically just the code from @dnewcome that he've tried to merge long time ago (#368).

This feature was asked a few times (#629, #65, #406, #362, ...), so I thought you could add it as a plugin. Especially because the icon-font-trend get's replaced by the just-SVG-trend:

- https://css-tricks.com/icon-fonts-vs-svg/
- https://www.sitepoint.com/icon-fonts-vs-svg-debate/
- https://blog.prototypr.io/align-svg-icons-to-text-and-say-goodbye-to-font-icons-d44b3d7b26b4#.8mi0fg1rd
- https://cloudfour.com/thinks/seriously-dont-use-icon-fonts/

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -38,6 +38,7 @@ plugins:
   - cleanupEnableBackground
   - removeHiddenElems
   - removeEmptyText
+  - convertCircleToPath
   - convertShapeToPath
   - moveElemsAttrsToGroup
   - moveGroupAttrsToElems
--- a/plugins/convertCircleToPath.js
+++ b/plugins/convertCircleToPath.js
@@ -22,12 +22,12 @@ exports.fn = function(item) {
         item.hasAttr('cy') &&
         item.hasAttr('r')
     ) {
+
         var cx = item.attr('cx').value,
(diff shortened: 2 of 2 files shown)


gopeter:
Oh, maybe we should disable this plugin by default :-)

GreLI:
Why not just to add it to `convertShapeToPath`? As disabled by default option.

gopeter:
Because you already told @dnewcome that you don't want to include `circleToPath` conversions into `convertShapeToPath` because it doesn't optimize SVG code but makes the files bigger. 

To quote you: *Paths are ok, they are just written with more bytes which is opposite to purpose of optimization.*

That's why I thought that this should go into an extra plugin.

GreLI:
Well, circles *may* be optimized, *if* resulting path will merged with another path. So the option does make sense. It's just not guaranteed, that's why I preferred no to spend time writing it.

gopeter:
OK, so I'll add it to `convertShapeToPath` and you would be fine with it? :)

Issue #638, opened 2016-12-22
removeUselessStrokeAndFill plugin doesn't remove invisible paths


caub:
I have this config:
```js
{multipass:true, floatPrecision:2, plugins: [
	{mergePaths: false}, 
	{convertShapeToPath: false}, 
	{removeUselessStrokeAndFill: true}
]}
```
shouldn't this clean an invisible path like: 
`<svg viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" fill-rule="evenodd" clip-rule="evenodd" stroke-linejoin="round" stroke-miterlimit="1.41"><path fill="none" d="M-107.52-76.8h407.25V32.84h-407.25z"/></svg>`

It doesn't

elrumordelaluz:
@caub the fact that there is a `<path>` with attribute `fill` with the value `none` doesn't means is useless, since `fill=none` is useful to show the `stroke` and so on.
I think because of this, the author (@deepsweet) checks if there is an attribute `fill` but with value different than `none` [here](https://github.com/svg/svgo/blob/master/plugins/removeUselessStrokeAndFill.js#L38)

caub:
ok, thanks

PR #655, opened 2017-01-27
additional option for removeUselessFillStroke, to actually remove inv…

caub:
…isible shapes

Tell me if it makes sense

#638

### DIFF
--- a/plugins/removeUselessStrokeAndFill.js
+++ b/plugins/removeUselessStrokeAndFill.js
@@ -9,14 +9,14 @@ exports.description = 'removes useless stroke and fill attributes';
 exports.params = {
     stroke: true,
     fill: true,
-    removeNone: false
+    removeNone: false,
+    hasStyleOrScript: false
 };
 
 var shape = require('./_collections').elemsGroups.shape,
     regStrokeProps = /^stroke/,
     regFillProps = /^fill-/,
-    styleOrScript = ['style', 'script'],
-    hasStyleOrScript = false;
+    styleOrScript = ['style', 'script'];
 
 /**
  * Remove useless stroke and fill attrs.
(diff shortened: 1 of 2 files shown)


GreLI:
It looks like an idea for a separate plugin.

GreLI:
Also better results may be achieved using `computedAttr` which considers inheritance.

caub:
I know, but I use fill and stroke that already use computedAttr, I did it in the same plugin for perf, avoiding to do the same thing (item.computedAttr('stroke'), !item.computedAttr('fill', 'none')) in another plugin

caub:
note: https://github.com/svg/svgo/blob/master/plugins/removeUselessStrokeAndFill.js#L18 is really bad, because this module is required, then this global variable can be changed when running SVGO on one svg, then will keep that value for a next svg

caub:
@GreLI by putting hasStyleOrScript in the config, it allows to make sure a new one is there when running a new instance of SVGO

https://github.com/svg/svgo/pull/655/commits/bcd908691f41b9e300e499d25967d3bb3e2a453f#diff-2b80c20a2dbfb8e48fe3a26d26111ddaR13

it could be possible to make this plugin in full mode, instead of per element, but I think that's fine like that

alexjlockwood:
I think @caub is right that the usage of global variables may be problematic elsewhere as well. For example, in the "removeUselessDefs" plugin, a global 'defs' variable is reassigned various times throughout the plugin's execution. When the plugin exits, the variable will continue to hold a reference to its current value. In this case, I don't think there will be any behavior changes... however, it does seem like at the very least there will be a temporary memory leak (i.e. the referenced element won't be garbage collected until the next time SVGO is run).

Link: https://github.com/svg/svgo/blob/master/plugins/removeUselessDefs.js#L10

GreLI:
In `removeUselessDefs` global variable is used for recursion because of external function. It always will have actual value, so nothing to worry there.

PR #657, opened 2017-01-29
Cleanup layers IDs in SVG exported from Adobe Illustrator

danielbayley:
Layer IDs exported from Adobe Illustrator are corrected, allowing .class or #id attributes to be specified as layer names.

Signed-off-by: Daniel Bayley <daniel.bayley@me.com>

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -25,6 +25,7 @@ plugins:
   - cleanupAttrs
   - minifyStyles
   - convertStyleToAttrs
+  - cleanupAiLayerIDs
   - cleanupIDs
   - removeRasterImages
   - removeUselessDefs
--- a/plugins/cleanupAiLayerIDs.js
+++ b/plugins/cleanupAiLayerIDs.js
@@ -1,0 +1,40 @@
+'use strict';
+
+exports.type = 'perItem';
+
+exports.active = true;
+
+exports.description = 'cleanup .classes or #ids in layer names exported from Adobe Illustrator';
+
+/**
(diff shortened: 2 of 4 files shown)


GreLI:
I've never saw `_x23_` or `_x2E_` in real SVGs. The code looks really hackish to merge it.

danielbayley:
@GreLI The "hackish"ness is what gets introduced by [surprise] Adobe! Both tests pass though.

If you specify a `.class` (i.e. a layer name beginning with a `.` or an `#id` [layer name beginning with a `#`]), then once that SVG is exported from Illustrator, for some reason these become `_x2E_` instead of `.`, and `_x23_` in place of `#`…

The idea here is that it's useful to be able to specify which attribute you are targeting (with CSS say) just by naming the layers accordingly.

I suppose you could take the view as with #648, and say that this possibly falls outside of the scope of SVGO if it focussed _purely_ on optimisation, but it is useful to be able to add plugins that solve problems like this into the build step, so I guess this is another argument for #564… along with other ideas for useful plugins such as adding an `<?xml-stylesheet` instead of inline `<style>`s, and/or adding `xlink`s to SVGs.

baldwmic:
@danielbayley yeah i could see this being quite useful for developers working with SVGs exported from Adobe Illustrator.

It would be great if we could write plugins like #564 describes that could hook into the `svgo` functionality without needing to fork the entire repo. if someone made progress on #564 that might help our PR's find a home that makes more sense

PR #664, opened 2017-02-17
removeViewBox: Handle more forms

i80and:
The spec says that each term in viewBox is "separated by whitespace and/or a comma", but the regex here does not allow a comma.

This updated regex allows a comma. It is, however, still more strict that the spec, in that it requires a space (AFTER the comma, if it exists).


GreLI:
Done in faabeff4f4b12527e334c4b338ae7eef5a41aca2.

PR #666, opened 2017-02-21
Adds a clip functionality

lordnox:
For a project of mine I need just parts of a larger svg image, I did implement the functionality as a plugin for svgo.

I'm happy to share the bit of work I did by taking the https://github.com/w8r/martinez algorithm and applying it to svgo's plugin system.

Thanks for svgo!


GreLI:
Thanks for your interest in SVGO! However, this plugin is about very specific and limited image manipulation, so its quite out of scope of SVGO. Thus I cannot accept it.

Issue #563, opened 2016-07-01
Inline things in defs referenced only once 


kirilloid:
Since ids are mangled by default, they hardly can be referenced reliably outside.
So it's worth inlining if we detect they're used only once (after removing invisible and other parts).
Such references might happen pretty often even in hand-written code.

GreLI:
Yep, but I don't understand your point. What's next?

kirilloid:
Sorry, I don't understand what you don't understand.
I'm suggesting a feature for the tool.

themadcreator:
Consider this example @GreLI.

This defs/mask/use structure is typical of some graphics programs output like Illustrator or Sketch.

With IDs:
```svg
<svg width="315" height="92" viewBox="0 0 315 92" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <defs>
        <rect id="a" x="7" y="5" width="301" height="31" rx="3"/>
        <mask id="m" x="0" y="0" width="301" height="31" fill="#fff">
            <use xlink:href="#a"/>
        </mask>
    </defs>
    <g fill="none" fill-rule="evenodd" opacity=".1">
        <g stroke="#FFF">
            <g>
                <g opacity=".8">
                    <g>
                        <use mask="url(#m)" stroke-width="4" opacity=".4" xlink:href="#a"/>
                    </g>
                </g>
            </g>
        </g>
    </g>
</svg>
```

Optimized by inlining referenced element:
```svg
<svg width="315" height="92" viewBox="0 0 315 92" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <g fill="none" fill-rule="evenodd" opacity=".1">
        <g stroke="#FFF" opacity=".8">
            <rect x="7" y="5" width="301" height="31" rx="3" stroke-width="4" opacity=".4"/>
        </g>
    </g>
</svg>
```

steadicat:
@themadcreator I like this idea a lot, but are you sure your example is correct? I believe Sketch uses the shape-with-mask pattern to recreate the stroke-position: inside setting which is not available in SVG. The mask is supposed to cover the part of the stroke that falls outside the shape. By removing the mask you are breaking this and effectively doubling the width stroke.

Here's a screenshot of the before and after:

![image](https://cloud.githubusercontent.com/assets/6031/20940127/830b096e-bba6-11e6-8caf-66fa3e534eb7.png)

[...]

PR #670, opened 2017-02-26
Convert <use> tags. By default only defs with a single <use>.

nikparo:
I ran into exported svg's that use a lot of unnecessary defs, and decided to take matters into my own hands...

This plugin:
- loops through the svg to first find all `<defs>` element definitions. It then finds all `<use>` items.
(In hindsight, this should have been done in a single loop.)
- Filters out the element definitions that are only used once (by default, but optional)
- Converts `<use>` to `<g>` and inserts a clone of the definition.

This plugin does not remove definitions or collapse the resulting groups, leaving that for `removeUselessDefs` and `collapseGroups`.

I believe this plugin would cover the request in https://github.com/svg/svgo/issues/563

### DIFF
--- a/plugins/convertUse.js
+++ b/plugins/convertUse.js
@@ -1,0 +1,153 @@
+'use strict';
+
+exports.type = 'full';
+
+exports.active = true;
+
+exports.description = 'Converts <use> links with their <defs> counterparts';
+
+exports.params = {
+    convertAll: false
+}
+
+/**
+ * Converts <use> tags with the elements they link to.
+ *
+ * @param {Object} item current iteration item
+ * @param {Object} params plugin params
+ *
+ * @author Nik Paro
+ */
+exports.fn = function(data, params) {
+    
+    /**
+     * Find definitions among items
+     *
+     * @param {Object} items current iteration item
+     * @param {Boolean} isDef
+     * @return {Object} output definitions by id
+     */
+    function findDefs(items, isDef) {
+        isDef = (isDef==true);
+        
+        var defs = {};
+        
+        for (var i=0; i < items.content.length; i++) {
+            var item = items.content[i];
+            
+            if (item.isElem('defs')) {
+                isDef = true;
+            }
+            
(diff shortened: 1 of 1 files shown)


nikparo:
May have created the request a bit too early, noticed that there are some problems with at least masks and clip paths.

GreLI:
There are several caveats in your code.

alexjlockwood:
I don't think this PR handles CSS cascades correctly. According to the SVG spec, CSS cascade is applied to the referenced elements, not the (conceptually) cloned elements that make up the shadow DOM. What I think you would have to do is:

1. Run some sort of inlineStyles plugin (i.e. #592) on the SVG so that all of the elements that are descendants of a `<defs>` element are assigned their styled attributes.
2. Then run this plugin (i.e. find the `<use>` tags and replace them with the deeply cloned elements).
3. Make sure that any property-inheritance-related plugins are run *after* step (2).

Also I would take a look at the [SVG spec](https://www.w3.org/TR/SVG/struct.html#UseElement) on the `<use>` and `<defs>` tags since I also saw some things that were missing. For example, the plugin doesn't handle cases where `x`, `y`, `width`, and/or `height` attributes are specified on the `<use>` tag, etc.

plugins/convertUse.js
```diff
@@ -0,0 +1,167 @@
...
+exports.description = 'Converts <use> links with their <defs> counterparts';
+
+exports.params = {
+    convertAll: false
+};
+
+/**
+ * Converts <use> tags with the elements they link to.
+ *
+ * @param {Object} item current iteration item
```
alexjlockwood: Pretty sure this comment is wrong since the plugin type is 'full'.

GreLI:
Well, SVGO doesn't handle CSS almost at all.

strarsis:
A Window.getComputedStyle() implementation would be useful here. It is quite similar to inlineStyles plugin and style related API, an extra lookup table or property would be required because it would be too slow otherwise.

PR #683, opened 2017-03-13
Convert classes to inline styles plugin

dpanchenko:


GreLI:
Since this plugin doesn't correctly parse CSS, it can break images in unexpected ways. You may be interested in the effort being done in #592.

PR #686, opened 2017-03-22
Plugin for adding dimensions from viewbox

nym:
This is a simple plugin that allows you to add height and width dimensions based off the viewbox. It was made to deal with browser issues with rendering SVGs by default with 0x0 dimensions. This can be solved alternatively by giving the SVG a height / width through CSS. Setting a width/height in the SVG may not be desired.

Definitely open to thoughts and improvements! This was based off other plugins, removeDimensions and addAttributesToSVGElement.

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -53,6 +53,7 @@ plugins:
   - removeTitle
   - removeDesc
   - removeDimensions
+  - addDimensions
   - removeAttrs
   - removeElementsByAttr
   - addClassesToSVGElement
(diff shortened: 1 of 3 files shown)


caub:
it's fun because I'm doing the opposite, removing width and height when there's a viewBox. Width/height are really not necessary, like for jpeg/png, it's stylable in css

nym:
@caub Totally width/height are not typically necessary. In my case, it allowed for just updating assets instead of deploying a new version of the app I'm working on.  Had I not been time constrained, I would have used removeDimensions ;-)

caub:
@nym for doing something simple like this, I would do that: 
https://github.com/caub/svgo/blob/browser/test/browser.js#L38-L45

I've added .js2svg in lib/svgo.js then I can modify the result tree, remove attributes on svg, then stringify again

```
result: <svg xmlns="http://www.w3.org/2000/svg" width="120" height="120" viewBox="0 0 120.45 120"><sty...
result2: <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 120.45 120"><sty...
```
and result2 is 25 chars shorter as expected

GreLI:
Thanks for commitment! However, this plugin has nothing to do with optimizing and thus I cannot accept it.

Unfortunately, SVGO is already somewhat bloated with such plugins, so it needs to focus on its main purpose to keep things as simple as possible. Feel free to use it by yourself, the plugin system was done to be easy to extend like this. Best wishes!

PR #687, opened 2017-03-23
Fix typo

Yeti-or:
@caub thanks for catching it
@GreLI Why tests were green?


GreLI:
This plugin isn't covered. Actually it's quite outdated and I'm going to remove it.

Issue #395, opened 2015-07-03
Security vulnerabilities


liukanghui:
I've noticed that this project doesn't do any sanitization to the inputted svg—i.e. if someone throws a ```<script>``` tag in the ```<svg></svg>``` it won't be stripped.

If I'm using the node module version of this project, where users input their own svgs which other users can later load in their browser, are there security vulnerabilities? My guess would be yes, since a user could just add a ```</svg>``` into the evaluated textarea and then put any script they wanted in after that.

GreLI:
Well, svgo isn't a security tool. You can write a plugin that removes script tags.

PR #688, opened 2017-03-23
add removeScriptElement plugin

pklingem:
addresses https://github.com/svg/svgo/issues/395

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -57,6 +57,7 @@ plugins:
   - removeElementsByAttr
   - addClassesToSVGElement
   - removeStyleElement
+  - removeScriptElement
   - addAttributesToSVGElement
 
 # configure the indent (default 4 spaces) used by `--pretty` here:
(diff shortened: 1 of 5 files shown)


GreLI:
Thanks!

pklingem:
@GreLI any chance you could cut a release for us now that this has been merged?

pklingem:
thanks for the quick response, we'll lock to the sha for the time being.

auralon:
Any ideas when this will get pushed in a release to npm?

GreLI:
As soon as it be ready. Unfortunately, there a lot of things needs to be managed, but things must get easier after that.

auralon:
OK, thanks for the super quick response!  This functionality would be very useful.  It also seems to work OK, from my initial tests.  

Although, at present I am manually setting the `exports.active = false` to `exports.active = true` in the plugin file.  This is the first time I've used `svgo`... how might I otherwise set a plugin to active?

pklingem:
`npm install --save svg/svgo`
`svgo --enable=removeScriptElement -i file -o -`

auralon:
Thanks, @pklingem :)  ...  the version that I'm getting through npm doesn't have the plugin.  I've manually downloaded the plugin and I can get it to work in a node script, but only by manually setting the `exports.active` to true.

That cli command you've given me is useful, but if possible, I'd like to know how it to do the same within a node app.

GreLI:
SVGO v1.0.0 with this plugin has been released now.

PR #692, opened 2017-03-28
Fix CLI stdin

pklingem:
at some point between v0.7.2 and current master, piping into the svgo binary broke.  This PR fixes that.

### DIFF
--- a/lib/svgo/coa.js
+++ b/lib/svgo/coa.js
@@ -146,7 +146,7 @@ module.exports = require('coa').Cmd()
             !opts.string &&
             !opts.stdin &&
             !opts.folder &&
-            Boolean(process.stdin.isTTY)
+            process.platform === 'win32' || process.stdin.isTTY
         ) return this.usage();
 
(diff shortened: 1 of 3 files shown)


lib/svgo/coa.js
```diff
@@ -146,7 +146,7 @@ module.exports = require('coa').Cmd()
             !opts.string &&
             !opts.stdin &&
             !opts.folder &&
-            process.stdin.isTTY !== false
+            Boolean(process.stdin.isTTY)
```
GreLI: This was done for a reason. In Windows CLI `isTTY` is `undefined`, that why it's being comparing to `false` here. Another solution is needed.

pklingem:
@GreLI updated, not sure if that's an acceptable solution, but a conversation starter at least.

arendvw:
On linux process.stdin.isTTY is also undefined when it's piped. 
To check if stdin.isTTY is true seems to solve it here.

process.stdin.isTTY === true 

Seems to be a better solution.

GreLI:
Made as @arendvw suggested in e3ddf1d96ab88beb9f496a122dee6ebcb4a0c907:159.

PR #696, opened 2017-04-02
remove useless switch groups

caub:
I don't use arrow functions, because ugligyJS doesn't like them sadly

### DIFF
--- a/test/plugins/_index.js
+++ b/test/plugins/_index.js
@@ -24,10 +24,10 @@ describe('plugins tests', function() {
 
             file = PATH.resolve(__dirname, file);
 
-            it(name + '.' + index, function(done) {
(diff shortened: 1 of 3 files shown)


test/plugins/_index.js
```diff
@@ -24,10 +24,10 @@ describe('plugins tests', function() {
 
             file = PATH.resolve(__dirname, file);
 
-            it(name + '.' + index, function(done) {
-
-                FS.readFile(file, 'utf8', function(err, data) {
+            it(name + '.' + index, function() {
 
+                return readFile(file)
```
caub: could simply do readFileSync too

GreLI:
Fine. Thanks!

Issue #694, opened 2017-03-29
Removing all .toFixed and .round'ing properly


caub:
Do you agree if I replace all usages of .toFixed for rounding values?

example: `+num.toFixed(params.floatPrecision)`

should be something like `Math.round(num*params.precision)/params.precision` (where precision is Math.pow(10, floatPrecision))

in terms of perf, run this to see the diff
```js
var p=3; console.time(1); var factor=Math.pow(10,p); for(var i=0;i<1e6;i++){ var x=Math.random(); Math.round(x*factor)/factor } console.timeEnd(1);
var p=3, re=/0+$/; console.time(2); for(var i=0;i<1e6;i++){ var x=Math.random(); x.toFixed(p).replace(re, '') } console.timeEnd(2);
var p=3; console.time(3); for(var i=0;i<1e6;i++){ var x=Math.random(); +x.toFixed(p) } console.timeEnd(3);
```

GreLI:
1. What for?
2. Are you sure that there never will be numbers like 12.00000000000001?

caub:
yes, try `Math.round(12.00000000000001 * 100)/100`

caub:
Other benchmark, with the smartRound function:

```js
function roundFixed(value, precision) {
    var tolerance = +Math.pow(.1, precision).toFixed(precision)
    var rounded = +value.toFixed(precision - 1);

    return +Math.abs(rounded - value) >= tolerance ?
            +value.toFixed(precision) :
            rounded;

}


function round(value, precision) {
    var tolerance = 1 / precision,
		p = precision / 10,
        roundedMore = Math.round(value*p) / p;
    return Math.abs(roundedMore - value) >= tolerance ?
                Math.round(value * precision) / precision :
                roundedMore;
}

console.time(1);
for (var i=0; i<1e6; i++){
	round(Math.random(), 1000);
}
console.timeEnd(1);


console.time(2);
for (var i=0; i<1e6; i++){
	roundFixed(Math.random(), 3);
}
console.timeEnd(2);
```

caub:
```js
/**
 * Rounding function 
 * - precision = number of decimals (ex: 2 or 3)
 */
function round(value, precisionDigits, error/*=Math.pow(.1, precision)*/) {
    if (precisionDigits==0) return Math.round(value);
    var rounded = +value.toFixed(precisionDigits),
        roundedMore = +value.toFixed(precisionDigits - 1);
[...]

PR #697, opened 2017-04-02
Round

caub:
attempt for #694, it's not urgent, I'm not 100% sure it's good

last commit, I had to change a little test data, 
```
 -    <path d="M60 0a5 5 0 1 0 .001 10.001A5 5 0 0 0 60 0z"/>
 +    <path d="M60 0a5 5 0 1 0 .001 10.001 5 5 0 0 0 0-10.001z"/>
```
the 2 svgs look the same though https://jsfiddle.net/crl/4gzwfyvg/


plugins/convertPathData.js
```diff
@@ -408,34 +400,34 @@ function filters(path, params) {
             // Rounding relative coordinates, taking in account accummulating error
             // to get closer to absolute coordinates. Sum of rounded value remains same:
             // l .25 3 .25 2 .25 3 .25 2 -> l .3 3 .2 2 .3 3 .2 2
-            if (precision !== false) {
```
caub: removed that check, I don't think that's useful, precision=false would keep raw values?

GreLI: Yep, that's right/

test/plugins/convertPathData.14.svg
```diff
@@ -17,7 +17,7 @@
     <path d="M15 10a5 5 0 1 1 5-5l-5 5"/>
     <path d="M41.008.102a5.006 5.006 0 0 1 3.849 3.705"/>
     <path d="M7.234 19.474a5 5 0 0 1-4.933-8.683"/>
-    <path d="M60 0a5 5 0 1 0 .001 10.001A5 5 0 0 0 60 0z"/>
+    <path d="M60 0a5 5 0 1 0 .001 10.001 5 5 0 0 0 0-10.001z"/>
```
GreLI: 4 characters longer, it's not good. Here is some problem.

caub: ok right, but I think it's random, I'll try to find a case where it's the opposite, round shorter and toFixed longer

caub:
I understand the fear of floating points error with Math.round(x * 1000) / 1000, but it can't happen

.999 factor in smartRound is a bit hacky, but it's just to pass the very rare case that round badly like `Math.round(9.995 * 100)/100 /* 9.99 instead of 1*/`  because 9.995*100 is 999.499999999

caub:
Closing, and just commit the removed globals in cleanup PR

PR #698, opened 2017-04-03
Allow element separator to be changed in config

### LINKED ISSUE #557: Remove xmlns:xlink
How do I remove: `xmlns:xlink="http://www.w3.org/1999/xlink"` from the SVG?

As using: `{removeAttrs: { attrs: ['svg:xmlns', 'svg:xmlns:xlink'] }},` doesn't work... only the `xmlns` is removed.

mikestreety:
This resolves #557 as you can pass in a different character for element seperation (e.g. `^`

### DIFF
--- a/plugins/removeAttrs.js
+++ b/plugins/removeAttrs.js
@@ -1,7 +1,5 @@
 'use strict';
 
-var ELEM_SEP = ':';
-
 exports.type = 'perItem';
 
 exports.active = false;
@@ -9,11 +7,15 @@ exports.active = false;
 exports.description = 'removes specified attributes';
 
 exports.params = {
+    elem_sep: ':',
     attrs: []
 };
 
 /**
  * Remove attributes
+ *
+ * @param elem_sep
+ *   format: string
  *
  * @param attrs:
  *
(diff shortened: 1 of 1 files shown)


GreLI:
Linter fails because the param name is not in camel case. Also, it's not the best name for the param.

mikestreety:
Ah appologies @GreLI - I just took the existing variable and dropped it in.

Suggestions on param name?

GreLI:
Ideally, I'd change syntax to something like `'path[fill]'` with taking attribute name in brackets like in CSS.

mikestreety:
Yeah, that would be great. Unfortunately that's beyond me :( I could correct the camel case then raise an issue for this to be an "upgrade"?

GreLI:
That's not a great improvement, probably, I'll do it myself before releasing new version.

mikestreety:
Ok thanks. 

To add a bit of context, Figma drops `figma:type="canvas"` and `xmlns:figma="http://www.figma.com/figma/ns"` attributes all over the SVG with invalidates it for anything but figma!

GreLI:
Adopted and merged in 645908923bb163e177023ca21fe04a380fade126.

Issue #674, opened 2017-03-02
Option to generate unique IDs?


pluma:
This is a feature request.

There's a plugin for minifying IDs but this is basically useless when inlining SVGs in HTML because it is guaranteed to cause problems (i.e. when inlining at least two SVGs that contain IDs you're guaranteed to have a conflict on ID "a").

It would be very useful if SVGO allowed prefixing IDs with a configurable key. This would allow tooling like `react-svg-loader` to generate a unique prefix and pass it to SVGO to guarantee the IDs don't break when the SVGs are inlined.

A problem is that the IDs can be referenced with `xlink:href` attributes and `url()` values but this is already addressed by the `cleanupIDs` plugin.

GreLI:
Actually, cleanupIDs has `prefix` option (works only with minifying).

Kraxxis:
I, too, need this feature.

the cleanupIDs `prefix` option is not a solution to this problem.  Consider two svg files both of which have a defined id `"id1"`.  When using the two separate svgs inline, these two ids will conflict, since they exist in the same document.  Using the `prefix` option only moves this problem from having two conflicting `"id1"` ids to having two conflicting `"prefix_id1"` ids.

Emasoft:
The prefix should be the name of the file. For example: 

 IconOfACar.svg -> iconofacar_id1
IconOfABeer.svg -> iconofabeer_id1

pluma:
Being able to supply the prefix should suffice. Seems more versatile than having SVGO pick one automagically.

Emasoft:
I don't think that is versatile for batch processing. I usually process thousands of svg files with a script at every site update, and I surely don't want to manually enter a prefix for each one.

strarsis:
Similar issue?: https://github.com/svg/svgo/issues/673

OwenMelbz:
Is there any movement on this?

As much as it sounds a dick move, I think you need to listen to the general consensus here, this is an issue that many users are having when using inlined SVGs when passed through SVGO - for example we even have this issue using Sketch SVGO Compressor plugin, we end up having 10 SVGs with `#a` that causes masking issues so have to manually go through the file and change them, which is not feasible for a fully automated system.

As @Emasoft has suggested, I think the only sensible solution would be to slugify the filename and use that as the id prefix, turn this on as a default, which means every user that has an automated workflow or has tools that then use svgo will be unaffected, unless they pass in an extra option to disable it something like `--disable-auto-id-prefix`

[...]

Issue #673, opened 2017-03-01
Set unique class names plugin


strarsis:
Not all styles can be inlined, like those in At-Rules.
For even further scoping the left over CSS one could rename all selectors in a way that makes collisions with ones from other SVGs highly unlikely - by prefixing them with a manually passed string, by the basename of the SVG file or randomly generated.

strarsis:
Plugin PR ready: https://github.com/svg/svgo/pull/700

Issue #651, opened 2017-01-21
Inline gradient/fills


strarsis:
It may be useful for an addtional plugin to inline fills (notably gradients), too. 
SVGs exported from Illustrator extensively use them and 
they could collide with gradients of same ID from other SVGs.

strarsis:
it would be much easier if the new css selection features from https://github.com/svg/svgo/pull/592 could be used, querying for (linear)Gradient element by ID from element fill attribute/style.

strarsis:
The easiest workaround now is using the [prefixIds](https://github.com/svg/svgo/pull/700) plugin that became available in svgo.

xseman:
Has this issue been resolved, or do I still need to use the workaround? Is there anything I can help with?

Issue #623, opened 2016-11-15
Option to pass the name of the svg being processed as a param into a plugin


harrisjose:
Is there any way to pass in the filename of the current svg into a plugin ?

The reason why i need this is because a lot of our svgs are exported from illustrator and have style tags with class names like `.st0 {} .st1 {}`. This causes a issues when we use a lot of svgs in a single page, as the styles get overriden.

An easy way to solve this would be to replace a class like `.st0` to `.svgName0`..which is why i need the file name. I'd also appreciate other ideas on solving this.

bomberstudios:
+1. I recently added an auto prefixing feature for `defs` to https://github.com/BohemianCoding/svgo-compressor, based on the SVG filename, and was considering sending a pull request to make it "native" to SVGO.

harrisjose:
Even though I mostly don't have this issue these days ( we just started using attributes or inline styles when exporting svgs ), I'd still love to see this being natively supported in svgo.

strarsis:
PR: https://github.com/svg/svgo/pull/700

bomberstudios:
👏👏👏👏

harrisjose:
Closing this since prefixIds plugin by @strarsis takes care of what I was originally trying to achieve.

PR #700, opened 2017-04-05
Add prefixIds plugin

strarsis:
This PR adds the prefixIds plugin which prefixes all identifiers (IDs, class names) and 
also updates the url(...) references.

Additionally support for extra information is added that can be optionally used by the plugins, 
currently containing the path to the input SVG file or that mode is text (when text input is used).

Related issues: https://github.com/svg/svgo/issues/494 https://github.com/svg/svgo/issues/595 https://github.com/svg/svgo/issues/623 https://github.com/svg/svgo/issues/673 https://github.com/svg/svgo/issues/674
Similar approach would be inlining all styles (see the [inlineStyles](https://github.com/svg/svgo/pull/592) plugin) - 
though this is not possible with elements that are `<used/>` (see https://github.com/svg/svgo/issues/651).
This prefixing approach could be viewed as a "polyfill" (partially) for [scoped styles](http://caniuse.com/#feat=style-scoped).


lib/svgo.js
```diff
@@ -20,7 +20,7 @@ var SVGO = module.exports = function(config) {
     this.config = CONFIG(config);
 };
 
-SVGO.prototype.optimize = function(svgstr) {
+SVGO.prototype.optimize = function(svgstr, info) {
```
caub: why you don't pass that `info` data in the config?

plugins/prefixIds.js
```diff
@@ -0,0 +1,151 @@
...
+
+exports.params = {
+    delim: '__'
+};
+
+exports.description = 'prefix IDs';
+
+
+var path      = require('path'),
+    csstree   = require('css-tree'),
```
caub: those 3 deps are really not necessary

just put the regex you need, and for tree walking it's just 
 ```js
function walk(node, cb) {
  cb(node);
  if (node.content) node.content.forEach(walk)
}
```

Comment (strarsis) [inline plugins/prefixIds.js]: Yes, I use perItem now.

File: lib/svgo.js (hunk 1 above)
Comment (strarsis) [inline lib/svgo.js]: The svgo architecture handles the config globally, it is harder to add per-plugin extra config than adding an extra argument to each plugin. Also this extra information is not config but rather environment/state instead.

Comment (caub) [inline lib/svgo.js]: global config can be passed at the root of the config https://github.com/svg/svgo/blob/master/test/plugins/_index.js#L42

Comment (caub) [inline lib/svgo.js]: good point, things like floatPrecision too, should be global more easily instead of spread like now https://github.com/svg/svgo/blob/master/lib/svgo/config.js#L36 I'll see if it could be done with your way

**Comment by strarsis** [comment]
Currently all attribute values are checked for url(...) / ID references. 
Maybe this can be made more efficient by only checking attributes that can actually reference to other elements, as defined in https://github.com/svg/svgo/blob/master/plugins/_collections.js#L2286.

File: plugins/prefixIds.js [hunk 3]
```diff
@@ -0,0 +1,145 @@
+'use strict';
+
+exports.type = 'perItem';
+
+exports.active = true;
```
caub: also I wouldn"t enable it by default

lib/svgo.js
strarsis: Yes, the file path must be somehow injected into the config object for each file (or set to empty/false/null for a string input): https://github.com/strarsis/svgo/blob/67e5a3ef10e349bded15bc9f508596252c2884be/lib/svgo/coa.js#L437

plugins/prefixIds.js
```diff
@@ -0,0 +1,171 @@
...
+exports.active = false;
+
+exports.params = {
+    delim: '__'
+};
+
+exports.description = 'prefix IDs';
+
+
+var path      = require('path'),
```
caub: hmm, forgot to say sorry, but on my side, I'm more using svgo from browser, and I try to remove those, node-only deps. But it shouldn't be hard to pass a prefix string, not a path

lib/svgo.js
caub: yea sorry again to be annoying, after that I stop:), but I think the best is to have an optional `prefix` property at the root of the config and pass it to any plugin that needs it like https://github.com/svg/svgo/blob/master/lib/svgo/config.js#L34, it's more consistent with current code. The other option would be to refactor more and do  `var svgo = require('svgo'); svgo.optimize(svgStr, options);` passing whole config at each call, possibly in the form `{common: {floatPrecision:.., prefix: '..'}, plugins: [...]}` so each plugin receive common and its plugin specific options. This plugin is interesting, even if it doesn't really minify, I use svgo to make png/jpg previews of them, and I do that in iframes to avoid those conflicts, so your way could make it work too.

plugins/prefixIds.js
strarsis: How are you running svgo in the browser? When you use string input method, path module is not needed.

caub: very simple https://github.com/caub/svgo/tree/browser (see dist/ for the bundle), and practical because svgo run fast

 webpack will complain about require('fs') or require('path'), maybe it's possible, but didn't check yet

strarsis: @caub: webpack should be able to shim or at least ignore fs/path modules: https://github.com/webpack/node-libs-browser

strarsis:
@GreLI: Code style adjusted, this can be merged now.

strarsis:
@GreLI: Merge?

GreLI:
Sorry for the delay. I'll take a look eventually.

kyleknighted:
@GreLI Any chance this could be reviewed? Would love to be able to use this feature!

funklos:
@GreLI merge and release would be so great! We're struggling heavily with the deployment of a huge svg library and that's exactly what we need! 
Resolving conflicting id's manually is really a pain..

strarsis:
@funklos: You can install it already from this PR branch:
````
$ npm install github:strarsis/svgo#prefixIds
````

**Comment by funklos** [comment]
@strarsis sure, thanks for the suggestion.. we'll do this to perform some tests. 

But I really can't reconfigure multiple CI systems and change multiple package.json to enable this for all our projects and revert all the changes once this branch is merged. Especially since I'm the one enforcing the usage of only stable packages.. ;)

**Comment by vzaidman** [comment]
guys, why doesnt the plugin enabled when i pass

```
plugins: [
    {convertPathData: false},
    {prefixIds: true}
  ]
```

**Comment by caub** [comment]
try `{prefixIds: {active: true}}`

**Comment by vzaidman** [comment]
looks like the plugin should be added to `svgo\.svgo.yml`

**Comment by vzaidman** [comment]
can you add
```
if(extra && typeof opts.prefix === 'function'){
        prefix = opts.prefix(node, extra)
    }
else if (extra && extra.prefix && extra.path.length > 0) {......
```
so we can create our own prefix?

**Comment by strarsis** [comment]
Yes, the plugin isn't enabled by default. Does it make sense to have it always enabled?

**Comment by strarsis** [comment]
Travis build issue, re-opening PR...

**Comment by vzaidman** [comment]
im not sure what's the purpose of " .svgo.yml" but following the code, it only adds plugins that are in there.

didn't fully understand the question.

the usecase is that we have some svgs with same classes that override each other, so i want a special prefix for each file.

**Comment by vzaidman** [comment]
ok looks like its not useful in my use case unless we add some indication that we are in the same file...

**Comment by strarsis** [comment]
@vzaidman: So you got a set of files and you want to use the same prefix for that set of files?
The callback function could accept the filename/path as argument so it can decide what prefix to use?

**Comment by vzaidman** [comment]
sorry i confused the svg sprites plugin with svgo. its confusing to work on a library inside a library inside a library.
i just need the same prefix for all, so the fix with the function can be reverted. my mistake.

**Comment by caub** [comment]
you can also, for this kind of svg processing (not necessarily minification), do:
```js
const {JSDOM} = require('jsdom')
const svg = new JSDOM(svgStr).window.document.querySelector('svg');
svg.querySelectorAll('[id]').forEach(el => {
  const id = el.id, re = new RegExp(`#(${id})\b`);
  svg.querySelectorAll(`[fill="url(#${id})"], [xlink:href="#${id}"]`).forEach(el => {
    if (el.hasAttribute('fill')) el.setAttribute('fill', el.getAttribute('fill').replace(re, `#${prefix}${id}`));
    if (el.hasAttribute('xlink:href')) el.setAttribute('xlink:href', el.getAttribute('xlink:href').replace(re, `#${prefix}${id}`))
  });
  svg.querySelectorAll('style').forEach(s => { s.textContent=s.textContent.replace(re, `#${prefix}${id}`) })
})
```

strarsis:
The API additions from the [inlineStyles plugin PR](https://github.com/svg/svgo/pull/592) also allow to use `querySelectorAll(...)` and getting/setting style element content.

strarsis:
FWIW, I just noticed that Illustrator CC offers 'Unique' for 'Object IDs' in SVG export options.

funklos:
Jeah we also found this option but we also noticed that illustrator export is not suitable in every case.
In cooperation with our graphics department we decided to use sketch for SVG exports, since sketch is gaining more and more traction for reponsive screen design this was the logical consequence. 
So the combination of sketch with svgo for batch optimization on deployment is our way in the future..

strarsis:
Ready to merge!

lib/svgo.js
strarsis: Yes, a separate prefix config option is supported now (see https://github.com/svg/svgo/pull/700/files#diff-623dc50b643d62febf2af73ea1856b51R120).

strarsis:
👍 Yes!

theKashey:
Awaiting for the version bump!

chenxiaochun:
Very good!

TongDaDa:
I want to cancel the class prefix in the svg file, What should I do?

strarsis:
@TongDaDa: You can disable plugins you don't want to use.

PR #704, opened 2017-04-09
round viewBox too

caub:


caub:
is it ok to merge? it's useful to round a 25.509999999999 in viewBox

Issue #732, opened 2017-05-29
Plugin idea/request for filter cross browser compatibility


drschwabe:
Firefox and Chrome treat SVG filters (ID's) differently - the long and short of it is to ensure filters work in Firefox (otherwise anything that uses a filter disappears) you must create a class with a relative reference to the appropriate filter ID. 

For example, this doesn't work in Firefox but works in Chrome: 
```
<svg>
    <defs>
        <filter id="a" color-interpolation-filters="sRGB">
            <feBlend in2="BackgroundImage" mode="lighten"/>
        </filter>
    </defs>
    <path transform="matrix(123456789)" filter="url(#a)"/>
</svg>
```

This however, works in both Chrome and Firefox: 
```
<svg>
    <defs>
        <style>
          .filterA {
            filter:"url(.#a)"; 
          }          
        </style> 
        <filter id="a" color-interpolation-filters="sRGB">
            <feBlend in2="BackgroundImage" mode="lighten"/>
        </filter>
    </defs>
    <path transform="matrix(123456789)" class="filterA" />
</svg>
```

This may be specific to my workflow which involves creating the SVGs in Inkscape which generates filters in this way (first example is saved from Inkscape and run through SVGO, second example is the manual hack/fix). 

There is some discussion about the issue on SO: 
https://stackoverflow.com/questions/33676342/svg-filter-not-showing-up-in-firefox-working-fine-in-chrome

At any rate, it seems like my solution above could be automated with SVGO so I just wanted to put this out there; I don't have any immediate intentions to write a plugin myself but would certainly welcome any feedback or insight if you care to share.  Thanks.

GreLI:
That's strange, looks like you have a bad filter. It should work in SVG, otherwise file a bug. The topic on SO is about using filters in HTML–that's a different topic.

drschwabe:
To clarify context, this class/filter workaround is necessary when using including the SVG in an HTML via the img tag ie: 
```
<img src="/img/star.svg" />
```

GreLI:
Well, then it may be related to [SVG as an Image](https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_as_an_Image) restrictions. Still it could be a bug.

drschwabe:
[...]

PR #733, opened 2017-06-02
Add chainFilters plugin

strarsis:
This PR adds the chainFilters plugin which chains filter elements using CSS filters as 
workaround for some issues with specific browsers, closes https://github.com/svg/svgo/issues/732.
Furthermore a DOM walker is added which can be used for traversing the svgo DOM (by full mode plugins).
Currently some plugins and svgo core all use their own custom "monkeys" implementations for this.

If a filter is already assigned to an element using styles and the plugin also got a filter attribute to the same filter element, this plugin would still create a chain - this may be a feature to add - but this kind of corner case should be very rare, also the element shouldn't have a filter assigned by filter attribute _and_ styles in the first place. Also this plugin currently always assumes relative paths in fragment identifiers.


strarsis:
@drschwabe: You can already test this plugin using:
````
# using npm:
$ npm install github:strarsis/svgo#chain-filters

# using yarn
$ yarn add github:strarsis/svgo#chain-filters
````

**Comment by strarsis** [comment]
@GreLI: The filter bug is indeed a problem in Firefox - maybe this plugin should be enabled by default?
Currently the filter element is chained with the element using a CSS class with filter property with IRI.
Would it be better to convert the whole filter element into CSS? - Is that always possible?

**Comment by strarsis** [comment]
For cleanly and correctly implementing this, CSSStyleDeclaration interface from inlineStyles PR is required.

**Comment by strarsis** [comment]
@GreLI: Any chances that inlineStyles PR branch is merged so I can implement support for filter in inline styles?

**Comment by strarsis** [comment]
@GreLI: Conflicts fixed, ready to merge.

**Comment by strarsis** [comment]
@GreLI: Ready to merge.

**Comment by strarsis** [comment]
@GreLI: What can I do for further speeding up the review/merge process?

**Comment by GreLI** [comment]
I haven't made a decision yet. Not sure about the plugin. Need some time to consider.

**Comment by strarsis** [comment]
Arguments against merging this plugin into svgo:
- It doesn't minify or simplify, rather it is a polyfill for a Firefox issue.
- (When a plugin-as-package architecture has been implemented in svgo) This could also fit in a separate package.

Arguments for merging this plugin into svgo:
- This Firefox bug can be subtle and puzzling when suddenly some parts of the SVG are missing when viewed in Firefox. Just having this extra step will ensure this problem doesn't happen again for all SVGs processed by svgo.
- One doesn't necessarily finds this bug when searching for it (I encountered this bug in a svgo issue created by another user and found the exact same symptoms).
- This plugin can always be disabled or even removed when Firefox had this bug closed for long enough time.

**Comment by strarsis** [comment]
@TrySound: Does it make sense to add this plugin to `svgo`? Loading plugins from other packages (installed from `npm`) isn't supported yet by `svgo`.

**Comment by TrySound** [comment]
@strarsis It is supported. Just require the plugin and pass to plugins array.

**Comment by strarsis** [comment]
@TrySound: So should this plugin now be merged into svgo repository or rather provided as standalone plugin (e.g. via npm)?

**Comment by TrySound** [comment]
This looks more like polyfill than optimisation. So yes, this would probably look better as separate package.

File: lib/dom-walker.js [hunk 1]
```diff
@@ -0,0 +1,9 @@
+'use strict';		
+
+var walk = require('tree-walk');		
+
+var domWalker = walk(function(node) {
```
TrySound, line 5: I added traverse utility in lib/xast.js.

TrySound:
See

strarsis:
@TrySound: Thanks for the adjustment. I had not looked at this PR for some while. 

When I try this test SVG in Chrome and Firefox (both, regular and Developer Edition), the SVG is displayed correctly and equally:
https://gist.githubusercontent.com/strarsis/2c769b63b9573b9cf81faa9242adf4ec/raw/1571560b8ca1c895a52fb3669a4ad32615350993/chainfilters-test.svg
So does this mean that this issue had been eventually fixed in Firefox?

SethFalco:
Hey! Thanks for creating the PR.

> So does this mean that this issue had been eventually fixed in Firefox?

Indeed, it looks like the issue has already been resolved. Makes sense given the amount of time that's passed. ^-^'

PR #735, opened 2017-06-02
Allow datauri option on js api

vcanales:
This allows the user to mimic using the `--datauri` flag programatically.


GreLI:

PR #741, opened 2017-06-13
Add keepAriaAttrs to removeUnknownsAndDefaults

### LINKED ISSUE #684: Prevent SVG aria attributes from being removed
Is it possible with any combination of settings to prevent `aria-*` attributes from being remove from the `<svg>` tag when optimising? 

I have tried all options that seem relevant to this but have had no luck so far:
```
.pipe($.svgo({
    addAttributesToSVGElement: true,
    removeUnknownsAndDefaults: false,
    cleanupAttrs: false,
    removeEmptyAttrs: false
}))
```

This is quite important as the `aria-labelledby` and `aria-describedby` attributes allow screen readers to interpret the content of inline SVG's.

davidtheclark:
Closes #684.

Adds a `keepAriaAttrs` option to `removeUnknownsAndDefaults`, with a default value of `true`. And adds two cases to make sure it works.

### DIFF
--- a/plugins/removeUnknownsAndDefaults.js
+++ b/plugins/removeUnknownsAndDefaults.js
@@ -11,7 +11,8 @@ exports.params = {
     unknownAttrs: true,
     defaultAttrs: true,
     uselessOverrides: true,
-    keepDataAttrs: true
+    keepDataAttrs: true,
+    keepAriaAttrs: true
 };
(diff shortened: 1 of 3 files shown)


GreLI:
Nice, thanks!

GreLI:

PR #742, opened 2017-06-14
fix viewBox round

caub:
@GreLI I did a stupid mistake, sorry, added a test


plugins/cleanupNumericValues.js
```diff
@@ -41,8 +41,9 @@ exports.fn = function(item, params) {
 
         if (item.hasAttr('viewBox')) {
             var nums = item.attr('viewBox').value.split(/[ ,]/g);
-            item.attr('viewBox').value = nums.map(function(num) {
-                return +num.toFixed(floatPrecision);
+            item.attr('viewBox').value = nums.map(function(value) {
+                var num = +value;
```
caub: to play defensive, but a non number here would break the svg most likely

PR #752, opened 2017-06-30
Fix arcto sweep flag when a transformation matrix is applied

### LINKED ISSUE #708: Bad Arcto values when applying transform matrix
Original
```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36"><path d="M32 4a4 4 0 0 0-4-4H8a4 4 0 0 0-4 4v28a4 4 0 0 0 4 4h20a4 4 0 0 0 4-4V4z" fill="#888" transform="matrix(1 0 0 -1 0 36)"/></svg>
```

Optimized
```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36"><path d="M32 32a4 4 0 0 0-4 4H8a4 4 0 0 0-4-4V4a4 4 0 0 0 4-4h20a4 4 0 0 0 4 4v28z" fill="#888"/></svg>
```

Tested on svgo 0.7.2 with its default config, I narrowed it down to the `convertPathData` plugin but that's as far as I got.

JoshyPHP:
When a transformation matrix flips the coordinates horizontally or vertically (but not both) clockwise arcs need to become counterclockwise and vice-versa.

Fixes #708


GreLI:
Thanks for handling this!

GreLI:

PR #755, opened 2017-07-05
Updated colorNames according to CSS Color Module Level 4

ydaniv:
Mostly missing variants of gray/grey and added rebeccapurple.


ydaniv:
Should note that currently `colorsNames` contain values in short hex format, which currently goes against current API description and docs, which state that it will translate color names to long hex.

I'd gladly add a commit that fixes all of these to long format if there's no objection.

Yeti-or:
convertColors plugin choose what is shorter `rgb()` `hex` or `name`. So I think that docs are correct right now

ydaniv:
@Yeti-or yes, took me awhile to figure out the purpose of all the different collections (:
but the problem is that if you set the config to:
```
shorthex: false
shortname: false
```
You'll expect all hexes to be in long format, but they won't.

And besides, there are some valid color names from the specs that are missing.

ydaniv:
:+1:

PR #758, opened 2017-07-13
#757 shorthex false does not transform f00

### LINKED ISSUE #757: shofthex: false does not transform ff0000 to red
With default settings (which includes shortname: true):
<g color="#ff0000"/>
is optimized to:
<g color="red"/>

When setting:
{ "shorthex": false }

the same behavior is expected but the color remains:
<g color="#ff0000"/>

davidleston:
Fixes #757


GreLI:
Good catch!

GreLI:

Issue #759, opened 2017-07-13
hex color in upper case does not convert to shortname


davidleston:
With default settings:
<g color="#ff0000"/>
is converted to:
<g color="red"/>

#FF0000" should also convert to red.

PR #760, opened 2017-07-13
Closes svg/svgo#759. Upper case hex color now converts to shortname.

davidleston:


GreLI:
Thanks!

GreLI:

PR #780, opened 2017-08-28
Added plugin for focusable="false" attribute.

keeganstreet:
Internet Explorer and Edge (prior to version 14) make all embedded <svg> elements keyboard focusable by default, which is not consistent with the SVG specification. To prevent this behaviour, you can add `focusable="false"` to the `<svg>`.

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8090208/

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -59,6 +59,7 @@ plugins:
   - removeStyleElement
   - removeScriptElement
   - addAttributesToSVGElement
+  - addFocusableFalseAttribute
 
 # configure the indent (default 4 spaces) used by `--pretty` here:
 #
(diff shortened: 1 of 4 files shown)


GreLI:
I think it's not worth adding non-optimizing plugin only for rather outdated browser.

keeganstreet:
IE11 may be an outdated browser, but it is a component of the operating system in Windows 7, Windows 8.1, and Windows 10. And so unfortunately, supporting it is a reality for many websites until Microsoft end support of Windows 10 (mainstream end support date 10/13/2020).

I understand you don't want a plugin specifically for IE11 so I have opened PR #817 which updates the `addElementsToSVGElement` plugin to support attribute values, as well as attribute names.

PR #781, opened 2017-08-28
Added plugin for adding width and height attributes.

keeganstreet:
Add `width` and `height` attributes to an outer `<svg>` element.
Values are extracted from the `viewBox` attribute.

This allows the SVG to display at the correct size when rendered in a web browser without additional CSS.

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -53,6 +53,7 @@ plugins:
   - removeTitle
   - removeDesc
   - removeDimensions
+  - addDimensions
   - removeAttrs
   - removeElementsByAttr
   - addClassesToSVGElement
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Today we have:
 | [sortAttrs](https://github.com/svg/svgo/blob/master/plugins/sortAttrs.js) | sort element attributes for epic readability (disabled by default) |
 | [transformsWithOnePath](https://github.com/svg/svgo/blob/master/plugins/transformsWithOnePath.js) | apply transforms, crop by real width, center vertical alignment, and resize SVG with one Path inside (disabled by default) |
 | [removeDimensions](https://github.com/svg/svgo/blob/master/plugins/removeDimensions.js) | remove `width`/`height` attributes if `viewBox` is present (disabled by default) |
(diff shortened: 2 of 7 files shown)


GreLI:
There is already `addAttributesToSVGElement`. Perhaps, it's worth extending it.

keeganstreet:
Hi @GreLI, thanks for the reply. `addAttributesToSVGElement` currently only supports attribute names, not values. I would suggest updating this plugin to support attribute values. Do you want me to open a PR for that?

keeganstreet:
Hi @GreLI, thanks for the reply. I would urge you to reconsider, as I think this plugin adds some real value that cannot be achieved with the `addAttributesToSVGElement` plugin.

This plugin retrieves the width and height from the `viewBox` attribute. Then it adds these values as `width` and `height` attributes.

This is really valuable, because if an SVG without `width` and `height` attributes is embedded into a webpage, it will expand to 100% width of its container. If CSS fails to load due to a network error, the SVG will expand to fill 100% width of the browser window, resulting in a completely illegible page. Many sites, like github.com, css-tricks.com, nytimes.com, use `width` and `height` attributes on their SVGs.

This is a common problem because when SVGs are saved from Adobe Illustrator, they include a `viewBox` attribute, but not `height` or `width` attributes, so it would be good if SVGO makes it easier for web developers to address this problem.

GreLI:
The quick googling shows, that one have to turn off “responsive” option in Adobe Illustrator dialog to keep “width” and ”height” parameters: https://css-tricks.com/illustrator-to-svg/#post-247993

keeganstreet:
Yep right you are. I was thinking that svgo removed the height and width attributes but that is not the case. Thats OK then. Thanks!

dietergeerts:
@GreLI , how can `addAttributesToSVGElement` be used to transform the `title` attribute to another attribute of chose? We use the `removeTitle` plugin now, but we still need some identifier on the svg element, which should be the same as the title was. I can't see any plugin that can do this, and there is almost no docs on how to use the `addAttributesToSVGElement` plugin. Some extra docs would be helpful.

dietergeerts:
I think that when we update the following line, we could set the value of an attribute based on the current svg: https://github.com/svg/svgo/blob/02e0e416c8145da6770fde2174b977a83daea008/plugins/addAttributesToSVGElement.js#L70
By giving the option of OR being a value, OR being a function which will return a value based on the svg

PR #717, opened 2017-05-11
Update README.md

Crissov:
reorganize plugins by type
emphasize plugins that are active by default (for `--pretty` I guess)
explain `install -g`

PR #788, opened 2017-09-08
alphabetize plugins

CITguy:
Alphabetize plugins for quick reference.


GreLI:
There was a better attempt in #717. Unfortunately, there are conflicts.

Issue #730, opened 2017-05-23
This useless attributes aren't removed


crantisz:
This useless attributes aren't removed (from inkscape 0.92):
```
style="isolation:auto;mix-blend-mode:normal" 
overflow="visible"
solid-color="#000000"
paint-order="fill markers stroke"
```
and
`color="#000"` from `<g>`

Tested via svgomg: https://jakearchibald.github.io/svgomg/

PR #790, opened 2017-09-10
feat: add default value for overflow attribute

Delapouite:
Hello.

Adding the default value `visible` of `overflow`, let us remove this useless attribute set by inkScape, as mentionned in https://github.com/svg/svgo/issues/730

MDN ref: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow


GreLI:
The tests are written bad, but I'm not sure how default the value is, especially with SVG 1.

GreLI:
Yeah, the link is wrong, it's for CSS only. The right link is https://www.w3.org/TR/SVG11/masking.html#OverflowProperty and there are many pitfalls.

PR #805, opened 2017-10-06
Fix #176: Don't escape special character inside `<script>` or `<style>` tag.

### LINKED ISSUE #176: Don't encode quotes to entities in <script> and <style>
By default any double quotes `"` withing a `<script>` element are converted to `&quot;`, increasing the size of the resulting file.

bentinata:
I break `createText` return, since using ternary operator would be too long.


GreLI:
I'm afraid, it's unsafe. E.g. unescaped ampersand can create unintended entity which then will be parsed wrong.

bentinata:
How about not escaping `apos` and `quot`? My only concern is this breaking `@import` inside `<style>` tag.

PR #807, opened 2017-10-12
New plugin: classToInlineStyle

theKashey:
So, all the things explained in test case. It just moves styles from <style> to inline style of some element.
The main question, you may ask - WHY?

Everything because CSP. You just cant manage styles inside SVG if you have strict CSP rules.
So solution is simple - first move styles from def to element inline style. Next use convertStyleAttrs to move styles to the attributes.

As result - your SVG file is not CSP compatible.

PS: Was discussed with @GreLI two years ago :) Used in production by YandexMaps. But nobody knew about this plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Today we have:
 
 | Plugin | Description |
 | ------ | ----------- | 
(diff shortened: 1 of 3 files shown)


theKashey:
Look like this is a dummy duplicate of https://github.com/svg/svgo/pull/592

GreLI:
Yeah, just have merged those.

PR #811, opened 2017-10-18
Optimize repeated commands

ThomasdenH:
According to the SVG spec, repeated commands can be omitted. While this is not shorter if a space is needed in its place, the space can be omitted if the next number is negative. This optimization does exactly that. The reduction in file size is very small, in my tests much less than 1%, but it does help and it's a relatively small change.


GreLI:
Actually `convertPathData` already does this. However, it's cannot be done always, since it can break image. You almost never can omit a curve, lines depends on directions, etc. You change breaks tests, by the way.

PR #702, opened 2017-04-07
Convert ellipses circles

alexjlockwood:
The ability to convert circles and ellipses to paths has been requested a few times, for example in #652.

This pull request adds the ability to do just that.

LMK what you think!

PR #818, opened 2017-10-27
Convert circles/ellipses to paths

alexjlockwood:
The ability to convert circles and ellipses to paths has been requested a few times, for example in #652.

This pull request adds the ability to do just that.

LMK what you think!

See #702 for earlier work on this.

### DIFF
--- a/plugins/convertShapeToPath.js
+++ b/plugins/convertShapeToPath.js
@@ -98,6 +98,47 @@ exports.fn = function(item) {
 
         item.renameElem('path')
             .removeAttr('points');
-    }
+    } else if (item.isElem('circle')) {
 
+        var cx = +(item.attr('cx') || NONE).value;
+        var cy = +(item.attr('cy') || NONE).value;
+        var r = +(item.attr('r') || NONE).value;
+        if (isNaN(cx - cy + r)) {
+            return;
+        }
+        var pathData =
+            'M' + ' ' + cx  + ' ' + (cy - r) 
+                + 'A' + r + ' ' + r + ' 0 1 0 ' + cx + ' ' + (cy + r) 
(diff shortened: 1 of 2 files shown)


GreLI:
Issues from the original one are still present here:

- it should be optional
- there're errors in code, it doesn't work
- lack of tests.

alexjlockwood:
Fixed the errors, will add some tests later tonight!

@GreLi how do you suggest I make this optional?

GreLI:
Just an option for the plugin like `convertEllipses`.

alexjlockwood:
Would you prefer two options (one for circles and a second for ellipses)?

GreLI:
I think one is enough, may be shorten to like `convertArcs`, but if you see such a point, you're welcome.

alexjlockwood:
Cool, sounds good. I'll add some tests tonight when I get back from work.

alexjlockwood:
@GreLI I added a simple test case and fixed up the code. LMK what you think!

GreLI:
:

GreLI:

buisine:
@alexjlockwood 

Hi looklike ellipse element are not converted to paths

made test on commande line like : 

➜  svg svgo --enable=convertShapeToPath 1.svg

➜  svg svgo --enable=convertShapeToPath --config=conf.json --pretty 1.svg 

with config.json having { "convertArcs": true }

Can you help please ?

alexjlockwood:
@buisine I recommend filing a bug with the SVG you are trying to convert and the svgo version you are using.

buisine:
@alexjlockwood 
Thanks.

For now i edited plugin source : 
    exports.params = { convertArcs: true }; 

in convertShapeToPath.js
and it does the job !
thought my config is bad.

PR #829, opened 2017-11-02
Beminfo

JoKalliauer:
Sorry if I post it on a wrong place, please correct/move.

I wrote a [shell-script](https://github.com/JoKalliauer/cleanupSVG/blob/master/svg2validsvg.sh) to 
-make SVGs [W3C](https://validator.w3.org/#validate_by_upload)-Valid
-remove [librsvg](https://en.wikipedia.org/wiki/Librsvg)-Bugs
-remove some useless data

I tested it on several [files on Wikimedia](https://commons.wikimedia.org/wiki/Category:SVG_files_with_multiple_W3C-errors) see: [examples](https://commons.wikimedia.org/wiki/Special:ListFiles?limit=50&user=JoKalliauer)

I would like to add the tool to svgo, do I have to rewrite [the Shell-Script](https://github.com/JoKalliauer/cleanupSVG/blob/master/svg2validsvg.sh) in JavaScript?


GreLI:
Something odd is here.

PR #830, opened 2017-11-06
Add plugin to cleanup XML namespaces

### LINKED ISSUE #844: removeUnusedNS plugin doesn't remove unused namespaces declared on non-svg nodes
For example:

```
$ svgo --pretty -o - -s  '<svg 
      xmlns:android="http://schemas.android.com/apk/res/android" 
      xmlns="http://www.w3.org/2000/svg">
      <path xmlns:unused="http://unused-namespace.com"
          d="M0 0l100 100" />
  </svg>'
```

results in:

```
    <path xmlns:unused="http://unused-namespace.com" d="M0 0l100 100"/>
</svg>
```

I would expect that the unused namespace declared on the `path` element would be removed in this scenario.

thomsj:
Added plugin to cleanup (remove unused and minify used) XML namespaces. Replaced _removeUnusedNS_ plugin with new plugin (_cleanupNS_). Fixed bug in _cleanupIDs_ which would allow a generated minified ID to be the same as a preserved ID. Close #844


thomsj:
Close #844

thomsj:
Closing as I don't think this should be part of SVGO as XML namespaces are not specific to SVG, and as the code in the commit is very unclean.

PR #837, opened 2017-11-13
Option recursive

dartess:
Tried to add the option "recursive". Seems to work, but I have little experience in Node.js. In particular, I did not understand whether it is possible to create directories asynchronously and how to properly compile errors when creating directories ([here](https://github.com/svg/svgo/commit/b129b4f14a4ee7ef2bfd6aef1100d2d4d19bb506#diff-e354e2080d6fcfd5d48d6ef7785a920dR472)). 

And I think the tests are not complete and should be supplemented by the actual size of the files and folders before and after optimization. But in the tests I also have little experience, so I did it by analogy.

This all code need code review.


GreLI:

Issue #139, opened 2013-09-23
Removing viewBox breaks scaling <img> in IE9+


samuelvogel:
When displaying a SVG inside an `<img>` tag inside IE9 & 10 and resizing it using CSS `width` & `height`, it is clipped instead of correctly resized. The [removeViewBox plugin](https://github.com/svg/svgo/blob/master/plugins/removeViewBox.js) causes this issue.

After optimization (broken):

```
<svg xmlns="http://www.w3.org/2000/svg" width="264" height="56">
...
```

Working in IE9/10 after restoring the `viewBox`:

```
<svg xmlns="http://www.w3.org/2000/svg" width="264" height="56" viewBox="0 0 264 56">
...
```
#19 and #124 seem related.

PR #839, opened 2017-11-16
Add addViewBox plugin

maccuaa:
Added a new plugin that will add the viewbox attribute. 

The default configuration for svgo is to remove the viewbox attribute.  Unfortunately, this can lead to unexpected behavior when viewing the SVG in IE because there's no viewbox attribute (see #139)

The plugin checks if there is a height and width , and no viewbox defined.  If that condition is met then it grabs the height and width and sets the viewbox to `0 0 [width] [height]`.

Also added tests and updated the README.

Usage:
```bash
$ svgo --enable=addViewBox --disable=removeViewBox *.svg
```


danaonel:
I hope this will be merged soon. I need it for my current project.

maccuaa:
@danaonel don't get your hopes up.  I ended up installing SVGO globally and then manually changing the default enabled plugins to include this one.  

Not sure why this PR isn't being merged in, it's not turned on by default.

GreLI:
This is image editing, not an optimization plugin. Moreover it's targeted to outdated browsers. Just disable `--disable=removeViewBox`, if you need so, but this task is out of SVGO's scope.

maccuaa:
I think you misunderstand the point of this plugin but that's ok, if you don't want to include it that's your decision.

PR #849, opened 2017-11-22
Fix for issue 848

strarsis:


GreLI:
Did you mean «test»? Or it's an intent to fix?

strarsis:
@GreLI: This should become the basis for a fix. I added a test (which fails currently). I have to add a small parser for [`<begin-value-list>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/begin) format.

PR #857, opened 2017-11-26
dont transform stroke width for non-scaling-strokes

### LINKED ISSUE #856: svgo doesn't properly handle "vector-effect: non-scaling-stroke"
Consider the simple SVG below:

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="500" height="500">
    <style> 
    	path {
    	    fill: #7ED321;
    	    stroke: #000;
    	    stroke-width: 15;
    	    vector-effect: non-scaling-stroke;
    	}
    </style>
    <g transform="translate(250, 250) scale(1.5, 1.5) translate(-250, -250)">
        <path d="M125 125h250v250h-250v-250z"/>
    </g>
</svg>
```

Actual visual result after optimization:

screen shot 2017-11-25 at 2 23 51 pm

Expected visual result after optimization:

screen shot 2017-11-25 at 2 27 07 pm

There is more information on non-scaling strokes in the SVG spec: https://www.w3.org/TR/SVGTiny12/painting.html#NonScalingStroke

alexjlockwood:
Fixes #856.

Ideally we would also remove the `vector-effect` attribute from the `<path>` after applying the transformation... but I don't think we can do this because there could be a parent transform that is also transforming the path. 

If you have any ideas on how we could remove the `vector-effect` attribute (maybe in another plugin?), let me know.

### DIFF
--- a/plugins/_path.js
+++ b/plugins/_path.js
@@ -208,16 +208,22 @@ exports.applyTransforms = function(elem, path, params) {
         if (scale !== 1) {
             var strokeWidth = elem.computedAttr('stroke-width') || defaultStrokeWidth;
 
-            if (elem.hasAttr('stroke-width')) {
-                elem.attrs['stroke-width'].value = elem.attrs['stroke-width'].value.trim()
-                    .replace(regNumericValues, function(num) { return removeLeadingZero(num * scale) });
-            } else {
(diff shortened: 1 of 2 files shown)


GreLI:

PR #860, opened 2017-11-27
Removal of a particular value of an attribute

prashjain309:
Example class="one two"
Calling removeAttr(one) will result in class="two"


prashjain309:
Hi, awaiting your reply.

GreLI:
Yeah, being rather busy lately. There is a caveat: it applies to any attribute, not just class, which can be confusing if working with spaces.

prashjain309:
So it is generic for all attribute and can be used for any attribute I believe. Should it be restricted to some attributes only?

GreLI:
`class` attributes are `CSSClassList` instances, so it should be handled differently.

PR #896, opened 2018-01-24
Add a de-duplicate paths plugin

sfoster:
* Configurable thresholds for minimum path lengths to attempt to optimize
* Configurable number of duplicates before a path is shared
* Duplicate paths matching these criteria are moved to a <defs> element, given an id and referenced using <use href="#generated-id"/>.
* To avoid breaking existing references and keep this plugin's logic simple, any path element with an existing id attribute is skipped


sfoster:
I don't know if there's a policy or guidelines on accepting plugin pull requests - clearly what works for one may not be useful for all. This was giving me 50+% optimization on a set of SVG  files I needed to optimize, where each "frame" in a film-strip arrangement was a transformed shape. 

I'm happy to add detail, docs, tests, refactor etc.

strarsis:
@sfoster: You may find the new API additions useful: https://github.com/svg/svgo/blob/master/docs/how-it-works/en.md#queryselectorallselectors

sfoster:
I'll re-work this to make use of querySelector/querySelectorAll and fix the ci issue - which I assume is to do with the 'active' property.

PR #906, opened 2018-02-16
Allow to forceAbsolutePath output for convertPathData plugin.

qool:
Added parameter forceAbsolutePath to get absolute SVG path (even when the relative path is shorter).

This is useful for generating mobile assets, android vector drawables etc.


GreLI:

PR #907, opened 2018-02-18
Prefix ids plugin don't add prefix if prefix is false

vzaidman:
i use it like this:
```
plugins: [
          {convertPathData: false},
          {prefixIds: {
            prefix: node => {
              const elementName = node && node.elem
              if(!elementName || elementName === 'svg' || elementName === 'use'){
                return false
              }

              return 'some-prefix'
            }
          }}
        ]
```

and i expect no prefix when i return prefix as false.

it can also be used as

```
plugins: [
          {convertPathData: false},
          {prefixIds: {prefix: false}}
        ]
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ bin/svgo-profiling
 *.sublime-*
 *.log
 .DS_Store
+.idea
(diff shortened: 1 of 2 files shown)


strarsis:
@vzaidman: This is very similar to this issue (please also check out the code sample in that issue):
https://github.com/svg/svgo/issues/913#issuecomment-369423275

vzaidman:
reabsed my fix on develop

vzaidman:
Why don't i see it in thew latest version?
can you please publish it?

GreLI:

PR #908, opened 2018-02-19
Fix issue with prefixIDs plugin not replacing url() refs correctly

harrisjose:
When using the prefixIDs plugin  on an SVG with multiple linear gradients, url() values are not replaced correctly because we use `exec` to capture values using the regex. Exec preserves the last matched index, which is undesirable in this case.

Please ref this [ stackoverflow answer ](https://stackoverflow.com/questions/4724701/regexp-exec-returns-null-sporadically) for more information about this.

### DIFF
--- a/plugins/prefixIds.js
+++ b/plugins/prefixIds.js
@@ -13,7 +13,7 @@ exports.description = 'prefix IDs';
 
 var path = require('path'),
     csstree = require('css-tree'),
-    cssRx = require('css-url-regex')(),
+    cssRx = require('css-url-regex'),
     unquote = require('unquote'),
     collections = require('./_collections.js'),
     referencesProps = collections.referencesProps,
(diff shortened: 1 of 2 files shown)


harrisjose:
@strarsis @caub 

Can you guys take a look at this and let me know if this is okay?

strarsis:
@harrisjose: Good catch. 
There is a still open issue about this: https://github.com/cssstats/css-url-regex/issues/4

harrisjose:
@strarsis I thought I'd looked in the css-url-regex module for this. Oh well.

If there aren't any further changes required, can we get this merged and released? I'd really appreciate it cause we're running into this issue with a few svgs at work. Thanks.

harrisjose:
@GreLI 
Sorry if I'm being a nuisance but can we get this released as a patch? Would help a lot. Thanks.

caub:
@harrisjose done

PR #916, opened 2018-03-02
Fixes #915 for collapseGroups

### LINKED ISSUE #915: collapseGroups and "inherit" value
For this file from the SVG Test suite: https://www.w3.org/Graphics/SVG/Test/20061213/svggen/color-prop-01-b.svg

svgo (specifically the `collapseGroups` plugin) removes the outermost `<g>` from this markup:

```svg
<g color="red">
	<g color="inherit" fill="none" stroke="none">
		<!-- inherit to check the property value, not the string, is being used -->
		<circle cx="130" cy="80" r="60" fill="currentColor"/>
		<circle cx="350" cy="80" r="60" stroke="currentColor" stroke-width="4"/>
	</g>
</g>
```

When the attribute is inheritable, and is present on the inner element, but has the value `inherit`, the attribute on the outer group should be maintained.

danburzo:
When the inner element has the `inherit` value, overwrite with the value from the parent `<g>` element.

### DIFF
--- a/plugins/collapseGroups.js
+++ b/plugins/collapseGroups.js
@@ -64,8 +64,13 @@ exports.fn = function(item) {
                             } else if (attr.name == 'transform') {
                                 inner.attr(attr.name).value = attr.value + ' ' + inner.attr(attr.name).value;
                             } else if (
-                                attrsInheritable.indexOf(attr.name) < 0 &&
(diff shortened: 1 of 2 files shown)


GreLI:
Good catch!

GreLI:

PR #924, opened 2018-03-19
Upgrade css-tree from alpha25 to alpha.28

shysteph:
API changed walkFoo(ast, fun) to walk(ast,{visit: 'Rule',enter: fun)
csstree.translate renamed to csstree.generate
csstree.generate throws an error on invalid style where translate silently ignored it - update style processing to catch and ignore instead
Updated tests - csstree now omits more whitespace

Should convertStyleToAttrs.0.2.svg actually pass like it currently does - dropping some of the most invalid properties or should I update the code to treat invalid styles as parse errors/some other behavior?


mhart:
Now up to alpha.29 – would love to see this go in so that there's deduplication with `csso`, which `svgo` also depends on

PR #928, opened 2018-03-23
Added visibility attribute handling for removeHiddenElems

mikolaj92:
The name of the plugin is misleading as previously it wasn't removing hidden attributes.


GreLI:

PR #929, opened 2018-03-24
Add removeAttributesBySelector plugin

bmease:
This pull request adds the `removeAttributesBySelector` plugin. Allowing the removal of attributes in elements found by a css selector.

For a project I'm currently working on, I need to remove the attribute `fill` only if it has the value `#00ff00`. The `removeAttrs` plugin only queries the element name and is not able to meet this need.

I'm happy to add other plugins like `addAttributesBySelector`, or refactor `removeAttrs`, or create another plugin that combines all this functionality. For now, I've just left it as a separate plugin.

```
plugins:
  - removeAttributesBySelector:
    selector: "[fill='#00ff00']"
    attributes: "fill"

<rect x="0" y="0" width="100" height="100" fill="#00ff00" />
  ↓
<rect x="0" y="0" width="100" height="100"/>
```

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -55,6 +55,7 @@ plugins:
   - removeDesc
   - removeDimensions
   - removeAttrs
+  - removeAttributesBySelector
   - removeElementsByAttr
   - addClassesToSVGElement
   - removeStyleElement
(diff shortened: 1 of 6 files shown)


Herman-Freund:
#977 Addresses the same issue utilizing the existing plugin.

However, The plugin proposed in this PR has an extra benefit of being able to remove a different attribute than the one matching the selector.

GreLI:

PR #977, opened 2018-06-06
removeAttrs: Added optional value filter

Herman-Freund:
Example:

Remove all "fill" attributes that the value is "none"
```
{
  removeAttrs: {
    attrs: '*:fill:none'
  }
}
```

Remove all "fill" attributes where the value is NOT "none"
```
{
  removeAttrs: {
    attrs: '*:fill:((?!^none$).)*'
  }
}
```

Remove all "fill" attributes except when the value is white (white, #fff, rgb(255,255,255), etc.)
```
attrs: [
  "*:fill:((?!^\\s*white\\s*$|^\\s*#(f{3}|f{4}|f{6}|f{8})\\s*$|^\\s*rgba?\\(\\s*(255|100%)\\s*([,\\s]\\s*\\3\\s*){2}([,\\s]\\s*(1|100%)\\s*)?\\)\\s*$|^\\s*hsla?\\(\\s*[\\w.]+\\s*[,\\s]\\s*\\d{1,3}%\\s*[,\\s]\\s*100%\\s*([/,\\s]\\s*(1|100%)\\s*)?).)*",
  ...
]
```

### DIFF
--- a/plugins/removeAttrs.js
+++ b/plugins/removeAttrs.js
@@ -21,10 +21,11 @@ exports.params = {
  *
  * @param attrs:
  *
- *   format: [ element* : attribute* ]
+ *   format: [ element* : attribute* : value* ]
  *
- *   element   : regexp (wrapped into ^...$), single * or omitted > all elements
+ *   element   : regexp (wrapped into ^...$), single * or omitted > all elements (must be present when value is used)
  *   attribute : regexp (wrapped into ^...$)
+ *   value     : regexp (wrapped into ^...$), single * or omitted > all values
  *
  *   examples:
  *
@@ -36,6 +37,10 @@ exports.params = {
  *     > remove fill attribute on path element
  *     ---
  *       attrs: 'path:fill'
+ *
+ *     > remove fill attribute on path element where value is none
+ *     ---
+ *       attrs: 'path:fill:none'
  *
  *
  *     > remove all fill and stroke attribute
@@ -55,6 +60,10 @@ exports.params = {
  *     [is same as]
  *
  *       attrs: '.*:(fill|stroke)'
+ *
+ *     [is same as]
+ *
+ *       attrs: '.*:(fill|stroke):.*'
  *
  *
  *     > remove all stroke related attributes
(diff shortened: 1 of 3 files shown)


Herman-Freund:
@GreLI How can we move this forward? Will pinging the original creator of this plugin help (@bennyschudel)?

Thanks

frebro:
This would be a very useful plugin.

plugins/removeAttrs.js
```diff
@@ -81,12 +90,16 @@ exports.fn = function(item, params) {
             // prepare patterns
         var patterns = params.attrs.map(function(pattern) {
 
-                // apply to all elements if specifc element is omitted
+                // if no element separators (:), assume it's attribute name, and apply to all elements and regardless of value
```
roblevintennis: maybe should say "to all element regardless of value" (I just removed the extraneous 'and')

roblevintennis:
I too think this would be a nice addition to the plugin...Herman chimed in on my PR https://github.com/svg/svgo/issues/999 which is a parity feature with grunt-svgstore and allows for using the SVG trick to style sprites with one more color by using the `currentColor` value to inherit font's `color`. That said, I could alternatively use his changes here to achieve with regex filtering `{"attrs":"*:(stroke|fill):((?!^currentColor$).)*"}` so that's a definite use case for getting this in.

Here's an article I wrote on CSS tricks which explains the use case aforementioned:
https://css-tricks.com/gotchas-on-getting-svg-into-production/#article-header-id-11

plugins/removeAttrs.js
Herman-Freund: Maybe but I was trying to emphasize that the plugin will make 2 assumptions if there are no separators, 1) match all elements and 2) match any value.

roblevintennis: Maybe:

>assume it's attribute name, and apply to all elements (regardless of value)

Herman-Freund: Doesn't parentheses imply "lesser importance" (e.g. afterthought)? Which is not exactly the message intended to convey.

Thanks

roblevintennis: I suppose you could state something like:

> if no element separators (:), apply the pattern matching to attribute name while otherwise matching on all elements and values

Seems to express the regex your join will result in.

Herman-Freund: Sorry but I don't understand how we got so affixed on some Grammer in a comment within the code. I think the intended message comes across pretty well in the original version (which for some reason - for me at least - wasn't so in any of the suggestions). Can we get past this please?

Thanks

plugins/removeAttrs.js
```diff
@@ -81,12 +90,16 @@ exports.fn = function(item, params) {
             // prepare patterns
         var patterns = params.attrs.map(function(pattern) {
 
-                // apply to all elements if specifc element is omitted
+                // if no element separators (:), assume it's attribute name, and apply to all elements *regardless of value*
```
roblevintennis: *test*
foo

plugins/removeAttrs.js
roblevintennis: 👍

raduflp:
any plans to merge this PR in the near future?

MattMakes:
@deepsweet @GreLI Any way to get this moved through?

Jetski5822:
Any way this can be merged in?

Herman-Freund:
> @Herman-Freund did you ever hear any feedback outside of this channel on your work?

No 😕

Thanks everyone for the positive feedback!

Herman-Freund:
@GreLI 
Other than #1055, are there any reasons for not merging this?

Thanks

GreLI:
Merged in d23355e91a20d39bfe13bfbadf81155cdc18c541 (there was a conflict with test from other PR).

GreLI:

PR #978, opened 2018-06-09
tweak: do not collapse group properties to children if they contain filters.

stristr:
Some CSS filters have layout properties, a canonical example being:

```
<svg>
  <defs>
    <filter id="a">
      <feOffset dx="10" />
      <feGaussianBlur …/>
      <feColorMatrix …/>
    </filter>
  </defs>
  <g filter="url(#a)">
    <circle … />
  </g>
</svg>
```

(This happens to be how Figma does drop shadows, which is how I noticed this.)

When `svgo` pushes the `filter` attribute down to the `<circle />`, the gaussian blur gets awkwardly clipped on the top-left edges of the circle's bounding box. Based on this, I concluded it's best to not push down filters, just like `clip-path` and so forth.

### DIFF
--- a/plugins/collapseGroups.js
+++ b/plugins/collapseGroups.js
@@ -50,7 +50,7 @@ exports.fn = function(item) {
                 if (g.hasAttr() && g.content.length === 1) {
                     var inner = g.content[0];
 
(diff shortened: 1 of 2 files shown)


GreLI:

PR #979, opened 2018-06-09
Add removeOffCanvasPaths plugin

JoshyPHP:
The plugin is rudimentary: it compares each `<path>` to the viewBox using `intersects()` and removes those that don't. The plugin is disabled by default.

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -60,6 +60,7 @@ plugins:
   - removeStyleElement
   - removeScriptElement
   - addAttributesToSVGElement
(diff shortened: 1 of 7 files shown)


JoshyPHP:
I have tested this plugin on various SVGs and while it works well enough for most, there are many edge cases that are difficult to handle. For instance, items whose geometry is modified by a transformation matrix.

GreLI:

hugolpz:
Related: #1116.

Issue #971, opened 2018-06-01
Plugin to optimize embedded images ?


Zulko:
Hi and thanks for a super-useful project. I have SVGs embedding PNG images (they are represented inline as base64 in the XML) and I want to convert them to JPG to reduce the size of my SVGs. It doesn't seem that thereis a plugin for this. Would you be interested by one ? has this been tried already ?

paperboyo:
Just a user here. Images without transparency should be converted to JPGs, but the ones with transparency have to be converted by [pngquant](https://pngquant.org/) to PNG8. This may not be trivial as you would need to be able to tell one from the other and maybe provide options to control both compression levels.

Would be very useful, though!

GreLI:
Yeah, it would be nice, but it's a complex topic. There a lot of optimizers with different optimizations degree and execution time, also they're a hardly depend on image type (low-color graphics compress differently from photo-like, dealing with semitransparency has its own tricks, etc).

Also, it would add extra dependencies, optimizers are binary, so you have to deal with different platforms, bloating dependencies even more, which I'd prefer to avoid, since far not everyone needs it. It more like a constructor for image optimization then a single format optimization like SVGO, so I believe it's out of SVGO's scope.

Zulko:
[jimp](https://www.npmjs.com/package/jimp) is a pure-javascript library that can do JPG. My intention is to just try some simple PNG -> JPG optimizer for my own SVG needs. Is it possible to make "external" plugins for SVGO or do plugins have to be hosted in this repository ?

paperboyo:
> some simple PNG -> JPG

Related to know where to NOT convert (if there is transparency): https://github.com/oliver-moran/jimp/issues/271. By having a very cursory look at jimp, I can’t tell if it supports pngquant-like PNG8 with 8-bit transparency (PNG32 is too weighty, PNG8 with 1-bit transparency is useless).

julienboulay:
Hi, I created a new plugin to optimise png and jpeg embedded in the svg as base64.
It works well for me, decreasing the size of the svg by 50%.
Let me know if it fits your needs

Zulko:
I went with [my own solution](https://github.com/RelaxedJS/relaxed-svg/blob/5c181e69ff47a6e97696a987e806fb9b40cb83c3/src/optimizeSVG.js#L11-L25) (upstream of SVGO) and it works well for me too, but I'd still be interested in seing your solution (and even better if it comes included in SVGO), as a SVGO plugin would be cleaner.

[...]

PR #982, opened 2018-06-13
Add minifyImages plugin to optimize embedded png and jpeg

julienboulay:
This new plugin allows to optimize png and jpeg with pngquant and jpegtran.
It fixes an existing issue : https://github.com/svg/svgo/issues/971


canercandan:
@julienboulay what is the current state of this PR?

julienboulay:
Hi @canercandan,
This PR has been rejected because it includes external non-js dependencies (pngquant and jpegtran)
See comments on issue #971

GreLI:
For the report, execa requires Node >= 6, while SVGO currently runs on 4 and higher. Should bump it, but not yet :man_shrugging:. Also, need to do something with tests and CI. Maybe run them only locally.

TrySound:
This PR adds a lot of packages. Though we would like to keep svgo as small as possible. You can publish this plugin as separate package. Configuration allows custom plugins.

PR #984, opened 2018-06-14
[plugin] sortAttrs: add xmlns order option

hellatan:
i wanted this option so that `xmlns` attributes weren't always at the front of the node.


hellatan:
thanks for merging @GreLI. do you know when the next version will be published?

GreLI:

PR #998, opened 2018-07-11
Return the path

gagyibenedek:
This is useful, when you want to optimize a bunch of files at once.  Thanks to this small change, you can just Promise.all the optimizations, then use the results to overwrite the source files with the optimized content.


GreLI:

GreLI:
Your change caused bug #1030, please be careful next time.

PR #1005, opened 2018-07-23
SVGO: Add plugin "reusePaths".

jhowcrof:
Find duplicated <path> elements and replace them with <use> links. (disabled by default)


tigt:
This maybe should check for other styling attributes than `stroke` and `fill` — it could change rendering if one of the paths has `stroke-dasharray`, for example. It probably shouldn't clone paths unless the differences fall under the attributes where it's safe for the `<use>` to differ:

* `id` and `class`
* `transform`
* Maybe some others

It also doesn't have to move the "source" `<path>` into a `<defs>` block — referencing an already-visible element with `<use> ` is perfectly okay.

GreLI:
@tigt worth sending PR?

PR #1009, opened 2018-07-26
Add doc entry for inlineStyles plugin

poeschko:


GreLI:
Well, `inlineStyles` converts `<style>` element content to `style` attr for elements, where `convertStyleToAttrs` converts appropriate `style` attrs to corresponding attributes, like `style="fill:red"` to `fill="red"`. Attributes of children elements can be moved to parent `<g>` element.

poeschko:
That makes sense, I missed that. Thanks for the explanation!

PR #1028, opened 2018-09-14
Don’t round `version` attribute

### LINKED ISSUE #1003: Rounding numbers to integers incorrectly affects `version` attribute
For example, `<svg version="1.1"/>` becomes `<svg version="1"/>`.

This means the `version` attribute isn't removed when "Remove default attributes" is run later.

tigt:
Fixes #1003


GreLI:

PR #1056, opened 2018-10-18
Make default config behave as immutable

### LINKED ISSUE #1052: Multiline svg with option `pretty: true` as output, when processing many files
When I'm processing many files with option `pretty` set to `true`, in output files there are multilines showing up. In each file, one by one, multilines are incremented by one, though in my 300th file I've got 300 empty lines between 'content' lines 😮 

It's because a file `js2svg.js` is called once, and `this.config` is a reference to the object in the higher scope. Though any changes are persistent.

I fixed in in my PR

panczarny:
Closes #1052


GreLI:

PR #1058, opened 2018-11-01
New plugin embedLinkedImages: embed images as base64 data URIs

shoogle:
Clearly embedding images as base64 will make the SVG bigger, not smaller, but the SVG is still optimized in the sense that it becomes a fully self-contained resource (no external dependencies).

Embedding images has the following __benefits__ compared to linking:

- More portable (move the SVG without risk of losing images)
- Fewer requests made to server / filesystem (potential speed increase)
- Increased compatibility with SVG viewers (loading of external resources not always implemented)
- Makes the file useable inside HTML's `` tag (loading of external resources [disallowed for security reasons](https://bugzilla.mozilla.org/show_bug.cgi?id=628747))

Embedding has the following __drawbacks__ compared to linking:

- Increased file size of SVG (since images are embedded rather than separate)
- Increased overall file size (base64 is inefficient compared to raw binary data)
- Embedded resources don't benefit from browser caching to the same extent as linked resources

Hence this is a useful feature, but one that should probably be optional and disabled by default (though it is enabled by default in other SVG optimizers, such as [Scour](https://github.com/scour-project/scour/blob/049264eba6b1a54ae5ba1d6a5077d8e7b80e8835/scour/scour.py#L3090)).


shoogle:
Embedding is not something you can rely on an editor to do, as you might want images to be linked in the source SVG and only embedded in the published version. This keeps the source SVG small, making it easier to keep track of in a Git repository, and edit by hand if necessary.

kimmobrunfeldt:
Not sure if svgo is the place for this task but it would certainly help our use case as well. We already optimize SVGs with svgo, and would need a way to embed external images as inline base64 versions.

PR #1060, opened 2018-11-22
Add config to preserve IDs based on prefix matching

bkotzz:
There is currently an option to opt out of the ID cleanup by specifying IDs that we want left alone. This would also be useful to do on prefixes. That way, we could say, preserve `rect_*`, without having to enumerate all the possibilities of `*`. And, having both prefix and full string matching is useful, so you ca, for example, preserve `rect` specifically but not `rect_blue`.

This change adds another config array that takes in a list of prefix strings, and preserves any IDs that start with one of those prefixes. Also adds tests for this behaviour standalone, and in conjunction with `force` and the full-string ID preserving config.

PTAL @GreLI


bkotzz:
@GreLI can you take a look when you get a chance please?

bkotzz:
Thanks @GreLI. Any plans on an npm release soon?

GreLI:
Yep, I've released it just now.

PR #1061, opened 2018-11-27
Replace colors dependency with chalk

### LINKED ISSUE #410: colors
Replace please colors with [chalk](https://github.com/chalk/chalk). It more safe, because do not modify prototypes.

xPaw:
Fixes #410. `colors` pollutes `String.prototype` which isn't a very good idea especially when other plugins require `svgo`. See https://www.npmjs.com/package/chalk#origin-story

I couldn't find any other places where colors are used (this is part of the part, as it makes it very hard to find it).

Quickly looking at git commits I saw that error messages were colored red when colors were added, but were removed in 2816424ed7148aa664ee8046c02b8320083bea28, I added this back, but maybe that's unwanted?


xPaw:
`coa` was switched to use `chalk` as well: https://github.com/veged/coa/pull/96

That means if this PR is merged, colors can be consistently disabled in `svgo` CLI by passing `--no-color` or setting `FORCE_COLOR=0` env.

xPaw:
@GreLI Mind taking a look please?

GreLI:

PR #1083, opened 2019-03-01
 Update removeElementsByAttr plugin

notiv-nt:
https://github.com/svg/svgo/pull/1083/commits/110f4244afd3fd4fe4e8ebd57cf7c0b500d867cd
for regex here:
https://github.com/svg/svgo/pull/1083/commits/84c818f774ce381bb03e745040cca8195c5cb3a4#diff-99062b3052822642f24f19f3b56bd7d1R11

### DIFF
--- a/test/plugins/_index.js
+++ b/test/plugins/_index.js
@@ -38,9 +38,11 @@ describe('plugins tests', function() {
 
                     var getParams = function() {
                         var t = {};
+                        /* jshint ignore:start */
                         eval('t = ' + params);
+                        /* jshint ignore:end */
                         return t;
-                    }
+                    };
 
                     plugins[name] = (params) ? getParams() : true;
 
--- a/test/plugins/_index.js
+++ b/test/plugins/_index.js
@@ -36,7 +36,13 @@ describe('plugins tests', function() {
                         plugins = {},
                         svgo;
 
-                    plugins[name] = (params) ? JSON.parse(params) : true;
+                    var getParams = function() {
+                        var t = {};
+                        eval('t = ' + params);
+                        return t;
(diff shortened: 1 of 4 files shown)


test/plugins/_index.js
```diff
@@ -36,7 +36,15 @@ describe('plugins tests', function() {
                         plugins = {},
                         svgo;
 
-                    plugins[name] = (params) ? JSON.parse(params) : true;
+                    var getParams = function() {
+                        var t = {};
+                        /* jshint ignore:start */
+                        eval('t = ' + params);
```
GreLI: Why `eval`? For regular expressions you can use `RegExp` constructor. `eval` is way too fragile and not-secure.

notiv-nt: Json.parse know about RegExp?

GreLI: FYI, JSON.parse [has optional `reviver` parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) that allow custom parsing.

notiv-nt: > parsing is transformed before being returned
Transformed, not parsed

notiv-nt:
```
JSON.parse('{ "d": /^M0\s0h\d0v\d0H0z$/i }')
JSON.parse('{ "d": /^M0\s0h\d0v\d0H0z$/i }', (k, v) => { /* ... */ })
```

are equal.

Do you really think, that parsing string that looks like regexp will be better instead of 3 lines of code?

TrySound:
This would be a breaking change.

PR #1085, opened 2019-03-09
Make removeDimensions add viewBox if it's missing

adipascu:
This change allows removeDimensions on SVG's that lack the viewBox attribute but has valid width and height attributes.

I have such SVG's in my project and I prefer to not edit them by hand.
This change is backwards compatible, I did not change any of the previous unit tests, but added two additional ones for the new functionality.


adipascu:
@GreLI thanks for the merge!
🤘

GreLI:
SVGO v1.3.0 with the fix is released.

PR #1091, opened 2019-03-27
Add ellipseToCircle plugin

tigt:
This converts non-eccentric ellipse elements to circle elements.
Adobe Illustrator seems to be fond of using ellipses even if their rx === ry.


GreLI:
SVGO v1.3.0 with the plugin is released.

tigt:
Awesome! Thank you for adding it to the README and default configuration — I should have done that.

PR #1098, opened 2019-04-15
compressRasterImages plugin support

CT1994:
this plugin adds support for lossless compression of base64 images within a svg.


TrySound:
Thanks. One of added packages is [very big](https://packagephobia.com/result?p=ect-bin). We would like to keep svgo as small as possible. Though you can publish your plugin as separate package.

Issue #1062, opened 2018-11-29
What options can change to hid prefix class of svg file?


TongDaDa:
<img width="387" alt="2018-11-29 11 50 13" src="https://user-images.githubusercontent.com/23241188/49198240-24708800-f3cd-11e8-9a2c-8a12b60ee150.png">

The className of text tag has named `virtualBox_TextInSvg`, but it carry a prefix that is filename after outputed, What should I do?

I want to keep the same classname in svgfile

![2018-11-29 11 49 45](https://user-images.githubusercontent.com/23241188/49198242-24708800-f3cd-11e8-87f6-6e1b1e53d998.png)

strarsis:
`svgo --disable=prefixIds` should disable the prefixIds plugin that prefixes IDs + classes.

valkko:
Is there a way to _only_ prefix IDs, but not classes? I have a case where the classes should stay as is, but IDs prefixed to avoid conflicts.

Since the config option is named `prefixIds` you'd think it would only affect IDs in the first place.

litinskii:
How to use prefix only for class ?

strarsis:
@litinskii: Hm, this would probably require code changes in the `prefixIds` plugin.
Only one method would have to be called then, `addPrefixToClassAttr`(https://github.com/svg/svgo/blob/master/plugins/prefixIds.js#L212).

litinskii:
@strarsis Thanks for answer. I'm asking this, because I'm trying to migrate from https://github.com/jkphl/svg-sprite, and there was an option like 
![image](https://user-images.githubusercontent.com/10028441/57966846-38e71380-7960-11e9-9aae-c2d2e3e4261a.png)
(https://github.com/jkphl/svg-sprite/blob/master/docs/configuration.md)

and it was very useful, I mean allow to specify if you want to add namespace only for class for example. 

I thinks it would be cool if we have some king of options for prefixIds plugin. @strarsis What you think about this ?

strarsis:
@litinskii: You can try the PR here: https://github.com/svg/svgo/pull/1106

litinskii:
@strarsis Thanks will try.

litinskii:
@strarsis seems it work for my case. Thanks.

PR #1106, opened 2019-05-18
Selectively prefixing IDs and/or class names by prefixIds plugin

strarsis:
This PR adds options for selectively prefixing IDs and/or class names by prefixIds plugin.

See https://github.com/svg/svgo/issues/1062


strarsis:
@GreLI: Any chances this can be merged in? It is only a small modification but is very useful.

PR #1122, opened 2019-06-19
support browser bundlers

cancerberoSgx:
with these changes browserify, parcel, etc will be able to bundle svgo for browsers.

I know there are still things missing but, before I continue, I want to ask you if this can of approach would be acceptable for merging?

Things missing: 

 * documentation (example of using browserify to build and run svgo in the browser)
 * Maybe there is some concern about require() all plugins at startup in node.js. If that's the case it can be fixed.

Less important: 

 * `pluginIndex.js` can be generated automatically at build time using 
[svgo-browser.js.zip](https://github.com/svg/svgo/files/3305542/svgo-browser.js.zip)
 * commands `make browserify` or `npm run browserify` to generate a bundle that can be distribute it
   * "browser" field in package.json

Thanks! Will wait for your reply, in the meantime at least can be helpful for those looking into the same problem.


GreLI:
That's not the way I'm going to.

dr-js:
Maybe try the portable webpack output https://unpkg.com/libsvgo/webpack/SVGO.js from [dr-js/libsvgo](https://github.com/dr-js/libsvgo), a re-formatted ES6+ fork of svgo.

PR #1127, opened 2019-07-11
Replace csso with cssnano

strarsis:
Next release of `cssnano` will support synchronous usage, see https://github.com/cssnano/cssnano/issues/68#issuecomment-510563780.


GreLI:
Why reduce functionality?

Issue #1133, opened 2019-07-24
Repeated values in class names


ilariaventurini:
Use `--multipass` option and `{"prefixIds":true}` creates css class names with repeated values.
For example run `svgo hello3.svg -o - --pretty --config='{"plugins":[{"prefixIds":true}]}' --multipass` using the svg file `hello3.svg`:

    <svg xmlns="http://www.w3.org/2000/svg" id="Livello_1" data-name="Livello 1" viewBox="0 0 300 170">
      <style>
        .cls-1{fill:#d9e7f2;}.cls-2{fill:#1191ae;}
      </style>
      <rect width="265.1" height="1" x="32.9" y="117.95" class="cls-1"/>
      <rect width="265.1" height="1" x="32.9" y="90.08" class="cls-1"/>
      <rect width="265.1" height="1" x="32.9" y="61.68" class="cls-1"/>
      <rect width="265.1" height="1" x="32.9" y="33.8" class="cls-1"/>
      <rect width="266.03" height="1" x="32.4" y="5.4" class="cls-1"/>
      <rect width="15.12" height="27.88" x="54.66" y="118.5" class="cls-2"/>
      <rect width="15.12" height="39.97" x="89.49" y="106.4" class="cls-2"/>
      <rect width="15.12" height="80.47" x="124.33" y="65.91" class="cls-2"/>
      <rect width="15.12" height="98.35" x="159.16" y="48.02" class="cls-2"/>
      <rect width="15.12" height="83.1" x="193.99" y="63.28" class="cls-2"/>
      <rect width="15.12" height="108.87" x="228.83" y="37.51" class="cls-2"/>
      <rect width="15.12" height="30.5" x="263.66" y="115.87" class="cls-2"/>
      <rect width="266.03" height="1" x="32.4" y="145.83" class="cls-1"/>
    </svg>

Creates this svg:

    <svg xmlns="http://www.w3.org/2000/svg" id="hello3_svg__hello3_svg__hello3_svg__Livello_1" data-name="Livello 1" viewBox="0 0 300 170">
        <style>
            .hello3_svg__hello3_svg__hello3_svg__cls-1{fill:#d9e7f2}
        </style>
        <path class="hello3_svg__hello3_svg__hello3_svg__cls-1" d="M32.9 117.95H298v1H32.9zm0-27.87H298v1H32.9zm0-28.4H298v1H32.9zm0-27.88H298v1H32.9zm-.5-28.4h266.03v1H32.4z"/>
[...]

PR #1135, opened 2019-07-24
Closes: #1133

ilariaventurini:
Repeated values in class names.
For more information, see issue #1133 .

### DIFF
--- a/plugins/prefixIds.js
+++ b/plugins/prefixIds.js
@@ -146,6 +146,10 @@ exports.fn = function(node, opts, extra) {
         if(prefix === false){
             return escapeIdentifierName(name);
         }
+        if (name.startsWith(escapeIdentifierName(prefix) + opts.delim)) {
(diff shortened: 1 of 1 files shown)


plugins/prefixIds.js
```diff
@@ -146,6 +146,10 @@ exports.fn = function(node, opts, extra) {
         if(prefix === false){
             return escapeIdentifierName(name);
         }
+        if (name.startsWith(escapeIdentifierName(prefix) + opts.delim)) {
```
XhmikosR: You should combine this check with the  previous one since they return the same value.

ilariaventurini: @GreLI @XhmikosR Thanks! Solved

XhmikosR:
@ilariaventurini You have unrelated changes though.

TrySound:
Prefixing with multipass option is fixed.

PR #1144, opened 2019-08-22
svgicons2svgfont support 

unbywyd:
After update from 1.2.2 to 1.3 plugin svgicons2svgfont stopped working after svg optimization, i tested with base plugins of svgo (without options)

Error`s type: Expected a flag, got "0024" at index "102"...


GreLI:
There is mess, I don't see a fix.

PR #1147, opened 2019-08-29
Fix cleanupids defs last elem

alexpavlovich:
Hi! Today I encountered with unexpected behavior of `cleanupIDs` plugin.

I had an SVG with `<defs>` declared at the end of file, yet still containing another element in front of `<defs>`. But plugin treated this SVG as `hasDefsOnly`.


alexpavlovich:
It checks if the whole SVG has only `<defs>` in it. If you do not think this is right, I would appreciate a valid test case in which this check fails.

PR #1154, opened 2019-09-19
Adds reuseImage plugin

jack-burridge-tp:
Adds a plugin that swaps out images with use


TrySound:
See https://github.com/svg/svgo/pull/1624
JSAPI and other plugin apis except visitor will be removed in v3

PR #1156, opened 2019-09-22
pull sources via secure protocol

vszakats:
(`git://` is unauthenticated/unencrypted)

Also upgrade rest of URLs to HTTPS.


vszakats:
It's still insecure, but closing due to no interest.

PR #1161, opened 2019-10-07
Added keyword feature [name] for addClassesToSVGElement

timwright12:
Adds the ability to use a keyword slug in classnaming similar to `file-loader`.
`svg-[name]` on home.svg will be processed to use svg-home as a class.


TrySound:
File name may have characters unsupported in classes. Also this adds dependency on node builtins. We are trying to keep core browser compatible. You can always use custom plugin btw.

PR #1162, opened 2019-10-08
change profit color output to green and red

gimler:
Actually the number is always green.

I change the color to red when the filesize is increased.


SethFalco:
Hey! Thanks for reporting this and submitting a fix. I rebased it for you, since there were conflicts.

This will be released in SVGO v3.0.5.

PR #1165, opened 2019-10-12
Selectively remove matched style/attr

strarsis:
This PR adds options to `removeMatchedSelectors` option for selectively removing style and/or attr.
This can be useful if the attrs should be kept for selecting or styling with pseudo-classes (e.g. `:hover`).

### DIFF
--- a/plugins/inlineStyles.js
+++ b/plugins/inlineStyles.js
@@ -6,7 +6,10 @@ exports.active = true;
 
 exports.params = {
     onlyMatchedOnce: true,
-    removeMatchedSelectors: true,
+    removeMatchedSelectors: {
+        style: true,
+        attr:  true,
+    },
     useMqs: ['', 'screen'],
     usePseudos: ['']
 };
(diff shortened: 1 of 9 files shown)


TrySound:
Hi @strarsis. I don't get what "attr" does here. Is this still relevant?

strarsis:
@TrySound: This PR adds optional sub-options to the `removeMatchedSelectors` option which allow finer control:
Should styles be kept after inlining?
Should attributes (`id` or `class` currently) be removed after the matching styles have been applied?
Especially the second option can be useful, e.g. styling parts of a SVG using its classes after inlining the styles.

SethFalco:
This shouldn't be neccesary anymore.

The problem this is solving has been address by:

* https://github.com/svg/svgo/pull/1829
* https://github.com/svg/svgo/pull/1832
* https://github.com/svg/svgo/pull/1836

Feel free to share an SVG if you can still reproduce scenarios where we're removing attributes too eagerly. However, in theory it should be possible for us to even drop `removeMatchedSelectors` entirely now I think. 🤔

Issue #1148, opened 2019-09-04
prefixIds: { true } generates prefix__prefix__cls-1 class names


patrick-wc:
Thanks very much to @strarsis for creating the plugin, I was having to manually change class names before using this which was awful. However when I use this plugin with gulp-imagemin I can't get the filename to be passed into the output svg filename. 

If I use this config (which from reading around this repo seems to be the way that is suggested, there is no documentation).

```
.pipe(
        imagemin([
          imagemin.gifsicle({ interlaced: true }),
          imagemin.jpegtran({ progressive: true }),
          imagemin.optipng({ optimizationLevel: 5 }),
          imagemin.svgo({
            plugins: [
              { removeViewBox: true },
              { cleanupIDs: false },
              { prefixIds: true }
            ]
          })
        ])
      )
```

The output file SVG class names (or IDs) are changed from "cls-1" to "prefix__prefix__cls-1" or in one file I noticed "Layer_2" being changed to "prefix__prefix__prefix__Layer_2"

If I use it with this config, (prefix part copied from [@ming-codes](https://github.com/svg/svgo/issues/674#issuecomment-328774019) )

```
.pipe(
        imagemin([
          imagemin.gifsicle({ interlaced: true }),
          imagemin.jpegtran({ progressive: true }),
          imagemin.optipng({ optimizationLevel: 5 }),
          imagemin.svgo({
            plugins: [
              { removeViewBox: true },
              { cleanupIDs: false },
              {
                prefixIds: {
                  prefix: {
                    toString() {
                      this.counter = this.counter || 0;

                      return `id-${this.counter++}`;
                    }
                  }
                }
              }
            ]
          })
        ])
      )
```

The output described above changes from "cls-1" to "id-125__id-111__cls-1" and "id-391__id-373__Layer_2" which although slightly off means I don't get any issues with the same SVGs sharing the same class names which makes them all screw up when inlining them on a page.

It still seems like something is being passed twice as there are repeated values (`prefix__prefix__` and `id-#__id-#`) which is similar to [this issue](https://github.com/svg/svgo/issues/1133), but I don't think I'm using multipass. I'm not sure why, it's probably my mistake on the configuration arguments but this has taken me a while to get working so I'm hoping this helps someone else. 

I am using the latest versions of gulp (4.0.2) and gulp-imagemin (6.1.0) on Windows 10 with Windows Subsystem Linux (WSL to help people out in future when searching).

[...]

PR #1168, opened 2019-10-21
Fix multiple prefixes in multipass

strarsis:
[`imagemin-svgo` enables `svgo` `multipass` by default](https://github.com/imagemin/imagemin-svgo/blob/master/index.js#L6), hence many users reports this issue (https://github.com/svg/svgo/issues/1148).

This PR will add a flag to the internal `svgo` AST for each node processed by the `prefixIds` plugin to prevent multiple prefixing, which currently happens in `multipass` mode.

(Originally the `prefixIds` plugin should check the beginning of each name and skip if it equals the prefix, but there can be occasions where the prefix is the same as a non-prefixed name, and this can result in collisions between SVGs because of skipped prefixing.)

### DIFF
--- a/plugins/prefixIds.js
+++ b/plugins/prefixIds.js
@@ -125,6 +125,13 @@ var addPrefixToUrlAttr = function(attr) {
  */
 exports.fn = function(node, opts, extra) {
 
+    // prevent multiple prefixing in multipass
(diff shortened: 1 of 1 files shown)


strarsis:
Retracting this PR for a better solution that adds `multipassCount` to info object that is passed to each plugin.

PR #1169, opened 2019-10-22
ReplaceUse plugin

bbqaaq:
Referring to https://github.com/svg/svgo/pull/808, this is an updated version of the replaceUse plugin. Indeed I have run this plugin for a long time.

So far my service has been running this plugin for more than a year of time and perform well indeed. The usage is to remove some <symbol> or <def> tag, and merge into the corresponding <use> tag.

I run it with multi-pass settings on, on thousands of noun project icon, and it really cleanup many redundant icons with lots of <use>.

### DIFF
--- a/.svgo.yml
+++ b/.svgo.yml
@@ -29,7 +29,6 @@ plugins:
   - cleanupIDs
   - prefixIds
   - removeRasterImages
-  - replaceUse
   - removeUselessDefs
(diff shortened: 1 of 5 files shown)


TrySound:
See https://github.com/svg/svgo/pull/1624
Only visitor plugin api will work in v3

PR #1171, opened 2019-10-24
Add bundle feature

larrabee:
The new command `npm run bundle` bundle the project into a single binary file.  
This allows you to run svgo without the node.js interpreter (it is packed into a binary file).


GreLI:
I don't understand the use case of it. Anyone can do this by self, why add it for everyone? Also, extra dependency is not acceptable.

PR #1173, opened 2019-10-29
Add multipassCount to info object; skip subsequent passes in prefixIds plugin

strarsis:
This PR adds `multipassCount` to the `info` object that is passed to the plugins.
This allows plugins to know whether they already processed a SVG node/document and how many times. This is necessary for the `prefixIds` plugin to skip in subsequent passes when multipass is used to prevent unintended, multiple prefixing (see https://github.com/svg/svgo/issues/1148).

The `info` object is perfectly suited for this kind of extra/runtime information for the plugins, 
and it is already used by `prefixIds` plugin for obtaining the path of the SVG (node) it is processing.

### DIFF
--- a/lib/svgo.js
+++ b/lib/svgo.js
@@ -38,6 +38,7 @@ SVGO.prototype.optimize = function(svgstr, info) {
                     return;
                 }
 
+                info.multipassCount = counter;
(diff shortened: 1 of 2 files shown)


GreLI:
released in v1.3.1

PR #1177, opened 2019-10-30
Add multipass tests

strarsis:
@GreLI:
- This PR adds tests for `multipass` in `svgo`. Currently, there aren't any tests for `multipass` **at all.**
- It also fixes a bug with `multipassCount` that was uncovered with these newly added multipass tests.
- A test for the `prefixIds` plugin with `multipass` is added.
- A test for file name handling of the `prefixIds` plugin is added (it is very similar to the test for `prefixIds` with `multipass`, so it seems to be a nice addition).
- A test for the `addAttributesToSVGElement` plugin with `multipass` is added.
- Adds a test (albeit retroactively) for invoking svgo.optimize() without passing it an info object.

@GreLI: Additional points:
- During adding the `multipass` tests I noticed that it doesn't seem to be possible to use a `plugins` property for the object passed to `svgo()`, it is simply ignored. `enable` can't be used for plugins that require extra configuration, like the `addAttributesToSVGElement` plugin, so I had to pass a raw config string (see https://github.com/svg/svgo/pull/1177/files#diff-58d889243ada560f65a0a1826461d5deR228).
- Also for better isolating the `prefixIds` plugin related tests, I disabled all other the plugins. But as `plugins` key seems to be ignored, I had to pass the `plugins` from `svgo` `defaults` to `disable` (see https://github.com/svg/svgo/pull/1177/files#diff-58d889243ada560f65a0a1826461d5deR177).

Edit: What about using a separate `tests` folder for the plugin tests with `multipass` enabled?


strarsis:
@nolanlawson: Sorry for the hassle. 🐱

dphil:
With this approach, the first pass has multipassCount as `undefined`, rather than 0. With the way multipassCount is currently used (the check in prefixIds), it makes no difference, but initializing it to 0 would likely be more appropriate and robust.

strarsis:
@dphil: `multipassCounter` is now initialized to 0 at the start:
https://github.com/svg/svgo/pull/1177/commits/ade4a8ecfabdd038fba1b3c8dd5fb9e0ef59199b#diff-2338951073175db78c5b00c29d4971e1R26

I added the `undefined` `info` object test from the other PR to this PR because it fits the whole issue well.

strarsis:
Strange, the PR passes the tests on all node versions except for 6.x:
https://travis-ci.org/svg/svgo/jobs/632515808?utm_medium=notification&utm_source=github_status
`css-select` complains about a selector. This doesn't make sense. Why does this only happen on node 6.x? Why a selector that doesn't match? Is the order different in the older node 6.x?

strarsis:
I noticed that the build fails for node v6. This is not caused by this PR but is an actual issue with the current `svgo` master.
PR for current master to fix failing build on node v6: https://github.com/svg/svgo/pull/1208

kylemh:
Hey @strarsis I'm maintaining a fork of this repo and brought this PR into it.

When using a config that includes the following:
```json
{
  "plugins": [
    {
      "prefixIds": {
        "delim": "_"
      }
    },
    {
      "cleanupIDs": { "minify": true }
    },
  "multipass": true
}
```

The filename is not applied as a prefix.

strarsis:
@kylemh: I got an email notification from GitHub that the `prefixIds` plugin doesn't work correctly with multipass and specific options. I added a test for this set of options and the test passes. Are you using a specific node version? There were node-specific issues in the past.

kylemh:
I was hasty and incorrect. It had to do with the difference of build-time implementation in one file versus the run-time CLI usage again source files in their individual locations.

strarsis:
@kylemh: Well, it never hurt to have some more tests.

kylemh:
I’ll appreciate it in my fork!

TrySound:
Hi @strarsis Could you please rebase?

kylemh:
Glad you got added to the team @TrySound

MrSweGyii:
ClosedSent from my Huawei phone-------- Original message --------From: Bogdan Chadkin <notifications@github.com>Date: Thu, Feb 18, 2021, 5:43 PMTo: svg/svgo <svgo@noreply.github.com>Cc: Subscribed <subscribed@noreply.github.com>Subject: Re: [svg/svgo] Add multipass tests (#1177)
Closed #1177.

PR #1178, opened 2019-10-30
Fix error "Cannot set property 'multipassCount' of undefined"

Y2-MosheHarush:
Error: postcss-svgo: TypeError: Cannot set property 'multipassCount' of undefined
eg: https://github.com/svg/svgo/issues/1174


GreLI:
Fixed per #1176

PR #1182, opened 2019-10-30
Add test for invoking svgo.optimize without an info object

strarsis:
This PR adds a test (albeit retroactively) for invoking `svgo.optimize()` without passing it an `info` object.

Related issue: https://github.com/svg/svgo/issues/1174


strarsis:
This test has been added to https://github.com/svg/svgo/pull/1177 instead.

PR #1203, opened 2019-12-31
Update convertPathData.js

topcss:
Fix the bug: Vue failed to publish


GreLI:
Can you provide an example SVG, what does it fix?

topcss:
> Can you provide an example SVG, what does it fix?

An error occurred while compiling the project. It is estimated that the passed in parameter is not of type Number. So, I think we need to add a type conversion.

PR #1206, opened 2020-01-03
Fix multipassCount initialization

dphil:
The previous implementation initialized `info.multipassCount` in the optimize callback, and before counter is incremented. Thus, on the first pass, it is `undefined`, on the second it is 0, third it is 1, etc.

This is problematic since the prefixIds plugin multipass check will pass for both undefined and 0, causing it to run twice (as opposed to once, which was the intention of this feature to prevent accumulating prefixes).


dphil:
Duplicate of #1177 , which implements a similar fix for this issue, and adds tests as well.

PR #1219, opened 2020-02-13
Fix erroneus omission of attribute corrections

writtmeyer:
The cleanupNumericValues plugin contained a condition that causes the
correction of values to be scrapped even though the correction should be
applied. A simply example would be the omission to change the value of
the width attribute, if the value is "20mm".
This condition was removed and the remaining three lines simplified into
two.

The commit adds two test files to verify the new behaviour (these files would fail without the fix).


TrySound:
Hi. I think the point of this option is to convert only if it helps to reduce svg size. This is why you see `string.length` check. Does this give you some visual regression?

PR #1223, opened 2020-02-26
Add 'Shrink Me' link to the README.md file

Safi1012:
I added the web tool 'Shrink Me' as an additional way to use SVGO on the web.

Shrink Me:
https://shrinkme.app/


TrySound:
It's not obvious "shrink me" is using svgo. It's not open source right? Let me know if I'm wrong.

PR #1231, opened 2020-03-26
FIX: convertStyleToAttrs: regexp string escaping for `Important rule`

dr-js:
so the resulting regexp will change from:
`/(\s*!important(?![-(w]))?/`
to:
`/(\s*!important(?![-(\w]))?/`


TrySound:
Thanks! This is already fixed.

PR #1235, opened 2020-03-27
FIX: test: usage of `should` & drop extra `return`

dr-js:
simple test script:
```js
require('should')

try {
  console.log('\n# TEST "value.should.be.false" vs "value.should.be.false()"')
  const a = true
  console.log('assign')
  a.should.be.false // no effect
  console.log('pass 1')
  a.should.be.false() // check and throw error
  console.log('pass 2')
} catch (error) {
  console.log(`stopped with error: ${error}`)
}

try {
  console.log('\n# TEST "value.should.exist()"')
  const a = 1
  console.log('assign')
  a.should.exist() // result in no method error
  console.log('pass 1')
} catch (error) {
  console.log(`stopped with error: ${error}`)
}

try {
  console.log('\n# TEST "should.exist(value)"')
  should.exist(1) // check and pass
  console.log('pass 1')
  should.exist(undefined) // check and error
  console.log('pass 2')
} catch (error) {
  console.log(`stopped with error: ${error}`)
}
```

### DIFF
--- a/test/config/_index.js
+++ b/test/config/_index.js
@@ -39,29 +39,29 @@ describe('config', function() {
             removeRasterImages = getPlugin('removeRasterImages', config.plugins);
 
         it('should have "multipass"', function() {
-            return config.multipass.should.be.true;
+            config.multipass.should.be.true();
         });
 
         it('removeDoctype plugin should be disabled', function() {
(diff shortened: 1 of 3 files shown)


TrySound:
Refactored with chai

PR #1240, opened 2020-04-09
Null checkin strongRound

kanishk-anand-oyo:
Build process failed when we updated `svgo` package, and on inspection we found that a null check was missing before parsing the floating number

### DIFF
--- a/plugins/convertPathData.js
+++ b/plugins/convertPathData.js
@@ -772,7 +772,7 @@ function getIntersection(coords) {
  */
 function strongRound(data) {
     for (var i = data.length; i-- > 0;) {
-        if (data[i].toFixed(precision) != data[i]) {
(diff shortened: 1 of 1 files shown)


TrySound:
Hi! Could you please add test?

SethFalco:
Closing this as the relevant code is as implementation detail, and should no longer ever receive `null` or `undefined` as a parameter nor as an array item in its parameter.

It may be that at the time the PR was opened, it was possible for `null`/`undefined` values to appear here, I can see the JS Doc type was just `Array` before. However, that shouldn't be the case anymore. This is already enforced during development with TypeScript, so I see no need to add runtime checks.

Just to be sure, I ran our regression tests (4,643 SVGs) with an explicit check for `null`/`undefined`, and it never logged either.

If this is still a problem, feel free to share an SVG that reproduces the issue with the latest version of SVGO, and I'll take a look!

Issue #1247, opened 2020-04-23
Feature Request:Round colors


17jiangz1:
Option (disabled by default) to round colours from #xxxxxx to #xxx using the closest #xxx colour in the CIELAB colour space.

jameswilddev:
I was after this too (using SVG in extreme minification projects).  Might give it a try.  Was just going to mask off the least significant bits of each channel, e.g. #123456 becomes #135.

SethFalco:
Hey! Thanks for reporting the issue. I'll close this as a duplicate of https://github.com/svg/svgo/issues/884, but I've made a note of the PR there.

Issue #763, opened 2017-07-13
Suggestion: Change colors slightly so they can be shorter


davidleston:
Similar to how cleanupNumericValues' option to reduce float precision: configure how much colors are allowed to change in order to be shorted.

Currently `#333333` is converted to `#333`. A configured value of `1` would replace `#323333` and `#333433` with `#333`. Would also implement detecting distance to [short color names](https://github.com/svg/svgo/blob/d0a511fc376b07c0a5874387f7b64f57691c73f8/plugins/_collections.js#L2492).

I'd be happy to create a PR if this sounds like something that'd be accepted. Would you like this as a new plugin or a part of convertColors plugin? Also, any suggestions on the name of the parameter to configure and a default value?

GreLI:
Nice idea! `convertColors ` would be nice place for the optimization. Perhaps, a good name would be something like ‘convertClose’.

The only thing I'm considering is what the threshold value. 0x10? 0.01%? An option value?

Issue #123, opened 2013-06-25
"Non-whitespace before first tag." error


mahish:
did $ svgo logo.svg 
got Non-whitespace before first tag

Did not find any info anywhere about this error msg. Can somebody give me a clue? 
Thanks.

AndyOGo:
Same here on SVG's which are run through svgo beforehand.
I double checked to have BOM removed and I check my files with an xml validator.

rantiev:
Heh, it seems i have anything else. 
svg is valid. UTF-8 without bom encoding. No any other svg loaders for fonts. But the error exists.

a-x-:
In my case I realized, that `create-react-app` have already good svg loader inside: svgr

And it's working out of the box:
```jsx
import { ReactComponent as Logo } from './logo.svg';
// or
import logoPath from './logo.svg';
```

PR #1248, opened 2020-04-25
Implement hex truncation.

jameswilddev:
This adds a flag to the convertColors plugin which is off by default which drops every second character in a 6-character hex color.  In a lot of cases 12bpp is enough and this can make a dent in file size.

Should help with #1247 and #763.

### DIFF
--- a/plugins/convertColors.js
+++ b/plugins/convertColors.js
@@ -11,7 +11,8 @@ exports.params = {
     names2hex: true,
     rgb2hex: true,
     shorthex: true,
-    shortname: true
+    shortname: true,
+    truncatehex: false
 };
 
 var collections = require('./_collections'),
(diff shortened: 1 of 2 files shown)


jameswilddev:
Hey @TrySound !  Was this feature added by someone else, or was my branch out of date?  I can make another version if you'd like some changes to my implementation.

TrySound:
Hey, I renamed master to main. I guess this closed this PR automatically. Sorry for that.

jameswilddev:
Ah cheers I'll catch that up, thanks!

TrySound:
I actually plan to use colord package for minification. @omgovich recently added utility which already used by cssnano.

TrySound:
@jameswilddev Btw could you help me to understand what your change does? Caz these regexps look equivalent to me
```
    regHEX = /^\#(([a-fA-F0-9])\2){3}$/,
    truncatableRegHEX = /^\#(?:([a-fA-F0-9])[a-fA-F0-9]){3}$
```

jameswilddev:
If I'm understanding `regHEX` correctly it will only shorten the color code if it's not a lossy change, e.g. `#112233` would be shortened but `#112133` would not.

The new code path added by this PR would _always_ shorten a hex color code, even if that means a lossy change; both `#112233` and `#112133` would shorten to `#123`.

This generally isn't as noticeable as a 24bpp -> 12bpp reduction would be in say a bitmap image as most SVGs use colors for flat shaded shapes or the key points of interpolated colors (gradients/etc.) but saves 3 characters per hex code.

TrySound:
Gotcha. Alright, we can implement this after migrating to colord.

PR #1260, opened 2020-06-09
Add plugin multiCollapseGroups

LuRsT:
collapseGroups is a great plugin, but unfortunately, it only deals with
single elements inside a <g>, which is why I made this plugin.

This will behave in the same way as `collapseGroups` but will also
transfer the attributes from the group into all child nodes.

### DIFF
--- a/plugins/multiCollapseGroups.js
+++ b/plugins/multiCollapseGroups.js
@@ -1,0 +1,96 @@
+'use strict';
+
+exports.type = 'perItemReverse';
+
+exports.active = true;
+
+exports.description = 'collapses useless groups';
+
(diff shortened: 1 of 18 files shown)


LuRsT:
Is there any way that I can use this plugin without it being part of the svgo package? Is there such a mechanism to use an "external plugin"?

jorisa:
I think you can just take the file and chuck it in your npm svgo install location (`npm list -g --depth=0` to show where that is)

LuRsT:
I guess that would work @jorisa, but I'd have to write some programatic way to keep chucking the file in the install location everytime I upgrade svgo.

PR #1271, opened 2020-07-31
Security Updates

georgehossa:
This solves some security issues with this dependencies:
- handlebars
- lodash


TrySound:
Not direct dependencies

PR #1272, opened 2020-08-01
convertPathData reduce move commands

jorisa:
Add optimization option to convertPathData that tries to minimize Move operations by sorting line elements to be adjecent. Particularly useful for optimizing SVG exports from some CAD data to be used by CNC software.

It's a bit of an edge case end not tested for all path commands, so disabled by default.

[Before-After.zip](https://github.com/svg/svgo/files/5010060/Before-After.zip)


jorisa:
Could anyone take a look at this? Would be nice if it can be merged.

TrySound:
HI! Test cases would help to understand the feature and prevent from breaking later. Could you please add some?

Issue #930, opened 2018-03-25
Using attribute selection with css-select is broken


ydaniv:
`getAttributeValue()` for a missing attribute returns `null`, but then this is compared to `!== undefined` in the base adapter in `hasAttr()` which returns `true`.

So either you return `undefined` as the base adapter expects, or you implement `hasAttr()` here.

I can send a PR, but would like to hear if there's any specific reason for preferring one over the other.

Thanks!

TrySound:
Fixed in https://github.com/svg/svgo/pull/1395

PR #1286, opened 2020-09-18
getAttributeValue returns undefined rather than null

acrazing:
Ref https://github.com/svg/svgo/issues/930

It should return `undefined` rather than `null` according to <https://github.com/nrkn/css-select-base-adapter/blob/master/index.js#L100>.

### DIFF
--- a/lib/svgo/css-select-adapter.js
+++ b/lib/svgo/css-select-adapter.js
@@ -43,7 +43,7 @@ var svgoCssSelectAdapterMin = {
     // getAttributeValue: ( elem:ElementNode, name:String ) => value:String
     // returns null when attribute doesn't exist
     getAttributeValue: function(elem, name) {
-        return elem.hasAttr(name) ? elem.attr(name).value : null;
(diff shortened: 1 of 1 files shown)


TrySound:
Could you please provide some test case?

lib/svgo/css-select-adapter.js
```diff
@@ -43,7 +43,7 @@ var svgoCssSelectAdapterMin = {
     // getAttributeValue: ( elem:ElementNode, name:String ) => value:String
     // returns null when attribute doesn't exist
     getAttributeValue: function(elem, name) {
-        return elem.hasAttr(name) ? elem.attr(name).value : null;
+        return elem.hasAttr(name) ? elem.attr(name).value : void 0;
```
XhmikosR, line 46: I'd just use `undefined` which is clearer even to newcomers.

TrySound:
Fixed in https://github.com/svg/svgo/pull/1395

PR #1290, opened 2020-10-02
Replace util.promisify with fs.promises

TrySound:
Node 10 add support for promisified fs methods.
https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api

In this diff I replaced util.promisify package with the new methods.
Though this requires to bump end-of-life node versions.


TrySound:
cc @GreLI

TrySound:
Oh, gosh. I just installed util.promisify in clean project and got 8mb in node_modules.

PR #1291, opened 2020-10-02
Replace mkdirp with fs.mkdir with recursive flag

TrySound:
Ref https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_mkdir_path_options_callback

Node 10 add "recursive" flag support to fs.mkdir which behaves similar
to mkdirp package.

Though this requires to bump node support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "should": "~13.2.3"
   },
   "engines": {
-    "node": ">=10.0.0"
+    "node": ">=10.12.0"
   },
   "license": "MIT"
 }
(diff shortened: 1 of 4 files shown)


TrySound:
cc @GreLI

package.json
```diff
@@ -73,7 +72,7 @@
     "should": "~13.2.3"
   },
   "engines": {
-    "node": ">=4.0.0"
+    "node": ">=10.0.0"
```
realityking, line 75:
```suggestion
    "node": ">=10.12.0"
```

According to the [docs](https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback), this was only added in 10.12

PR #1292, opened 2020-10-02
Drop object.values polyfill

TrySound:
Ref https://packagephobia.com/result?p=object.values

Object.values is supported since node 7. It can be removed with major
bump especially since it's quite heavy.


TrySound:
cc @GreLI

PR #1293, opened 2020-10-09
Replace coa with commander

TrySound:
Ref https://packagephobia.com/result?p=coa https://packagephobia.com/result?p=commander

Coa was not updated for 2 years already and has outdated dependencies.

In this diff I replaced it with commander which has simpler syntax, none
dependencies and require arguments out of the box. Also array argument
is set syntactically which means more user-friendly "help".

### DIFF
--- a/bin/svgo
+++ b/bin/svgo
@@ -1,3 +1,6 @@
 #!/usr/bin/env node
 
-require('../lib/svgo/coa').run();
+const { program } = require('commander');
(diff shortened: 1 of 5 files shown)


manbearwiz:
Well this ended up being a good call https://github.com/veged/coa/issues/99

TrySound:
Yep, sadly couldn't port to v1.

PR #1296, opened 2020-10-16
Add plugin to normalize the case of colors

melusc:
It bothers me when the case of colors isn't consistent across a file so I wrote this plugin that takes hex, rgb, and hsl colors and sets the correct case. This doesn't modify or set the case of css colors, though.
The different types of colors each have their own setting where `upper` or `lower` set the set the respective case or `false` to not change anything.
Of course, this is only for readability and consistency and doesn't reduce the file size, similarly to sortAttrs.


XhmikosR:
I'm definitely interested in this since lowercasing hex colors is one of the things I have to do manually. Lowercase hex colors will probably result in (slightly) better gzip compression.

XhmikosR:
@melusc Need to add the new plugin to https://github.com/svg/svgo/blob/master/plugins/plugins.js :)

XhmikosR:
You should probably rebase the branch properly and keep only the needed patches. Now you seem to have patches from master.

melusc:
Dang, I thought that might happen. I rebased master onto my branch and now it shows in changed files.

How would I do that correctly next time?

XhmikosR:
You can fix it by doing this

```sh
git fetch -all
git checkout normalizeColorCase
git rebase -i upstream/master # assuming upstream is the upstream repo
# make sure the commits are fine and then
git push -f
```

melusc:
What do you mean by `upstream/master`? Because I get `fatal: invalid upstream 'upstream/master'`.

XhmikosR:
`upstream` is this repo. `origin` is probably your repo.

XhmikosR:
@melusc I'd say if you can't do a proper rebase based on my instructions above, you better close the PR and make a new one from scratch. Just make sure tests pass and you have adapted any code to the recent master changes.

melusc:
I tried what you said but I think it would add even more commits, so I'll make a new PR when I get a bit of time.

XhmikosR:
When you do `git rebase -i` you get the chance to drop any unrelated commits. It's not so hard, you just need to read about rebase and understand how it works. Rebasing is mandatory knowledge when collaborating on forks :)

PR #1311, opened 2020-12-22
Fixed reference in config.js

josephtyler:


TrySound:
It's correct now. Though in the next major release defaults will be hardcoded in js.

PR #1312, opened 2020-12-29
Improve path merging for non-filled paths

yGuy:
If two paths are not filled, it does not matter whether they intersect or not and they can be merged. This helps with the common case of many different "lines" (which have no fill) that will fail to be merged if they intersect, which is common in many graphics. 

With this change many of my files I got another 10% size reduction, because path merging almost always failed for my non-filled paths in "crowded" svgs..

### DIFF
--- a/plugins/mergePaths.js
+++ b/plugins/mergePaths.js
@@ -58,7 +58,8 @@ exports.fn = function(item, params) {
                 prevPathJS = path2js(prevContentItem),
                 curPathJS = path2js(contentItem);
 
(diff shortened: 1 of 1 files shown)


TrySound:
Hi please rebase and add tests

hashimaziz1:
@yGuy Any update on this? Sounds promising and would be good to get it merged in.

yGuy:
Sorry, no time for this, right now. But anyone can rebase and add tests. It doesn't have to be me.

PR #1316, opened 2021-01-26
Added file detection for *.SVG files when working with folders

AlpayY:
Currently, files named *.SVG are not being detected under Windows because of the case-sensitivity. This commit includes these in addition to *.svg files.


janmoritz (commented):
A simpler solution would be to include the case sensitivity in the RegEx of ```regSVGFile = /\.svg$/i;``` by adding the _i_.

AlpayY:
Great idea! I tried a variation of this idea first, but couldn't get it to work. Your solution is better.

PR #1326, opened 2021-02-15
Provide es module bundle for browser

TrySound:
The was a lot of requests to provide browser compatible module. In this
diff I added `svgo/dist/svgo.browser.js` bundle and tested it with
playwright.


TrySound:
@caub @jakearchibald you might be interested in this

PR #1331, opened 2021-02-18
Fix 'ReferenceError: indent is not defined'

Brooooooklyn:
BTW, `eslint` could prevent this type of mistake.


TrySound:
Good catch, thanks!

PR #1341, opened 2021-02-18
Fix `new Buffer` usage

XhmikosR:
Node.js 10.x which is the minimum supported version supports `Buffer.from`


TrySound:
Thanks a lot!

PR #1345, opened 2021-02-19
Setup recommended eslint preset

TrySound:
This is usually enough to prevent silly mistakes.
Formatting will be done by prettier after merging pull requests.

cc @XhmikosR


XhmikosR:
It's a good start but IMHO there are plenty of stuff that aren't checked. You are still using var for example for in isn't guarded etc, etc.

Personally, I don't like reinventing the wheel, so I'd go with xo which also supports prettier.

Just my 2 cents, since this a subjective topic :)

TrySound:
I understand a lot not checked. Though I like wheels. And I don't feel comfortable with such big packages
https://packagephobia.com/result?p=xo

XhmikosR:
I understand that, I don't like big deps either. That being said, you are certainly not checking for pretty important things.

Sure, you could add eslint security and other configs, but then you are reinventing the wheel and you won't achieve all the rules.

Anyway, it's your call but issues like #1341 could have been caught.

.editorconfig
```diff
@@ -1,19 +0,0 @@
...
-insert_final_newline = true
-
-[Makefile]
-indent_style = tab
-
-[package.json]
-indent_size = 2
-
-[*.yml]
-indent_size = 2
```
yisibl, line 19: Why removed .editorconfig? This should not conflict with ESLint.

yisibl:
Can we use https://prettier.io/?

TrySound:
Yes, I will setup prettier after dealing with pull requests to not introduce too much conflicts.

Issue #1317, opened 2021-02-01
[Bug]Path compression lost space


yisibl:
Original svg

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 750 1125">
    <path d="M448.6,997.7c.2,5.6,16.6,9.4,36.5,8.4s35.8-6.2,35.5-11.8a1.7,1.7,0,0,0-.1-.7c-1.5-5.2-17.3-8.6-36.4-7.7s-34.4,5.8-35.5,11.1Z" fill="green" />
</svg>
```

After [svgo](https://github.com/svg/svgo) compression, the spaces are removed, and an error will be reported:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 750 1125">
    <path d="M448.6 997.7c.2 5.6 16.6 9.4 36.5 8.4s35.8-6.2 35.5-11.8a1.7 1.7 0 00-.1-.7c-1.5-5.2-17.3-8.6-36.4-7.7s-34.4 5.8-35.5 11.1z" fill="green" />
</svg>
```

### Use svgcleaner

```svg
<svg viewBox="0 0 750 1125" xmlns="http://www.w3.org/2000/svg">
  <path d="m448.6 997.7c.2 5.6 16.6 9.4 36.5 8.4s35.8-6.2 35.5-11.8a1.7 1.7 0 0 0 -.1-.7c-1.5-5.2-17.3-8.6-36.4-7.7s-34.4 5.8-35.5 11.1z" fill="#008000"/>
</svg>
```

`00-.1` Spaces are required here.

![image](https://user-images.githubusercontent.com/2784308/106470467-75172a80-64db-11eb-80b7-2a2b3a69007a.png)

TrySound:
Hi @yisibl. Where the error is reported? I see no errors or visual differences in chrome and firefox.

Jupi007:
You can see visual problems here: https://github.com/ubuntu/yaru/issues/2630

TrySound:
@Jupi007 I see. Could you help me to find this particular icon?

Jupi007:
In fact almost all icons of the icon theme was bugged using svgo.
For example with **folder-documents-symbolic**:
- source file: https://github.com/ubuntu/yaru/blob/master/icons/src/scalable/places/folder-documents-symbolic.svg
[...]

Issue #1189, opened 2019-11-11
Overzealous whitespace removal in paths


davidar:
Sometime in the past year, svgo has started removing more whitespace from paths than it used to (it looks like it may be since commit 258fecf?), in a way that breaks the SVG for some viewers. For example, with this image:

```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="573.331" height="367.563" viewBox="0 0 429.998 275.672"><defs><symbol overflow="visible" id="c"><path d="M1.922-.828h3.437V0H.734v-.828c.375-.383.883-.906 1.532-1.563a52.138 52.138 0 0 0 1.218-1.265c.313-.352.532-.649.657-.89.125-.25.187-.493.187-.735 0-.383-.137-.696-.406-.938-.274-.25-.625-.375-1.063-.375-.312 0-.64.059-.984.172-.336.106-.7.266-1.094.484v-1a8.838 8.838 0 0 1 1.11-.359c.343-.082.66-.125.953-.125.75 0 1.347.188 1.797.563.457.375.687.882.687 1.515 0 .305-.058.59-.172.86-.117.261-.32.574-.61.937-.085.094-.343.371-.78.828-.438.45-1.055 1.078-1.844 1.89zm0 0"/></symbol></defs><g transform="translate(-11.094 -12.11)"><use height="100%" width="100%" y="208.938" x="16.004" xlink:href="#c"/></g></svg>
```

svgo v1.3.2 modifies the path in these places:

```
Before: a52.138 52.138 0 0 0 1.218-1.265
After:  a52.138 52.138 0 001.218-1.265
Before: a8.838 8.838 0 0 1 1.11-.359
After:  a8.838 8.838 0 011.11-.359
```

Some viewers (e.g. Firefox) seem to be ok with this, but others (e.g. `rsvg-convert`) are not:

Before:

![Screenshot_20191111_173105](https://user-images.githubusercontent.com/24291/68569110-0a7a6900-04a9-11ea-9a0c-2942755fadf3.png)

After:

![Screenshot_20191111_173027](https://user-images.githubusercontent.com/24291/68569080-f3d41200-04a8-11ea-9d8a-dd907c1d7c6d.png)

davidar:
Oops, looks like this is a duplicate of #1137 and #1164

Issue #1183, opened 2019-10-31
SVGO messes SVGs 


Hlsgs:
I'm using SVGO though "grunt-contrib-imagemin". I've reverted to an earlier package.json and package-lock.json that installs SVGO 1.2.2 and everything is fine, so the regression must have creeped up after that version.

Here is an example before/after that was already optimized multiple times with no issue with versions <= 1.2.2. Now I just ran my build process with 1.3.2 and it got ruined:

Before:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 111.946 75.259"><path id="img" d="M262.946 1229.358H151V1154.1h111.946zm-9.328-65.846h-93.29v37.622l23.322-23.517 27.841 26.018 21.137-18.8 20.99 19.43zm0 51.727l-21.141-18.946-20.844 18.946-28.123-26-23.176 23.37v7.341h93.29v-4.706zm-41.981-23.5a9.407 9.407 0 1 1 9.329-9.412 9.444 9.444 0 0 1-9.329 9.409z" transform="translate(-151 -1154.099)" fill="#fff" fill-rule="evenodd"/></svg>
```

After:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 111.946 75.259"><path id="img" d="M262.946 1229.358H151V1154.1h111.946zm-9.328-65.846h-93.29v37.622l23.322-23.517 27.841 26.018 21.137-18.8 20.99 19.43zm0 51.727l-21.141-18.946-20.844 18.946-28.123-26-23.176 23.37v7.341h93.29v-4.706zm-41.981-23.5a9.407 9.407 0 119.329-9.412 9.444 9.444 0 01-9.329 9.409z" transform="translate(-151 -1154.099)" fill="#fff" fill-rule="evenodd"/></svg>
```

The problem seems to be changing `9.407 0 1 1 9.329-9.412` to `9.407 0 119.329-9.41`

Thank you!

GreLI:
What viewer do you use? It's a valid optimization, introduced in v1.3. Whitespace is optional after `arcto` flags, it's in SVG 1 specification. However, as always, you can turn it off as described in https://github.com/svg/svgo/issues/1137#issuecomment-517738958.

Hlsgs:
You are correct in assuming modern viewers(Firefox, Chrome) view the result properly, but up to date Illustrator outright fails to understand the optimized result(you can check the example I gave). Does Illustator have a non-conformant SVG engine?

[...]

Issue #1164, opened 2019-10-11
Latest SVGO is creating SVG files that some renderers can’t handle


viljamis:
**SVGO version:**
```
 svgo@1.3.0
```

**I'm submitting a:**
[x] bug report
[ ] feature request
[ ] support request

**Current behavior:**
It seems like the latest version of SVGO is creating SVG files that some renderers can’t handle. 

A little backstory to explain the situation: 

Our team relies on SVGO automation in our design system to optimize SVG icons used. In addition to developers, those icons are used by the design team as well inside vector based tools such as Sketch. 

Using the latest version of SVGO the icons look like this when opened in Sketch app (left side we have an icon optimized with SVGOMG using SVGO v1.0.3 and on the right using SVGOMG that has been updated to the latest v1.3.0):

<img width="876" alt="Screen Shot 2019-10-11 at 7 11 37" src="https://user-images.githubusercontent.com/671836/66654542-1d4c1480-ec43-11e9-8d30-a9dce6937902.png">

**Expected behavior:**
SVG files produced should work in any vector based design tool in addition to browsers.

**Steps to reproduce:**
1. Download the following icon https://cdn.jsdelivr.net/npm/@duetds/icons@2.0.3/lib/assets/action-car-transfer.svg
2. Open in Sketch to see that it works just fine.
3. Now use f.ex. https://jakearchibald.github.io/svgomg/ (SVGOMG now uses 1.3.0 as well) to optimize it with just the default settings.
4. Download and try to open in Sketch. You get the above broken results.

**Other information:**
There’s a possibly related issue in SVGOMG repository as well: https://github.com/jakearchibald/svgomg/issues/167

GreLI:
What renderer do you use? Probably, it's not fully compliant, since browser render its ok. Have you tried option like described in https://github.com/svg/svgo/issues/1137#issuecomment-517738958?

wbervoets:
FYI importing an SVG file optimized with SVGO 1.3 + standard setting did not work, using the following config did work :-)

{
    "plugins": [
        { "convertPathData": { "noSpaceAfterFlags": false } },
        { "mergePaths": { "noSpaceAfterFlags": false } }
    ]
}

viljamis:
@GreLI @wbervoets To be completely honest, I don’t know what’s the rendering engine here, but this issue is fixed with what @wbervoets suggested above too. So:

```json
{
    "plugins": [
        { "convertPathData": { "noSpaceAfterFlags": false } },
[...]

PR #1353, opened 2021-02-20
After ‘arcto’ path command(a or A), spaces are no longer removed by default

yisibl:
## Background

In the [svg elliptical arc curve commands](https://www.w3.org/TR/SVG11/paths.html#PathDataEllipticalArcCommands), the white space after `large-arc-flag` and `sweep-flag` can be removed. Svgo was implemented here(https://github.com/svg/svgo/commit/258fecfa6b4a037d98ad2614f5c9573728f6959d) before and enabled by default. 

[The idea of the implementer](https://github.com/svg/svgo/issues/1137#issuecomment-518351171) @GreLI  at the time was:

> It's a totally standard syntax from a 8-years old specification (and was introduced in a working draft a year before). If some software has a problems with it—those should be addressed to that software.

However, [more and more people report that there will be problems](https://github.com/svg/svgo/issues/1137 ). I think that as an svg compressor, security and compatibility are the most important. Although browsers all support omitting the white space after large-arc-flag and sweep-flag, mainstream vector software such as Sketch and Adobe Illustrator do not support it. Even if these software are supported, they are paid software, and the upgrade cost for users is huge. As a designer, I can't understand why the SVG file compressed by svgo will make an error when reopened. We can't think about the problem from the perspective of a developer who can write code. Even for a developer, the rules of path data are obscure.

image

So this PR disables the `noSpaceAfterFlags` option by default. Although the compressed file size increases slightly, I think it is worth it. If your SVG file is only used in the browser or pursues the ultimate file size, you can enable it through the following options:

```json
{
    "plugins": [
        { "convertPathData": { "noSpaceAfterFlags": true } },
        { "mergePaths": { "noSpaceAfterFlags": true } }
    ]
}
```

Fixed:  https://github.com/svg/svgo/issues/1317 https://github.com/svg/svgo/issues/1138  https://github.com/svg/svgo/issues/1164 https://github.com/svg/svgo/issues/1183 https://github.com/svg/svgo/issues/1189


yisibl:
@TrySound PTAL

TrySound:
Thanks a lot @yisibl!

Jupi007:
Thanks @yisibl 👍

yisibl:
Can you help me change the git email and username of the master branch commit? I used the wrong company mailbox. Or I resubmit a PR. Thanks a lot.

TrySound:
Sorry, don't want to rewrite master history especially after release. Looks like both are here. Is it a big problem?
image

yisibl:
You are so fast! That's it for now.

GreLI:
It's a fame, that software doesn't support  SVG well 😒

Issue #1362, opened 2021-02-22
[BUG] Optimized SVG is different than source Escudo de Agulo.svg


sk-:
Using the (wip) integration test tool we identified an issue while optimizing the file `Escudo de Agulo.svg` with the default settings.

## SVGO version
2.0.2

## Differences

The dragon is completely broken and also some of its colors changed.
## Sources
Original:

[source.svg.txt](https://github.com/svg/svgo/files/6022627/source.svg.txt)
Optimized:

[optimized.svg.txt](https://github.com/svg/svgo/files/6022626/optimized.svg.txt)
## Renders
Here are the two rendered images for comparison:

Original:

![source](https://user-images.githubusercontent.com/911768/108722601-19e8be80-7502-11eb-8e68-4a4b7e58d1fb.png)

Optimized:

![optimized](https://user-images.githubusercontent.com/911768/108722582-16553780-7502-11eb-8ed8-1f81989ec76f.png)
Diff:

![diff](https://user-images.githubusercontent.com/911768/108722572-13f2dd80-7502-11eb-9798-eaaf31a9a301.png)

TrySound:
Fixed in https://github.com/svg/svgo/releases/tag/v2.1.0

Issue #1360, opened 2021-02-22
Please establish lossless and secure compression as the primary idea


yisibl:
As an svg compressor, I think it is most important not to change the display effect of the svg itself, which means that the compressed result cannot change the shape, position, size, color, etc. (svgcleaner has a [detailed description](https://github.com/RazrFalcon/svgcleaner/blob/master/docs/testing_notes.rst#rendering-errors)). As svgo is adopted by more and more non-browser terminals ([Sketch](https://github.com/sketch-hq/svgo-compressor), Adobe Illustrator, other back-end languages), this will become more and more difficult, because each side implements svg differently. So we urgently need to introduce Correctness test like [svgcleaner](https://github.com/RazrFalcon/svgcleaner#correctness) to help svgo become the best svg compressor.

The following work is a good start:

- [x]  [The `noSpaceAfterFlags` option is disabled by default](https://github.com/svg/svgo/pull/1353) @yisibl 
- [x] [Test regressions with W3C SVG 1.1 Test Suite](https://github.com/svg/svgo/pull/1357) @TrySound 
- [ ] Add sample documents and compatibility instructions for each option of svgo, Similar to:
![image](https://user-images.githubusercontent.com/2784308/108680243-942c3980-7528-11eb-919d-b04950d1ce2d.png)

cc @deepsweet

sk-:
May I add that we should change the default options to only include those plugins known/supposed to be lossless. For example `convertPathData` probably should not be a default plugin unless the default options are safe. I've also seen that gradients are usually changed.

TrySound:
Or we should make it safe.

mischnic:
Another case here is `cleanupIds`, which is generally good, unless you use something like `<use xlink:href="square.svg#square" x="20" fill="white" stroke="red"/>` where removing the IDs inside `square.svg` isn't safe.

TrySound:
We have now presets concept. I'm ok to implement different presets. Though would be good if somebody explore all plugins.
I'm currently busy preparing svgo 3.

SethFalco:
Closing this as we're now actively striving for this, but the issue isn't presented in a way that we can track progress and know when it's complete.

If you have concrete suggestions, you're encouraged to open another issue with what we can specifically do better. Meanwhile, I'm not at all dismissing the issue.

[...]

PR #1365, opened 2021-02-22
Disable convertStyleToAttrs plugin by default

TrySound:
Ref https://github.com/svg/svgo/issues/1362 https://github.com/svg/svgo/issues/1360

From the [spec](https://www.w3.org/TR/SVG11/styling.html#UsingPresentationAttributes):

> Presentation attributes have lower priority than other CSS style rules specified in author style sheets or ‘style’ attributes.

Though we replace inline styles with attributes without checking if
thare is any `<style>` element. This makes the plugin unsafe and it
should not be enabled by default.

cc @sk- @strarsis

### DIFF
--- a/plugins/convertStyleToAttrs.js
+++ b/plugins/convertStyleToAttrs.js
@@ -2,7 +2,7 @@
 
 exports.type = 'perItem';
 
-exports.active = true;
+exports.active = false;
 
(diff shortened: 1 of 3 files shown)


strarsis:
@TrySound: I agree that it should be disabled by default until better logic is added that takes the specificity into account.

PR #1369, opened 2021-02-23
Fork sax

TrySound:
There were a lot of PRs with switching to another xml parser because sax
is not maintained for almost 4 years already.

Though switching parser to solve a few problems may introduce many new
bugs. This is why we decided to fork sax and maintain own version.

For initial release I removed node streams support which allows to get
rid from some magic in browser bundle build config and removed
String.fromCodePoint polyfill.

Forked package is here https://github.com/svg/sax

### DIFF
--- a/lib/svgo/svg2js.js
+++ b/lib/svgo/svg2js.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var SAX = require('sax'),
+var SAX = require('@trysound/sax'),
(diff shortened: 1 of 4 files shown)


deepsweet (approved):

XhmikosR:
@TrySound shouldn't the forked sax package name be @svgo/sax?

TrySound:
We wait for `@svg` scope to be available.

PR #1370, opened 2021-02-23
Add TS support

sk-:
Added Typescript support via JSDoc comments for some files.

The code has really outdated type signatures, so in order to fully type the codebase the changes need to be incremental.

To check the types run:
`npx tsc`

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
   "scripts": {
     "test": "c8 --reporter=html --reporter=text mocha \"test/*/_index.js\" \"**/*.test.js\" --ignore=\"node_modules/**\"",
     "lint": "eslint .",
+    "types:check": "tsc",
     "test-browser": "rollup -c && node ./test/browser.js",
     "prepublishOnly": "rm -rf dist && rollup -c"
   },
@@ -114,7 +115,8 @@
     "mock-stdin": "^1.0.0",
     "playwright": "^1.8.1",
     "prettier": "^2.2.1",
-    "rollup": "^2.39.0"
+    "rollup": "^2.39.0",
+    "typescript": "^4.2.2"
   },
   "engines": {
     "node": ">=10.13.0"
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
   "scripts": {
     "test": "c8 --reporter=html --reporter=text mocha \"test/*/_index.js\" \"**/*.test.js\" --ignore=\"node_modules/**\"",
     "lint": "eslint .",
-    "types:check": "tsc",
+    "typecheck": "tsc",
     "test-browser": "rollup -c && node ./test/browser.js",
     "prepublishOnly": "rm -rf dist && rollup -c"
   },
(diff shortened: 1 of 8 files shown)


sk-:
Note: the failing test on NodeJS is due to an error while installing `playwright`. Probably we should be using a cache, to speed up the installation of dependencies.

XhmikosR:
@sk- how about adding a new npm script and add the script in GitHub Actions too?

BTW I plan to tweak the actions a bit later, move linting to a separate workflow and add caching too.

.tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "target": "es2018",
+    "lib": ["es2018"],
+    "allowJs": true,
+    "checkJs": true,
+    "strict": true,
+    "resolveJsonModule": true,
+    "noImplicitAny": false
+  },
+  "include": [
+    "lib/css-tools.js",
```
XhmikosR, line 14: shouldn't this be a wildcard like `"lib/**/*.js`?

package.json
```diff
@@ -109,7 +109,8 @@
     "mocha": "^8.3.0",
     "mock-stdin": "^1.0.0",
     "playwright": "^1.8.1",
-    "rollup": "^2.39.0"
+    "rollup": "^2.39.0",
+    "typescript": "^4.1.5"
```
XhmikosR, line 113: Might be worth updating to the latest version before landing the PR?

.tsconfig.json
sk-: Ideally yes, but the files have many issues and doing it all at once won't be possible. That's why I preferred to only add those files that were fixable with minimum changes.

Also, when you include one file it will also check all the transitive dependencies, so including any of the svgo entrypoints or config, will bring all the plugins as well.

package.json
sk-: Done.

sk-:
@XhmikosR PTAL. I updated the TS version, added a script to package json and the check to the github action.

The github action should probably use caching and be split so that we only check linting and types with one node version.

ryansmith94:
Hey guys, hope you're enjoying your weekend and not looking at this until some time in the week. You might want to consider using `emitDeclarationOnly` in your tsconfig ([docs here](https://www.typescriptlang.org/docs/handbook/compiler-options.html)). That would allow people to use the package with TS without having to use @types/svgo which is now quite outdated since v2.

XhmikosR:
@sk- can you rebase this please? I split the actions so types:check should be included in the lint script or something like that.

sk-:
@XhmikosR please take another look.

XhmikosR:
LGTM but I'll leave the final choice to @TrySound since I'm just a user myself :P

package.json
```diff
@@ -51,6 +51,7 @@
   "scripts": {
     "test": "c8 --reporter=html --reporter=text mocha \"test/*/_index.js\" \"**/*.test.js\" --ignore=\"node_modules/**\"",
     "lint": "eslint .",
+    "types:check": "tsc",
```
TrySound, line 54: Call it typecheck please

sk-: done

TrySound (approved):

TrySound:
Thanks!

PR #1357, opened 2021-02-21
Test regressions with W3C SVG 1.1 Test Suite

TrySound:
Visual regressions tests will give us more confidence in SVGO stability.
Test cases are downloaded from w3.org, screenshooted with playwright and
compared with pixelmatch.

See test suite https://www.w3.org/Graphics/SVG/Test/20110816/

Bugs detected with these tests:
- https://github.com/svg/svgo/commit/9b97e06ef69b7961f342c9ee8468c552c9b503a3
- https://github.com/svg/svgo/pull/1371
- https://github.com/svg/svgo/commit/c1d5f0f7a93a6f1c947baa5860ddd40f8d1bb701
- https://github.com/svg/svgo/commit/9de471a0f4bea20581034c2d74b0652c27e0fad0
- https://github.com/svg/svgo/commit/091172a392f6d751855477d51e5c8bd3e44fde97
- https://github.com/svg/svgo/commit/36391564f2d0b2c7ad86c746bacdd6de030ebd3d
- https://github.com/svg/svgo/commit/d14315b68fc1fcb43212219a0da5857ad7a23cd0
- https://github.com/svg/svgo/commit/9d67586787ad16799d7a1cfce1eb0bce829470da
- https://github.com/svg/svgo/commit/d06747abca1e9e652a764db9747395d14f7f6f98
- https://github.com/svg/svgo/commit/dd37fcfebbe2f3cea687ad0bf890ff83ab3c9eb7
- https://github.com/svg/svgo/pull/1378

PR #1371, opened 2021-02-23
Fix parsing xml entities

TrySound:
Ref https://github.com/svg/sax/releases/tag/v0.1.1

This fixes enother case in [regression tests](https://github.com/svg/svgo/pull/1357).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -180,9 +180,9 @@
       }
     },
     "@trysound/sax": {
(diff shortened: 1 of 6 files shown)


yisibl:
Is the next plan to remove `<ENTITY>`?

TrySound:
@yisibl Why? Doctype is removed with plugins enabled.

Issue #1226, opened 2020-03-05
empty style tag not removed


wenerme:
I think this should be removed

```
    <defs>
        <style/>
    </defs>
```

from

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="200" height="200" class="icon" viewBox="0 0 1024 1024">
    <defs>
        <style/>
    </defs>
    <path fill="#555" d="M883.832 211.353l-309-176.02L510.593 0l-64.241 35.332-321.204 181.802L75.039 245.4v531.273l61.028 34.69 351.398 199.146L510.592 1024l64.24-35.332L885.76 812.005l61.029-34.69V246.043zM479.756 932.136L200.951 774.103a73.235 73.235 0 0011.563-37.902 75.162 75.162 0 00-73.877-75.162V317.35l279.448 157.39a93.15 93.15 0 0064.241 106.64zm-33.405-510.073L169.473 264.03 446.35 106.64a73.235 73.235 0 00128.482 0l278.805 158.675-278.805 156.748a93.15 93.15 0 00-128.482 0zM880.62 661.039a75.804 75.804 0 00-75.805 75.804 73.235 73.235 0 0012.206 38.545L542.07 932.136V582.023a93.15 93.15 0 0064.24-107.283l275.594-157.39v344.331z"/>
</svg>
```

SVGO version 1.3.2

Config

```json
{
  "plugins": [
    {
      "removeAttrs": {
        "attrs": "data-.*"
      }
    },
    {
      "sortAttrs": {}
    }
  ]
}
```

TrySound:
Fixed in https://github.com/svg/svgo/releases/tag/v2.3.0

PR #1373, opened 2021-02-24
Remove <style> element from the minifyStyles plugin when it is empty

yisibl:
Fixed: https://github.com/svg/svgo/issues/1226

In the minifyStyles plugin, if the <style> is empty or the css rule in it is not applied to the current svg node, we should remove the empty <style> element.

A similar process has been done in the [inlineStyles plugin](https://github.com/svg/svgo/blob/091172a392f6d751855477d51e5c8bd3e44fde97/plugins/inlineStyles.js#L220-L223).

https://github.com/svg/svgo/blob/091172a392f6d751855477d51e5c8bd3e44fde97/plugins/inlineStyles.js#L220-L223

## Origin

```svg
    <style></style>
    <style />
    <rect x="10" y="10" width="100" height="100"/>
</svg>
```

## Before: 

```svg
    <style />
    <style />
    <rect x="10" y="10" width="100" height="100"/>
</svg>
```

## After:

```svg
    <rect x="10" y="10" width="100" height="100"/>
</svg>
```

### DIFF
--- a/test/plugins/minifyStyles.13.svg
+++ b/test/plugins/minifyStyles.13.svg
@@ -1,23 +1,22 @@
 <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 48 48">
     <defs>
-        <style></style>
-        <linearGradient id="file-name_svg__file-name_svg__original-id" x1="12" y1="-1" x2="33" y2="46" gradientUnits="userSpaceOnUse">
+        <style>/* test */</style>
+        <linearGradient id="foo" x1="12" y1="-1" x2="33" y2="46" gradientUnits="userSpaceOnUse">
             <stop offset="0" stop-color="#6b5aed" stop-opacity="0" />
             <stop offset="1" stop-color="#6b5aed" />
         </linearGradient>
     </defs>
-    <path d="M5 24a21.9 21.9 10 1 1 8 18" fill="url(#file-name_svg__file-name_svg__original-id)"/>
+    <path d="M5 24a21.9 21.9 10 1 1 8 18" fill="url(#foo)"/>
 </svg>
 
 @@@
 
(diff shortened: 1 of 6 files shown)


test/plugins/cleanupIDs.19.svg
```diff
@@ -6,7 +6,7 @@
             <stop offset="1" stop-color="#6b5aed" />
         </linearGradient>
     </defs>
-    <path d="M46 24a21.9 21.9" fill="url(#file-name_svg__file-name_svg__original-id)"/>
+    <path d="M5 24a21.9 21.9 10 1 1 8 18" fill="url(#file-name_svg__file-name_svg__original-id)"/>
```
yisibl, line 9: `M46 24a21.9 21.9` is invalid path data, fixed it.

test/plugins/minifyStyles.13.svg
```diff
@@ -0,0 +1,22 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 48 48">
+    <defs>
+        <style>/* test */</style>
```
yisibl, line 3: At present, the test here fails, and it actually generates:
```svg
<style>
</style >
```
I don't know how to solve it yet.

strarsis:
My two cents here: Wouldn't it make sense to add a plugin that just removes empty `<style>` elements?
As multiple plugins should now implement the same logic, it should make sense to put this into a dedicated plugin.

TrySound:
@strarsis I think https://github.com/svg/svgo/pull/1381 perfectly fits into this task since it combines multiple `<style>` content it may collapse `<style>` into none.

yisibl:
@strarsis If there can be better done, I support it.

TrySound:
This is implemented as part of new mergeStyles plugin. Thanks anyway!

yisibl:
Continue to improve: https://github.com/svg/svgo/pull/1457

PR #1374, opened 2021-02-25
Use https when possible and remove an inactive domain

XhmikosR:
I had to remove `http://entropyinteractive.com/2011/04/gjk-algorithm/` since the domain is up for sale.

### DIFF
--- a/plugins/_path.js
+++ b/plugins/_path.js
@@ -75,7 +75,7 @@ exports.path2js = function(path) {
 
             data = data.map(Number);
             // Subsequent moveto pairs of coordinates are threated as implicit lineto commands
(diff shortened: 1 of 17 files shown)


plugins/_path.js
```diff
@@ -611,7 +611,7 @@ function set(dest, source) {
 /**
  * Checks if two paths have an intersection by checking convex hulls
  * collision using Gilbert-Johnson-Keerthi distance algorithm
- * http://entropyinteractive.com/2011/04/gjk-algorithm/
+ * https://web.archive.org/web/20180822200027/http://entropyinteractive.com/2011/04/gjk-algorithm/
```
XhmikosR, line 614: Switched this to web.archive.org not sure if we should drop it

PR #1376, opened 2021-02-25
Add plugin to normalize the letter case of colors

melusc:
I hope everything is fine this time.

### DIFF
--- a/plugins/normalizeColorCase.js
+++ b/plugins/normalizeColorCase.js
@@ -1,0 +1,92 @@
+'use strict';
+
+exports.type = 'perItem';
+
+exports.active = false;
+
+exports.description
+  = 'Makes colors consistent and more readable by making the letter case consistent';
+
+/* Each can be {lower|upper|false} */
+exports.params = {
+  hex: 'lower',
+  rgb: 'lower',
+  hsl: 'lower',
+  colorName: 'lower',
+};
+
+exports.fn = function (
+  item, params
+) {
+  if ( typeof params === 'object' && item.isElem() ) {
+    item.eachAttr( attrObj => {
+      normalizeColor(
+        attrObj,
+        params
+      );
+    } );
+  }
+};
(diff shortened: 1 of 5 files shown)


TrySound:
Thank you! I will look in a few days.

plugins/plugins.js
```diff
@@ -49,3 +49,4 @@ exports.removeXMLProcInst = require('./removeXMLProcInst.js');
 exports.reusePaths = require('./reusePaths.js');
 exports.sortAttrs = require('./sortAttrs.js');
 exports.sortDefsChildren = require('./sortDefsChildren.js');
+exports.normalizeColorCase = require('./normalizeColorCase.js');
```
XhmikosR, line 52: please add the missing EOF newline gere :)

plugins/normalizeColorCase.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+
+exports.type = 'perItem';
+
+exports.active = false;
```
XhmikosR, line 5: @TrySound this plugin should be enabled by default in the future IMHO since it should improve gzip compression

sk-:
I feel this plugin is not doing too much, specially given that this [default plugin](https://github.com/svg/svgo/blob/master/plugins/convertColors.js) will already convert `rgb(r, g, b)` to `#rrggbb`, and it also converts from hex to named colors.

Maybe that plugin should be extended as to always output a cased color, and also support `hsl` colors.

@TrySound it'd be good to define certain guidelines for including plugins in the core. For example obtain savings of X% in a benchmark dataset.

XhmikosR:
IMHO normalizing the case of colors is valuable because it helps with gzip compression. The same should be investigated for the sortAttrs plugin at some point, but I'm not sure how this can be implemented. Now, if this should be a separate plugin or not, that's another discussion. :)

PR #1278, opened 2020-09-02
Add mergeStyles plugin

strarsis:
This PR adds the mergeStyles plugin that merges `<style>` elements into one, 
preserving media queries by the `<style>` `media` attribute.

You can already try this feature branch by installing it using this npm command:
`npm install github:strarsis/svgo#mergeStyles-plugin`
The plugin would also be enabled by default in this PR as it can help with compression by subsequent plugins.

PR #1381, opened 2021-02-27
Add mergeStyles plugin

strarsis:
This PR adds the mergeStyles plugin that merges `<style>` elements into one, 
preserving media queries by the `<style>` `media` attribute.

You can already try this feature branch by installing it using this npm command:
`npm install github:strarsis/svgo#mergeStyles-plugin`
The plugin would also be enabled by default in this PR as it can help with compression by subsequent plugins.

Outdated, closed PR: https://github.com/svg/svgo/pull/1278


strarsis:
Hm, strange, the CI test fails (https://github.com/svg/svgo/pull/1381/checks?check_run_id=2087136821#step:6:469).
I can't reproduce this locally with the same node version.

TrySound:
I migrated ast to xast https://github.com/syntax-tree/xast

strarsis:
@TrySound: Oh, this would explain the issue. The docs hadn't been updated yet, right?
https://github.com/svg/svgo/blob/master/docs/how-it-works/en.md#2-svg2js

TrySound:
I never updated how-it-works 😅

strarsis:
OK, fixed API usage and lint error.

strarsis:
@TrySound: Ready to merge!

plugins/mergeStyles.js
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Merge multiple style elements into one.
+ *
+ * @param {Object} document document element
+ *
+ * @author strarsis <strarsis@gmail.com>
+ */
+exports.fn = function (document) {
+  // collect <style/>s (preserve order)
+  var styleEls = document.querySelectorAll('style');
```
TrySound, line 18: Please avoid using JSAPI where possible. Import similar utility from lib/xast.js

plugins/mergeStyles.js
```diff
@@ -0,0 +1,70 @@
...
+      // skip empty <style/>s or <foreignObject> content.
+      continue;
+    }
+
+    var cssStr = cssTools.getCssStr(styleEl);
+
+    styles.push({
+      styleEl: styleEl,
+
+      mq: styleEl.attr('media'),
```
TrySound, line 37: All attributes manipulation should be direct element.attributes object access

plugins/mergeStyles.js
```diff
@@ -0,0 +1,70 @@
...
+        '@media ' + style.mq.value + ' {' + '\n' + style.cssStr + '\n' + '}';
+      collectedStyles.push(wrappedStyles);
+    } else {
+      collectedStyles.push(style.cssStr);
+    }
+
+    if (styleNo > 0) {
+      // remove all processed style elements – except the first one
+      var styleParentEl = style.styleEl.parentNode;
+      styleParentEl.spliceContent(
```
TrySound, line 56: Manipular element.children directly

plugins/mergeStyles.js
```diff
@@ -0,0 +1,70 @@
...
+      styleEl: styleEl,
+
+      mq: styleEl.attr('media'),
+      cssStr: cssStr,
+    });
+  }
+
+  var collectedStyles = [];
+  var styleNo = 0;
+  for (var style of styles) {
```
TrySound, line 44: If you need current index use classic for loop instead

plugins/mergeStyles.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  var styles = [];
+  for (var styleEl of styleEls) {
+    if (styleEl.isEmpty() || styleEl.closestElem('foreignObject')) {
+      // skip empty <style/>s or <foreignObject> content.
+      continue;
+    }
+
+    var cssStr = cssTools.getCssStr(styleEl);
```
TrySound, line 32: Use const where possible or let if you need mutation

plugins/mergeStyles.js
strarsis: Ah, the reason for this was that the linter complained about comparison of a string with a number in the `if (styleNo > 0) {` line, because the counter variable in `for ... of` loops are strings. I could use `parseInt(...)` in the if condition or disable the linter for that line.

TrySound: I mean
```js
for (let i = 0; i < styles.length; i += 1) {
  const style = styles[i]
}
```

strarsis: Yes, this is much better.

plugins/mergeStyles.js
strarsis: Done.

plugins/mergeStyles.js
strarsis: OK

plugins/mergeStyles.js
strarsis: I tried to use the JavaScript array directly, including its methods. But the node references don't always match, `indexOf` returns `-1` in one of the tests. And the reference previously stored and compared in array doesn't match.

plugins/mergeStyles.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
+
+exports.type = 'full';
+exports.active = true;
+exports.description = 'merge multiple style elements into one';
+
+const cssTools = require('../lib/css-tools'),
+  xast = require('../lib/xast');
```
TrySound, line 8: Add own const for each variable

TrySound: Also import each function separately like named imports

plugins/mergeStyles.js
```diff
@@ -0,0 +1,71 @@
...
+  var styleEls = xast.querySelectorAll(document, 'style');
+
+  // no <styles/>s, nothing to do
+  if (styleEls === null || styleEls.length <= 1) {
+    return document;
+  }
+
+  var styles = [];
+  for (var styleEl of styleEls) {
+    if (styleEl.isEmpty() || styleEl.closestElem('foreignObject')) {
```
TrySound, line 28: There is closestByName

test/plugins/mergeStyles.04.svg
```diff
@@ -0,0 +1,23 @@
+<svg id="test" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
```
TrySound, line 1: Add description for each test case

```
description

===

<svg...
```

plugins/mergeStyles.js
```diff
@@ -0,0 +1,71 @@
...
+      styleParentEl.children = styleParentEl.children.splice(
+        styleParentEl.children.indexOf(style.styleEl),
+        1
+      );
+    }
+  }
+
+  // assign the collected styles to the first style element
+  styles[0].styleEl.removeAttr('media'); // remove media mq attribute as CSS media queries are used
+  var collectedStylesStr = collectedStyles.join('\n\n');
```
TrySound, line 67: Replace all vars with consts and lets please

plugins/mergeStyles.js
```diff
@@ -0,0 +1,71 @@
...
+    if (style.mq) {
+      let wrappedStyles =
+        '@media ' + style.mq + ' {' + '\n' + style.cssStr + '\n' + '}';
+      collectedStyles.push(wrappedStyles);
+    } else {
+      collectedStyles.push(style.cssStr);
+    }
+
+    if (styleNo > 0) {
+      // remove all processed style elements – except the first one
```
TrySound, line 56: Move comment before condition

plugins/mergeStyles.js
strarsis: OK.

plugins/mergeStyles.js
strarsis: OK

plugins/mergeStyles.js
strarsis: OK

plugins/mergeStyles.js
strarsis: Yes, this is better as it describes why the condition is there.

plugins/mergeStyles.js
strarsis: OK, should be const where it isn't mutated.

package-lock.json
```diff
@@ -1,8 +1,3527 @@
 {
   "name": "svgo",
   "version": "2.2.2",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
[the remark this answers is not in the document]
strarsis: Done

test/plugins/mergeStyles.04.svg
strarsis: I added descriptions with style similar to git commit messages, like 
`Check whether plugin works with only one style element (no further merging needed, noop).`.

plugins/mergeStyles.js
```diff
@@ -6,6 +6,10 @@ exports.description = 'merge multiple style elements into one';
 
 const cssTools = require('../lib/css-tools');
 const xast = require('../lib/xast');
+const querySelectorAll = xast.querySelectorAll;
+const closestByName = xast.closestByName;
+const getCssStr = cssTools.getCssStr;
+const setCssStr = cssTools.setCssStr;
```
TrySound, line 12: I mean this
```
const { querySelectorAll, closestByName } = require('../lib/xast.js');
```

TrySound: And put all requires before exports please

plugins/mergeStyles.js
```diff
@@ -0,0 +1,75 @@
...
+  const styleEls = querySelectorAll(document, 'style');
+
+  // no <styles/>s, nothing to do
+  if (styleEls === null || styleEls.length <= 1) {
+    return document;
+  }
+
+  let styles = [];
+  for (let styleEl of styleEls) {
+    if (styleEl.isEmpty() || closestByName(styleEl, 'foreignObject')) {
```
TrySound, line 32: Check children.length instead of isEmpty

plugins/mergeStyles.js
```diff
@@ -0,0 +1,75 @@
...
+ * @param {Object} document document element
+ *
+ * @author strarsis <strarsis@gmail.com>
+ */
+exports.fn = function (document) {
+  // collect <style/>s (preserve order)
+  const styleEls = querySelectorAll(document, 'style');
+
+  // no <styles/>s, nothing to do
+  if (styleEls === null || styleEls.length <= 1) {
```
TrySound, line 26: querySelectorAll always return array

TrySound:
Could you add removing empty <style> support?

test/plugins/mergeStyles.02.svg
```diff
@@ -0,0 +1,22 @@
...
+    <rect width="100" height="100" class="st0" style="stroke-width:3;margin-top:1em;margin-right:1em;margin-bottom:1em;margin-left:1em"/>
+</svg>
+
+@@@
+
+<svg id="test" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
+    <style>
+        .st0{ fill:red; padding-top: 1em; padding-right: 1em; padding-bottom: 1em; padding-left: 1em; }
+
+@media screen and (max-width: 200px) { .st0 { display: none; } }
```
yisibl, line 19: Formatting is not currently considered?

strarsis: Compact style output is used. The newlines are added tor better understanding the different merged parts.
The minifyStyles plugin in a step afterwards should then minify the styles as much as possible.

strarsis:
Yes, the mergeStyles plugin would remove the left over style element when the styles to be added are empty.

plugins/mergeStyles.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
+
+const { querySelectorAll, closestByName } = require('../lib/xast.js');
+const { getCssStr,        setCssStr     } = require('../lib/css-tools');
```
TrySound, line 4: You can install prettier in your ide.

strarsis:
@TrySound, @yisibl: Sorry for the delay. I will push some updates soon.

plugins/mergeStyles.js
```diff
@@ -0,0 +1,71 @@
...
+      const styleParentEl = style.styleEl.parentNode;
+      styleParentEl.children.splice(
+        styleParentEl.children.indexOf(style.styleEl),
+        1
+      );
+    }
+  }
+
+  // assign the collected styles to the first style element
+  styles[0].styleEl.removeAttr('media'); // remove media mq attribute as CSS media queries are used
```
TrySound, line 66: Use delete operator

strarsis: OK: `delete firstStyle.styleEl.attributes.media;`

plugins/mergeStyles.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  return document;
+};
+
+
+function removeFromParent(el) {
+  const parentEl = el.parentNode;
+  return parentEl.children.splice(
+	parentEl.children.indexOf(el),
```
TrySound, line 78: Use full names please. element, parentNode etc.

test/plugins/mergeStyles.02.svg
strarsis: Alright, no newlines are added when the collected styles are merged. This reduces the file size a tiny bit. Formatting is not considered by this particular plugin because it only merges styles. But the minifyStyles plugin that uses CSSO could also be used for beautifying styles. But this wouldn't probably be considered an optimization.

plugins/mergeStyles.js
strarsis: Alright, I renamed the symbols so that they use full names (where possible).

TrySound:
Also check `<style>` for invalid "type" attribute. See https://github.com/svg/svgo/blob/master/lib/style.test.js#L193-L206

plugins/mergeStyles.js
```diff
@@ -15,8 +15,11 @@ exports.description = 'merge multiple style elements into one';
  * @author strarsis <strarsis@gmail.com>
  */
 exports.fn = function (document) {
-  // collect <style/>s (preserve order)
-  const styleElements = querySelectorAll(document, 'style');
+  // collect <style/>s with valid type attribute (preserve order)
+  const styleElements = querySelectorAll(
+    document,
+    'style:not([type]), style[type=""], style[type="text/css"]'
```
TrySound, line 21: Let's avoid complex selectors and check attributes manually.

strarsis:
Alright, the plugin now only processes valid `<style>` elements.

plugins/mergeStyles.js
```diff
@@ -15,8 +15,11 @@ exports.description = 'merge multiple style elements into one';
...
+  );
 
   // no <styles/>s, nothing to do
   if (styleElements.length <= 1) {
```
TrySound, line 25: I think this prevents single empty `<style>` from removing

plugins/mergeStyles.js
strarsis: Right, the type attribute is now checked directly.

plugins/mergeStyles.js
strarsis: OK, it works now. I first thought, `splice` doesn't mutate but returns the changed array, but it actually returns the removed elements.

plugins/mergeStyles.js
```diff
@@ -28,7 +25,8 @@ exports.fn = function (document) {
 
   let styles = [];
   for (let styleElement of styleElements) {
-    if (closestByName(styleElement, 'foreignObject')) {
+    if ((styleElement.attributes.type && styleElement.attributes.type !== 'text/css') ||
```
TrySound, line 28: Split into two if statements please

plugins/mergeStyles.js
strarsis: So these extra spaces for padding should be rather avoided?

plugins/mergeStyles.js
strarsis: OK

plugins/mergeStyles.js
strarsis: Replaced with if statement.

strarsis:
Hm, I still have to find out why one of the empty `<style>` element is not removed in two of the tests.

strarsis:
There is one scenario where more logic is needed though: `<link>` elements are rarely used in SVGs but they are possible.
The style priority of `<link>` elements and `<style>` is the same - and only influenced by their order in the SVG document.
To take this into account, the plugin would have to cluster the `<style>` styles between the `<link>` styles and merge only the clusters between the `<link>`.

package-lock.json
```diff
@@ -5,12 +5,12 @@
   "requires": true,
   "dependencies": {
```
TrySound, line 6: Please revert again

plugins/plugins.js
```diff
@@ -14,6 +14,7 @@ exports.convertPathData = require('./convertPathData.js');
 exports.convertShapeToPath = require('./convertShapeToPath.js');
 exports.convertStyleToAttrs = require('./convertStyleToAttrs.js');
 exports.convertTransform = require('./convertTransform.js');
+exports.mergeStyles = require('./mergeStyles.js');
```
TrySound, line 17: Add this plugin to config.js list

strarsis: Reverted to package-lock of master.

package-lock.json
strarsis: Reverted to upstream package lock.

TrySound:
I meant revert whole package-lock.json. We are not on npm 7 yet.

TrySound:
Still here.

plugins/mergeStyles.js
TrySound: Still single empty style is not removed because of this condition.

plugins/mergeStyles.js
```diff
@@ -0,0 +1,88 @@
...
+exports.fn = function (document) {
+  // collect <style/>s with valid type attribute (preserve order)
+  const styleElements = querySelectorAll(document, 'style');
+
+  // no <styles/>s, nothing to do
+  if (styleElements.length <= 1) {
+    return document;
+  }
+
+  let styles = [];
```
TrySound, line 26: const

plugins/mergeStyles.js
```diff
@@ -0,0 +1,88 @@
...
+
+    styles.push({
+      styleElement: styleElement,
+
+      mq: styleElement.attributes.media,
+      cssStr: cssString,
+    });
+  }
+
+  let collectedStyles = [];
```
TrySound, line 51: const

plugins/mergeStyles.js
```diff
@@ -0,0 +1,88 @@
...
+  // collect <style/>s with valid type attribute (preserve order)
+  const styleElements = querySelectorAll(document, 'style');
+
+  // no <styles/>s, nothing to do
+  if (styleElements.length <= 1) {
+    return document;
+  }
+
+  let styles = [];
+  for (let styleElement of styleElements) {
```
TrySound, line 27: const

plugins/mergeStyles.js
```diff
@@ -0,0 +1,88 @@
...
+    });
+  }
+
+  let collectedStyles = [];
+  for (let styleNo = 0; styleNo < styles.length; styleNo += 1) {
+    const style = styles[styleNo];
+
+    if (style.mq) {
+      const wrappedStyles =
+        '@media ' + style.mq + ' {' + '\n' + style.cssStr + '\n' + '}';
```
TrySound, line 57: use template literal

lib/svgo/config.js
```diff
@@ -34,6 +34,7 @@ const pluginsOrder = [
   'collapseGroups',
   'convertPathData',
   'convertTransform',
+  'mergeStyles',
```
TrySound, line 37: I think it should be before minifyStyles an inlineStyles

lib/css-tools.js
```diff
@@ -189,7 +189,10 @@ function csstreeToStyleDeclaration(declaration) {
  * @return {string} CSS string or empty array if no styles are set
  */
 function getCssStr(elem) {
-  if (elem.children[0].type === 'text' || elem.children[0].type === 'cdata') {
+  if (
+    elem.children[0] &&
```
TrySound, line 193: Check children for length instead

plugins/mergeStyles.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    // remove all processed style elements – except the first one
+    if (styleNo > 0) {
+      removeFromParent(style.styleElement);
+    }
+  }
+  const collectedStylesString = collectedStyles.join('');
+
+  // re-assign the collected styles to the first style element
```
TrySound, line 70: "combine collected styles in the first style element"

plugins/mergeStyles.js
```diff
@@ -0,0 +1,89 @@
...
+      cssStr: cssString,
+    });
+  }
+
+  const collectedStyles = [];
+  for (let styleNo = 0; styleNo < styles.length; styleNo += 1) {
+    const style = styles[styleNo];
+
+    if (style.mq) {
+      const wrappedStyles = `@media ${style.mq} {
```
TrySound, line 56: `\n` instead of newlines please

plugins/mergeStyles.js
```diff
@@ -0,0 +1,89 @@
...
+ * @param {Object} document document element
+ *
+ * @author strarsis <strarsis@gmail.com>
+ */
+exports.fn = function (document) {
+  // collect <style/>s with valid type attribute (preserve order)
+  const styleElements = querySelectorAll(document, 'style');
+
+  // no <styles/>s, nothing to do
+  if (styleElements.length <= 1) {
```
TrySound, line 22: Previous question probably hidden by github. This still prevents one empty style from removing.

package-lock.json
```diff
@@ -5,12 +5,12 @@
...
     "@babel/code-frame": {
-      "version": "7.12.13",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.13.tgz",
-      "integrity": "sha512-HV1Cm0Q3ZrpCR93tkWOYiuYIgLxZXZFVG2VgK+MBWjUqZTundupbfx2aXarXuw5Ko5aMcjtJgbSs4vUGBS5v6g==",
+      "version": "7.12.11",
```
TrySound, line 8: Why I still see package-lock changes here?

plugins/mergeStyles.js
strarsis: Replaced with one line. This has the added benefit of saving some space characters.

plugins/mergeStyles.js
strarsis: Done.

lib/css-tools.js
strarsis: Done.

plugins/mergeStyles.js
strarsis: Indeed. I added a test and a fix for this.

lib/svgo/config.js
strarsis: Done. Moved it above the two.

package-lock.json
strarsis: 🤔 From where should I revert the package-lock? Should I use the current one from svgo or from a previous commit of this branch? `npm install` seems to update the package lock on its own.

plugins/mergeStyles.js
strarsis: Done (continues in https://github.com/svg/svgo/pull/1381#discussion_r602719351).

package-lock.json
TrySound: git checkout HEAD~48 -- package-lock.json

TrySound:
Thank you!

test/plugins/mergeStyles.10.svg
```diff
@@ -0,0 +1,15 @@
+Check whether plugin removes one empty <style> element that is also the only <style> element.
+
+===
+
+<svg id="test" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
+    <style>
+	  </style>
```
yisibl, line 7: A great job. There are a few tips:

The test case should include:
```html
  <style>/* test */</style>
  <style />
```

TrySound: @yisibl do you want to send PR?

yisibl:
In the next PR, do you plan to remove unused CSS?

TrySound:
Maybe eventually

test/plugins/mergeStyles.10.svg
yisibl: OK.

yisibl: @TrySound Do we now have an API to recognize CSS comments?

strarsis: The minify styles plugin can handle this before and after this plugin.

TrySound: What if we merge styles after minifying? I asked to move before first because there was unwanted newlines though @strarsis removed them and maybe it makes sense to merge after now.

strarsis: Yes, merging after minification is nice. The minifier removed all comments and only left the bare minimum.

PR #1382, opened 2021-02-27
refactor: replace uses of for in

sk-:
Replace the uses of `for in` with `for of` and `Object.entries`/`Object.values`/`Object.keys`.

One case was further simplified by using a spread.


XhmikosR:
Did you benchmark this? `for...in` could be faster although probably not by a lot, plus there were no checks for prototype own properties before. Might be worth benchmarking this with complex files.

I also plan to take a stab at tweaking things after every file is formatted. :)

TrySound:
Thanks!

PR #1397, opened 2021-03-02
Update CI config

XhmikosR:
* separate lint step to a new job and run it only once
* add caching


XhmikosR:
Do note that I had to use the same cache paths even for the lint job, otherwise `npm ci` would be slow in lint due to playwright downloading all browsers.

Issue #777, opened 2017-08-16
svgo improvements


caub:
I wanted to make svgo more flexible, using jsdom, but that change a lot plugins usage, so I started https://github.com/caub/svgz, I'd like to port a bunch of svgo plugins there.

So far it's going good, I still have some big plugins to convert like the ones with path data, but the rest is often easier to do

I'd be glad to set svgo main contributors owners there, and have contributors, if people like this direction

GreLI:
🤔

caub:
maybe @deepsweet would see the idea interesting? not trying to steal anything, just trying to make things easier to write plugins (one example: some plugins need to know if there are any `<script>` or `<style>`, it becomes [easier](https://github.com/caub/svgz/blob/master/plugins/removeUselessIds.js#L7) to do, and more correct (because I think it's often detected during the tree walk, but it should be done beforehand). I'm not sure it'll be faster, that's what I was trying to see

deepsweet:
My intent (happened few times actually) to rewrite SVGO from a scratch was based exactly on JSDOM. I see many benefits:

* DOM Tree for free
* DOM API for free
* It should be possible to abstract SVGO core from Nodejs at all so we can have a build for browser (I remember there were 1+ requests for this)

I had some very raw prototype but now it's just lost in time.

Another ideas which I'd like to see if someone would like to take over this huge effort:

* ESNext all the things with Babel, Nodejs 4+
* modularize
  * monorepo
  * each plugin `svgo-plugin-foo` is a package
  * decouple `svgo-cli`, "API first"
  * presets
    * `svgo-preset-safe`
    * `svgo-preset-aggressive`
* [Cheerio](https://github.com/cheeriojs/cheerio) on top of JSDOM if it's really necessary for even more handy DOM tree traverse/mutation/etc
* Strict semver with no fear of positive integers and much faster release cycle
* …

Regarding "svgz" in particular – I don't personally think that it's a good idea to have it as a separate project and split the effort. Many SVG optimizations are based on pretty complex math and geometry, I know zero to few JS devs who are good at it and able to fix such an issues. @GreLI did a really amazing job on maintaining SVGO and he knows about this a lot, more than me actually :)

My point is to 1) discuss more tech details 2) create a branch 3) collaborate. It's REALLY important to have all the devs who are deep in SVG context together.

I'm no longer in the context of SVG optimization and not going to do any significant amount of work by myself for SVGO for sure, but I'll be happy to help with any tech questions and discussions.

[...]

PR #1399, opened 2021-03-02
Implement style computing

TrySound:
Ref https://github.com/svg/svgo/issues/777

Currently a lot of optimisations are attributes specific and may be
broken because of inline or shared styles.

In this diff I'm trying to solve the problem with getComputedStyle
analog.

`computeStyle` collects attributes, shared css rules, inline styles
and inherited styles and checks whether they can be statically optimised
or left as deoptimisation.

As an example I integrated computeStyle into `removeHiddenElems` plugin
which is fixed bugs with styles in style rules and inline styles.

### DIFF
--- a/lib/style.js
+++ b/lib/style.js
@@ -75,16 +75,13 @@ const parseStylesheet = (css) => {
 
 const computeOwnStyle = (node, stylesheet) => {
   const computedStyle = {};
+  const importantStyles = new Map();
 
   // collect attributes
-  for (const attr of Object.values(node.attrs)) {
-    if (attrsGroups.presentation.includes(attr.name)) {
-      computedStyle[attr.name] = {
-        type: 'static',
-        inherited: false,
-        value: attr.value,
-        important: false,
-      };
+  for (const { name, value } of Object.values(node.attrs)) {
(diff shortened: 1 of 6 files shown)


TrySound:
cc @strarsis @XhmikosR @caub @sk-

strarsis:
The inlineStyles plugin [already does this](https://github.com/svg/svgo/blob/master/plugins/inlineStyles.js#L101), it collects the styles and then stably sorts them which surprisingly is sufficient. It should be possible to extract this part into a dedicated function (css tools module). As a performance optimization it would be also possible to provide a prepared presorted structure as the plugin does so this doesn't has to be performed each time. Atlernatively transparent caching could be used (as it was added for element styles support).

TrySound:
Proper caching will require some architecture changes. First we should fix bugs.

And yeah, I saw inlineStyles though the point here is resolving style from all sources including inheritance and presentation attributes.

All plugins should work correctly in isolation. So caching will be implemented on core level not in plugin.

caub:
looks really good 👍🏽 from the tests, never thought getComputedStyle could be done in this quite short way

TrySound:
Me neither 😄

Issue #1344, opened 2021-02-19
v2 plugins not working as intended


vincentbriglia:
with config 
```
const { extendDefaultPlugins } = require('svgo');
module.exports = {
  plugins: extendDefaultPlugins([
    { name: 'removeXMLNS' },
    { name: 'sortAttrs' },
    { name: 'removeDimensions' },
    { name: 'removeXMLNS' },
    { name: 'removeViewBox', active: false },
    { name: 'removeAttrs', params: { attrs: '(fill|fill-rule|clip-rule)' } },
    { name: 'removeUselessStrokeAndFill' },
    { name: 'addAttributesToSVGElement', params: { attributes: { fill: 'currentColor' } } },
  ]),
};

```

fills are not removed, dimensions are not removed.

vincentbriglia:
it did work in 1.x, i have merely changed the config to adapt to v2

TrySound:
Hi. Could you please provide svg you working with?

TrySound:
Also please check 2.0.2

vincentbriglia:
```
<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
<path d="M13.934 21.0214C13.5403 20.9016 13.1123 21.1241 13.0096 21.5349C12.8727 21.98 12.5474 22.3053 12.1023 22.4422C11.76 22.5449 11.4005 22.5107 11.0923 22.3395C10.7842 22.1683 10.5446 21.8944 10.459 21.5521C10.3391 21.1583 9.92829 20.9358 9.53457 21.0385C9.14084 21.1583 8.9183 21.5692 9.02101 21.9629C9.22643 22.6819 9.72287 23.281 10.3734 23.6576C10.8013 23.8802 11.2635 24 11.7257 24C11.9996 24 12.2564 23.9658 12.5303 23.8802C13.4547 23.6063 14.1737 22.8873 14.4475 21.9629C14.5674 21.5521 14.3277 21.1412 13.934 21.0214Z" fill="#2F2F2E"/>
[...]

PR #1401, opened 2021-03-03
Update example for removeAttrs

Scionar:
Old example didn't work with `removeAttrs` plugin. This worked for me and also validated here:
https://github.com/svg/svgo/issues/1344


TrySound:
Thanks!

PR #1405, opened 2021-03-04
chore: Improve TS config

sk-:
Changed the TS config file to include all relevant folders and explicitly exclude those files that fail the type checks. In this way new files will be automatically type checked and we can see the list of files that need types improvements.

In this pass I also fixed some js files.


lib/path.js
```diff
@@ -25,10 +25,16 @@ const argsCountPerCommand = {
   a: 7,
 };
 
+/**
+ * @param {string} c
+ */
 const isCommand = (c) => {
```
TrySound, line 31: TS cannot infer all this, right?

sk-: It cannot infer the argument types for standalone functions, but it can infer the return types.

However if you pass a closure to a typed function, then the arguments are inferred. For example if `data` is of type `string[]`, then `data.map(s => s.toUpperCase())`, it will infer that `s` is of type `string`.

lib/path.js
```diff
@@ -39,16 +45,29 @@ const isWsp = (c) => {
   );
 };
 
+/**
+ * @param {string} c
+ */
 const isDigit = (c) => {
   const codePoint = c.codePointAt(0);
+  if (codePoint === undefined) {
```
TrySound, line 53: I'd prefer `codePoint == null`

lib/path.js
```diff
@@ -39,16 +45,29 @@ const isWsp = (c) => {
...
+    return false;
+  }
   return 48 <= codePoint && codePoint <= 57;
 };
 
+/**
+ * @typedef {'none' | 'sign' | 'whole' | 'decimal_point' | 'decimal' | 'e' | 'exponent_sign' | 'exponent'} ReadNumberState */
```
TrySound, line 60: newline before closing comment

lib/path.js
```diff
@@ -109,15 +128,17 @@ const readNumber = (string, cursor) => {
...
+/**
+ * @param {string} string
+ */
 const parsePathData = (string) => {
   const pathData = [];
-  let i = 0;
   let command = null;
-  let args;
-  let argsCount;
+  let args = /** @type {any[]} */ ([]);
```
TrySound, line 137: args are numbers

lib/path.js
sk-: Ok, done. But IMO it makes the code much harder to reason about when one relies on auto-conversion and `==` checks. For a developer is quite baffling why if the method can return `undefined` you are checking for null.

lib/path.js
sk-: done

lib/path.js
sk-: Thanks, fixed.

lib/path.js
TrySound: `value == null` checks for nullable. This makes less painful dilemma whether use null or undefined for not existing value and works as interop for existing api.

lib/path.js
```diff
@@ -260,6 +304,23 @@ const stringifyArgs = ({
   return result;
 };
 
+/**
+ *
+ * @typedef {{
+ *   command: string;
+ *   args: any[];
```
TrySound, line 311: Fix args everywhere please

sk-: Sorry I had forgotten one. Done.

lib/path.js
sk-: Well in this case the API was already written and is from the standard library. This has nothing to do with interoperability for the existing API.

lib/path.js
```diff
@@ -39,16 +45,30 @@ const isWsp = (c) => {
...
+ * @param {string} string
+ * @param {number} cursor
+ * @return {[number, number | null]}
+ */
 const readNumber = (string, cursor) => {
   let i = cursor;
   let value = '';
-  // none | sign | whole | decimal_point | decimal | e | exponent_sign | exponent
-  let state = 'none';
+  let state = /** @type {ReadNumberState} */ ('none');
```
TrySound, line 71: Did prettier wrap 'none' with parentheses?

lib/svgo/css-select-adapter.js
```diff
@@ -1,5 +1,9 @@
 'use strict';
 
+/**
+ * @param {any} node
+ * @return {node is any}
+ */
```
TrySound, line 6: No other types requires?

sk-: The other types seemed to be compatible. Of course there are some annotations missing, but because all of them are treated as `any` then it works, as it confirms to the adapter interface.

Here it was needed because the `isTag` needs to return a type assertion.

lib/path.js
sk-: Casts in JSDoc require the parentheses.

package.json
```diff
@@ -107,6 +107,7 @@
     "@rollup/plugin-commonjs": "^17.1.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^11.2.0",
+    "@types/mocha": "^8.2.1",
```
TrySound, line 110: Let's exclude tests for now and cover only library code

sk-: Actually checking the tests helped me come up with the right typings for the path module. Without the test and based just on the code, the argument `disableSpaceAfterFlags` would have been required, but in reality it's optional.

Or do you mean to only exclude "legacy" tests in the `test` folder?

TrySound: Ok then, keep it

Issue #1392, opened 2021-03-01
Add `--exclude` option to the CLI script


Tokimon:
**Is your feature request related to a problem? Please describe.**
I run the SVGO recursively on a single folder (`-r -f`), but I have some auto generated SVG files that I for structural purposes have placed in a nested folder in the given folder. However since I know this file is already optimized there is no reason to use resources on parsing it (as well the risk of accidental clearing of the file as it is only symbols), so I would like to exclude them when executing.

I have look in the code (and documentation) and I have not been able to find any way to do this.

**Describe the solution you'd like**
An option to exclude certain files (`--exclude` or `-e`), possibly as regular or glob expression (eg `--exclude "auto/.+\.svg"`)

**Describe alternatives you've considered**
I have considered building my own script that uses the SVGO module that parses the files I need. But that seems redundant, since all other options would be the same and thus a simple exclude in the CLI would be simpler.

**Additional context**
N/A

TrySound:
Fixed in https://github.com/svg/svgo/releases/tag/v2.3.0

PR #1409, opened 2021-03-08
Implement exclude pattern cli option

chambo-e:
Hello, 

I ran accross #1392 because we have almost exactly the same issue
We currently have a kinda big custom script allowing us to do filtering on recursive svgo processing and we were trying to migrate to the cli but this feature is missing

So I implementend a tiny option which accept an array of regex pattern that we can use to filter out file descriptors

### DIFF
--- a/lib/svgo/coa.js
+++ b/lib/svgo/coa.js
@@ -56,11 +56,11 @@ module.exports = function makeProgram(program) {
     .option('--indent <INTEGER>', 'Indent number when pretty printing SVGs')
     .option(
       '-r, --recursive',
-      "Use with '-f'. Optimizes *.svg files in folders recursively."
+      "Use with '--folder'. Optimizes *.svg files in folders recursively."
     )
     .option(
-      '-e, --exclude <PATTERN...>',
-      "Use with '-f'. Exclude files matching regular expression pattern."
+      '--exclude <PATTERN...>',
+      "Use with '--folder'. Exclude files matching regular expression pattern."
     )
     .option(
       '-q, --quiet',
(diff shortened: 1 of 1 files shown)


lib/svgo/coa.js
```diff
@@ -156,6 +160,9 @@ async function action(args, opts, command) {
     config.recursive = opts.recursive;
   }
 
+  // --exclude
+  config.exclude = opts.exclude ? opts.exclude.map(pattern => RegExp(pattern)) : [];
```
Tokimon, line 164: Don't you need to `new` the `RegExp`? (`new RegExp(pattern)`)

Tokimon: and maybe break the line up to make it a bit more readable?

```suggestion
  config.exclude = opts.exclude
    ? opts.exclude.map(pattern => RegExp(pattern))
    : [];
```

lib/svgo/coa.js
```diff
@@ -291,7 +299,7 @@ function processDirectory(config, dir, files, output) {
  */
 function getFilesDescriptions(config, dir, files, output) {
   const filesInThisFolder = files
-    .filter((name) => regSVGFile.test(name))
+    .filter((name) => regSVGFile.test(name) && !config.exclude.some(regExclude => regExclude.test(name)))
```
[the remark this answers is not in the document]
chambo-e: Looks better, thanks :)

Tokimon (approved):

chambo-e:
cc @TrySound @deepsweet if you have 5 min to take a look :)

TrySound:
Thanks!

PR #1417, opened 2021-03-10
Remove whitespaces and comments from the browser bundle

### LINKED ISSUE #1390: Make the browser bundle size smaller
I noticed that SVGO now provides an official browser bundle, but it costs 825 KB (huge!).
Size matters for everything that intends to work in the browser.
IMO if we want to see the browser version using in the production, we must spend some time to reduce the browser bundle.

I'm a micro-libraries developer and I have good expertise in bundle size optimizations.
If you agree that investing in the bundle size reducing worth it, I would be glad to create a few PRs reducing the `svgo.browser.js` size. Decided to ask you first before bringing the PRs you might not need. Maybe somebody also wants to help to optimize the browser version, so declaring the problem might involve them as well.

omgovich:
Closes #1390

**Results:** 826 KB → 551 KB _(–275 KB)_

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,15 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.12.13",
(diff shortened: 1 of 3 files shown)


TrySound:
Thanks!

PR #1430, opened 2021-03-16
README.md: better plugins' sorting

XhmikosR:
Maybe it's my OCD, but I feel like sorting alphabetically is more natural. In the plugins table, I kept the enabled ones at the top.

Let me know what you think :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <div align="center">
-  <img src="./logo/logo-web.svg" width="348.61" height="100" alt="SVGO logo"/>
+  <img src="./logo/logo-web.svg" width="348.61" height="100" alt="SVGO logo">
 </div>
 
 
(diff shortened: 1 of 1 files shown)


TrySound:
Plugins are specified in order they are run which is critical for optimisation and correctness.

XhmikosR:
Yeah, but in the README, it's harder to find which plugin is enabled if they are sorted randomly...

TrySound:
Copy pasted config will not work so I'd prefer to not confuse anybody.

XhmikosR:
This is more confusing, trying to find a plugin randomly. The CLI help text sorts them by name for example.

Sorting alphabetically is more natural for human beings than random sorting.

TrySound:
@XhmikosR I agree with that. But providing broken config is not human friendly.

PR #1431, opened 2021-03-17
Replace attributes iterators with for/of

TrySound:
These iterators allows to directly manipulate passed value
which does not let us to get rid from legacy "attrs" field.

Object.entries makes it easier to get an access to both attribute
name and value.

### DIFF
--- a/lib/svgo/jsAPI.js
+++ b/lib/svgo/jsAPI.js
@@ -202,6 +202,7 @@ JSAPI.prototype.hasAttr = function (name, val) {
   if (name == null) {
     return true;
   }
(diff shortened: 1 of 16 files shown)


deepsweet (approved):

TrySound:
Not significant

XhmikosR:
Pretty significant with huge SVG files, at least 20% with `https://upload.wikimedia.org/wikipedia/commons/archive/2/2e/20180507223736%21Germany_%28%2Bdistricts_%2Bmunicipalities%29_location_map_2013.svg`.

And the includes change is just more verbose than it can be...

PR #1443, opened 2021-03-21
Specify --ignore-path for ESLint

XhmikosR:
Otherwise, all files are included

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,8 +50,8 @@
   ],
   "scripts": {
     "test": "c8 --reporter=html --reporter=text mocha \"test/*/_index.js\" \"**/*.test.js\" --ignore=\"node_modules/**\"",
-    "lint": "eslint . && prettier --list-different \"**/*.js\" --ignore-path .gitignore",
-    "fix": "eslint . --fix && prettier --write \"**/*.js\" --ignore-path .gitignore",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -50,8 +50,8 @@
   ],
   "scripts": {
     "test": "c8 --reporter=html --reporter=text mocha \"test/*/_index.js\" \"**/*.test.js\" --ignore=\"node_modules/**\"",
-    "lint": "eslint . && prettier --list-different \"**/*.js\" --ignore-path .gitignore",
-    "fix": "eslint . --fix && prettier --write \"**/*.js\" --ignore-path .gitignore",
+    "lint": "eslint --ignore-path .gitignore . && prettier --list-different \"**/*.js\" --ignore-path .gitignore",
+    "fix": "eslint --ignore-path .gitignore --fix . && prettier --write \"**/*.js\" --ignore-path .gitignore",
```
TrySound, line 54: I used "ignorePatterns" in eslint config. So not all.

XhmikosR: I'd say better remove ignorePatterns then and just use .gitignore. It's the right file and includes everything that needs to be ignored.

TrySound: Ok, you can remove

PR #1445, opened 2021-03-21
removeOffCanvasPaths: Add one more test

XhmikosR:


XhmikosR:
@TrySound I still think we should add Coveralls or something like that, with the option to comment on PRs disabled. The current situation isn't ideal. And I'd also switch to `text-summary` for c8.

PR #1279, opened 2020-09-02
Add dereferenceUses plugin

strarsis:
This PR adds the dereferenceUses plugin, which dereferences `<use/>` elements, replacing it with the targetted node.
It also takes the [special attribute inheritance rules](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use) into account.

PR #1449, opened 2021-03-22
Add fix for cloning a JS Map when cloning a node

strarsis:
This PR fixes deep cloning of CSSStyleDeclaration AST nodes.

JSON.stringify is used as a classic technique for easy deep cloning of the AST nodes.
In this case the CSSStyleDeclaration AST node uses a Map for performance.
The ES MAP is not cloned by the JSON.stringify approach, it has to be cloned separately.

### DIFF
--- a/lib/svgo/css-style-declaration.js
+++ b/lib/svgo/css-style-declaration.js
@@ -34,6 +34,10 @@ CSSStyleDeclaration.prototype.clone = function (parentNode) {
 
   var clone = new CSSStyleDeclaration(parentNode);
   Object.assign(clone, nodeData);
+
+  // JS Maps require explict cloning
+  clone.properties = new Map(node.properties);
+
   return clone;
 };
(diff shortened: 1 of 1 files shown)


strarsis:
@TrySound: Would it make sense to add a test for the cloning of a `CSSStyleDeclaration` node or is it very unlikely that there will be a regression?

TrySound:
Does this fix some particular issue?

TrySound:
Tbh I don't like this abstraction and would rewrite it eventually.

strarsis:
@TrySound: Yes, this fixes an issue in the `dereferenceUses` plugin: https://github.com/svg/svgo/pull/1279#issuecomment-803706532
The current implementation mirrors the one of the CSSStyleDeclaration that can be used in the browser.
Some caching is involved to avoid unnecessary style parsing.

PR #1450, opened 2021-03-22
Implement simple node clone

TrySound:
https://github.com/svg/svgo/pull/1279 faced some problems with node cloning.
In this diff I moved class/style live objects to JSAPI and made it non
enumerable to avoid deep cloning issues.

class list and style declaration classes are not longer need own clone
methods.

cc @strarsis 

Reparsing is not a big deal since this operation is quite rare. And we achieve here less code and more correctness.

Closes https://github.com/svg/svgo/pull/1449


strarsis:
LGTM. When the caching hadn't improved performance, there is no reason to keep it.

PR #1454, opened 2021-03-27
Add "visitor" plugins support

TrySound:
Visitor is a simple pattern which helps to avoid many type checks
and provide both "perItem" and "perItemReverse" functionality without
fragmentation.

The most important case is an ability to define state which in many
plugins specified either on module level or by polluting `params`.

In this diff I added visit and detachFromParent utilities and refactored
new mergeStyles plugin with it.

Also fixed bug when cdata content is merged into "text" node which is
not always valid.

cc @strarsis @XhmikosR @sk-

### DIFF
--- a/lib/svgo/plugins.js
+++ b/lib/svgo/plugins.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const { visit } = require('../xast.js');
+
 /**
  * Plugins engine.
  *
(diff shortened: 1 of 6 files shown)


strarsis:
+1, this is a great addition.

deepsweet (approved):

XhmikosR:
LGTM from a quick look 👍

strarsis:
@TrySound: So after using this plugin type I found an improvement:
After the visitor visited all nodes it should call the plugins that use that kind of pattern a last time so the plugin can do manipulation after having visited all DOM nodes. This is necessary sometimes, e.g. when the plugin first has to collect all `<use>` and then, afterwards, query all elements that are referenced.

PR #1455, opened 2021-03-28
Sort plugin lists alphabetically

yialo:
To simplify search of particular plugin description at page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,40 +116,40 @@ See the list of the default plugins:
 ```js
 module.exports = {
   plugins: [
-    'removeDoctype',
-    'removeXMLProcInst',
-    'removeComments',
-    'removeMetadata',
-    'removeEditorsNSData',
     'cleanupAttrs',
-    'mergeStyles',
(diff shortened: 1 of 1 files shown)


TrySound:
See https://github.com/svg/svgo/pull/1430

yialo:
Is it crucial to follow this order in `plugins` array in config file?

TrySound:
Yep. Currently some plugins cannot work in isolation and requires others. I'm working on decoupling though.

polarathene:
Regarding copy/paste concern, how does that apply to the built-in plugins table? I've often found that a bit annoying to scan through. If the concern is that someone may copy/paste and edit it to add non-default plugins, then it's better to add a note before the table that order of appearance is important, then provide a copy/paste list for that which if you don't want to add too much noise to the README can link to another markdown document (or JS config), or hide the content in an expandable `<details>` block.

TrySound:
Table is fine to reorder

PR #1456, opened 2021-03-29
Collect stylesheet once per plugin

TrySound:
computeStyle(node) in isolation is quite slow utility because it
collects style elements across whole document, parses and sort them.

In this diff I splitted it into `collectStylesheet(root)` and
`computeStyle(stylesheet, node)` which are easy integrate with new
visitor plugin api.

cc @XhmikosR

### DIFF
--- a/lib/style.js
+++ b/lib/style.js
@@ -3,19 +3,13 @@
 const stable = require('stable');
 const csstree = require('css-tree');
 const specificity = require('csso/lib/restructure/prepare/specificity');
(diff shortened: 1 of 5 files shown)


deepsweet (approved):

XhmikosR:
Wow, impressive!

```
master avg 3 runs: 194s
stylesheet-cache avg 3 runs: 120s
```

Approximately -38%.

Nice!

PR #1457, opened 2021-03-29
fix: remove the <style> tag when the style is empty after CSS minified

yisibl:
This mainly includes comments and unused CSS.

## Input: 

```svg
    <style>/* test */</style>
    <rect x="10" y="10" width="100" height="100"/>
</svg>
```

## Output:

```svg
    <rect x="10" y="10" width="100" height="100"/>
</svg>

### DIFF
--- a/test/plugins/mergeStyles.12.svg
+++ b/test/plugins/mergeStyles.12.svg
@@ -1,23 +1,22 @@
 <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 48 48">
     <defs>
-        <style></style>
-        <linearGradient id="file-name_svg__file-name_svg__original-id" x1="12" y1="-1" x2="33" y2="46" gradientUnits="userSpaceOnUse">
+        <style />
+        <linearGradient id="foo" x1="12" y1="-1" x2="33" y2="46" gradientUnits="userSpaceOnUse">
             <stop offset="0" stop-color="#6b5aed" stop-opacity="0" />
             <stop offset="1" stop-color="#6b5aed" />
         </linearGradient>
     </defs>
-    <path d="M5 24a21.9 21.9 10 1 1 8 18" fill="url(#file-name_svg__file-name_svg__original-id)"/>
+    <path d="M5 24a21.9 21.9 10 1 1 8 18" fill="url(#foo)"/>
 </svg>
 
 @@@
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: test/plugins/cleanupIDs.19.svg [hunk 1]
```diff
@@ -6,7 +6,7 @@
             <stop offset="1" stop-color="#6b5aed" />
         </linearGradient>
     </defs>
-    <path d="M46 24a21.9 21.9" fill="url(#file-name_svg__file-name_svg__original-id)"/>
+    <path d="M5 24a21.9 21.9 10 1 1 8 18" fill="url(#file-name_svg__file-name_svg__original-id)"/>
 </svg>
```
yisibl, line 10: `M46 24a21.9 21.9` is invalid path data, fixed it.

test/plugins/mergeStyles.12.svg
```diff
@@ -0,0 +1,22 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 48 48">
+    <defs>
+        <style />
```
yisibl, line 3: Add a test case for the mergeStyles plugin.

TrySound:
I can't reproduce this in master. Do you know why?

yisibl:
Sorry, I didn't understand. Do you mean that you can't pull this branch of mine?

TrySound:
I mean I can't reproduce the problem in the topic in master. I still get style element with comment removed.

yisibl:
The removal of comments is the original behavior. This PR will further remove style elements that only contain comments.Please refer to the `minifyStyles.11.svg` file.

SethFalco:
Hey hey! Thanks for opening the PR.

Looks like this has already been handled in `main` and released in the latest version of SVGO, so I'll close this as resolved.

yisibl:
Thanks, which commit did this fix?

SethFalco:
> Thanks, which commit did this fix?

I apologize for the terribly slow response! Finally trying to get back to zero-inbox again. ^-^'

This was resolved in b15da27af4832bbe9efbec2bef03eff0c0ed110d (introduced in v3.0.3). It incidentally implements the same solution as you, but just with fewer tests.

I didn't add more thorough tests as my focus was on a different issue, and removing the style tag if it became empty was just an aside for me.

> Also refactors the `minifyStyles` plugin while I was working here, and removes the `style` node entirely if it's empty after we're done with it.
>
> — https://github.com/svg/svgo/pull/1800

Issue #1128, opened 2019-07-15
What is the benefit of stripping viewBox?


avwie:
Some libraries use SVGO for optimizing the SVG and I notice that stripping the viewBox is almost always used. However, I always disable stripping because it breaks proper scaling of the SVG. Unless I totally don't understand, what is the 'optimization' that stripping a viewBox provides?

GreLI:
Just as any stripping—minimize number of bytes, increase informational density.

avwie:
According to your readme:
> SVG files, especially those exported from various editors, usually contain a lot of redundant and useless information. This can include editor metadata, comments, hidden elements, default or non-optimal values and other stuff that can be safely removed or converted **without affecting the SVG rendering result.**

Removing viewBox definitely affects the rendering result, since scaling is then made almost impossible. The question is, why is it turned on by default? It isn't simple meta-data, it is critical viewport information.

GreLI:
Why? It's removed only when `width` and `height` attributes are present and equal to, so no information is lost. Finally, one can turn it off if needed.

robertknight:
Removing `viewBox` still affects the rendering output when `width` and `height` are present and `viewBox` is equal to `0 0 <width> <height>`.

I encountered this issue when an upgrade to svgo broke rendering of some scaled inline SVGs on our website (https://github.com/hypothesis/h/issues/5656). 

The SVGs in question had an outer `<svg>` element like this:

```js
<svg xmlns="http://www.w3.org/2000/svg" width="120" height="120" viewBox="0 0 120 120">
…
</svg>
```

And were being rendered with a CSS class applied that scaled them down to a much smaller size (eg.  16px x 16px).

GreLI:
@robertknight that's incorrect usage, nothing to do with SVGO. Of course if you rewrite `width` and `height`, you'll get an error, no matter in which way.

robertknight:
Please excuse the stupid question, but can you explain what you mean by incorrect usage? My understanding is that `viewBox` specifies the user-space coordinates which are then mapped to the viewport which has a default `width` and `height` controlled by the corresponding attributes on the `<svg>`, but can be overridden using CSS to rescale the image.

FWIW the original SVG file was produced some time ago by a designer using Sketch.

GreLI:
[...]

PR #1461, opened 2021-03-31
Remove the `removeViewBox` plugin from the default plugins list

JohnAlbin:
Given the discussions in #1128, we should remove the `removeViewBox` plugin from the list of default plugins.

The stated goal of **svgo** is to optimize SVGs _while not changing their renderings_. While SVGs can be rendered in many different contexts, one import context is rendering inside HTML documents. When an svg is placed inline with other HTML, removing the `viewBox` attribute (even when it is a duplicate of the data in the `width`/`height` attributes) does change how the svg is rendered.

According to the [SVG 1.1 specification, section 7.7 The ‘viewBox’ attribute](https://www.w3.org/TR/SVG11/coords.html#ViewBoxAttribute):

> It is often desirable to specify that a given set of graphics stretch to fit a particular container element. The `viewBox` attribute provides this capability.

When svgo removes the `viewBox` attribute, it removes that SVG's ability to be scaled to fit its containing HTML element via CSS. In other words, the Scalable Vector Graphic is no longer _scalable_.

Here's a codepen with an extensive example: https://codepen.io/johnalbin/pen/yLggyXv You can see from the screenshot below that the SVG does not scale properly when CSS resizes the SVG without the `viewBox` attribute.

screenshot of codepen URL above showing how viewBox affects rendering

Almost all responsive web design websites use this (or a slight variant of the following CSS):

```css
img,
svg {
  max-width: 100%;
  height: auto;
}
```

When that CSS is used, the browser treats the `width` and `height` attributes as the intrinsic size of the image as it scales the SVG _canvas_ (what the SVG spec calls the viewport), but the _contents_ of the SVG (what the spec calls the user space coordinates) are not scaled unless the `viewBox` attribute is present.

That means removing the `viewBox` attribute changes the rendering of the SVG. And, from the svgo homepage:

> SVG files, especially those exported from various editors, usually contain a lot of redundant and useless information. This can include editor metadata, comments, hidden elements, default or non-optimal values and other stuff that can be safely removed or converted **without affecting the SVG rendering result.** _(emphasis mine)_

Which means the `removeViewBox` should not be included in the list of default plugins.

Here's the PR that fixes this bug.

### DIFF
--- a/docs/03-plugins/removeViewBox.mdx
+++ b/docs/03-plugins/removeViewBox.mdx
@@ -2,7 +2,6 @@
 title: removeViewBox
 svgo:
   pluginId: removeViewBox
-  defaultPlugin: true
 ---
 
 Removes the [`viewBox`](https://developer.mozilla.org/docs/Web/SVG/Attribute/viewBox) attribute where it matches the documents width and height.
--- a/plugins/preset-default.js
+++ b/plugins/preset-default.js
@@ -16,7 +16,6 @@ import * as convertColors from './convertColors.js';
 import * as removeUnknownsAndDefaults from './removeUnknownsAndDefaults.js';
 import * as removeNonInheritableGroupAttrs from './removeNonInheritableGroupAttrs.js';
 import * as removeUselessStrokeAndFill from './removeUselessStrokeAndFill.js';
-import * as removeViewBox from './removeViewBox.js';
 import * as cleanupEnableBackground from './cleanupEnableBackground.js';
 import * as removeHiddenElems from './removeHiddenElems.js';
 import * as removeEmptyText from './removeEmptyText.js';
@@ -56,7 +55,6 @@ const presetDefault = createPreset({
     removeUnknownsAndDefaults,
     removeNonInheritableGroupAttrs,
     removeUselessStrokeAndFill,
-    removeViewBox,
     cleanupEnableBackground,
     removeHiddenElems,
     removeEmptyText,
--- a/plugins/removeDimensions.js
+++ b/plugins/removeDimensions.js
@@ -1,6 +1,6 @@
 export const name = 'removeDimensions';
 export const description =
-  'removes width and height in presence of viewBox (opposite to removeViewBox, disable it first)';
+  'removes width and height in presence of viewBox (opposite to removeViewBox)';
 
 /**
  * Remove width/height attributes and add the viewBox attribute if it's missing


ryanb:
Any update on this? It would be great to get this merged in.

 I just finished adding back a bunch of missing `viewBox` attributes because `svgo` stripped them. I assumed `svgo` default settings would not strip anything important. This PR would have saved me a lot of time.

rossmoody:
I noticed this as I was working through [my little browser extension](https://github.com/rossmoody/svg-gobbler) that uses SVGO and thought it was odd. The default configuration made it much less dependable to scale the svgs I'm scraping. I removed it from the "default config" as well.

lzl124631x:
+1 on this. Without the viewBox I can't scale the svg icons. I'm surpised that they remove it by default for saving bits.

firefoxic:
Merge this PR, please 🥺

oetiker:
meanwhile, this setting seems to help ... 

```json
{
    name: "preset-default",
    params: {
        overrides: {
           removeViewBox: false,
        }
    }
 },
 ```

JohnAlbin:
I've updated the PR to match the code updates made in prep for 3.x releases.

JohnAlbin:
@TrySound I've updated the initial PR description comment above to point out what the SVG 1.1 spec says about the `viewBox` attribute's importance.

I would very much like to see the `removeViewBox` plugin removed from the default preset. If other maintainers would like to create a preset that uses that plugin, that's fine with me, but it really shouldn't be a part of the _default_ preset for all the reasons stated above.

mahnunchik:
I'm looking forward to have it merged

XhmikosR:
@TrySound could you also have a look at this PR?

We plan to disable the plugin in SVGOMG, but if you land this it would make more sense :)

TrySound:
Breaking change 😄

XhmikosR:
Or a feature since it fixes important issues 🙂

Maybe a minor release would be OK then?

firefoxic:
> Breaking change smile

I doubt this change will break anything. Rather, it will do the opposite.

JohnAlbin:
@TrySound This is a fix for a bug that prevents all SVGs from being _scalable_ on all web pages.

So its a bug fix, not a breaking change feature request.

firefoxic:
In my [gulp-stacksvg](https://www.npmjs.com/package/gulp-stacksvg) I even have to [replace](https://github.com/firefoxic/gulp-stacksvg/blob/fd31aa5d4aa1de42d049f80466a10a5c7d318153/index.js#L53-L59) `width`/`height` with `viewBox`, mainly because of the default svgo setting.

Lucienest:
> Breaking change 😄

it fixes things (not breaking it) and makes the svgo fully SVG-compliant

Lucienest:
> In my [gulp-stacksvg](https://www.npmjs.com/package/gulp-stacksvg) I even have to [replace](https://github.com/firefoxic/gulp-stacksvg/blob/fd31aa5d4aa1de42d049f80466a10a5c7d318153/index.js#L53-L59) `width`/`height` with `viewBox`, mainly because of the default svgo setting.

I wasn't much aware of the viewBox property until the same thing happened,
why would u remove it to save a few bytes?

while most projects are bloated with countless dependencies.

hashimaziz1:
Since this project apparently now has a new maintainer, who was presumably the one who closed #1128, is any work being done to finally merge this PR?

SethFalco:
Unless I meet any resistance, I'd be happy to merge this in future. Unfortunately… this will be a long-winded response due to the heated discussions surrounding this topic historically, especially as my stance differs from other maintainers.

### Why won't I merge it now?

Based on the discussions in other issues, I'd have to treat this as a feature rather than a bug, so it is indeed a breaking change.

I would also like to give other maintainers time to contest this response if they have any qualms with it. This is important because if other maintainers disagree, it'll just be enabled again anyway. I would rather the project show a unified front than make SVGO a playground where maintainers with differing opinions override each other's changes.

With that in mind, I would like to merge this into v4.

### Why wasn't it disabled sooner?

I think the problem stems from the mentality that we must reduce bytes **at all costs**. In my opinion, this isn't what an optimizer should be striving for. In fact, reducing bytes should be 4th or 5th priority.

In my opinion, an optimizer's priorities should be:

1. Adhering to the spec. Even if clients allow us to do something, we probably shouldn't if it's not compliant with the spec.
2. Adhering to the law. We must preserve copyright notices, licenses, and attribution by default.
3. Preserving (not adding) accessibility, like `title` and `description`. Reducing bytes isn't worth discriminating against users. This also ties in with adhering to the law.
4. Preserving (not adding) compatibility, SVG 1.1 has more support than SVG 2, and we know this. That's why we still add the deprecated `xlink` namespace rather than use the `href` attribute available in SVG 2. 
5. And **finally**… we can reduce as many bytes as possible.

Optimizers aren't meant to "reduce bytes **at all costs**" because even for an optimizer, there are things more important than optimization.

### Why do I agree to disable it?

After a user reported an issue with this plugin in 23 September 2013, [Kir Belevich](https://github.com/deepsweet) (the original author of SVGO) disabled this plugin himself. But 3.5 years later, it was [silently enabled again](https://github.com/svg/svgo/commit/055e303607d53a3d539761d8727d86c03b2e88c1#diff-5e7553079e2475fcd0a5b3bc358706fe4fd8c58bfa8e8106d8843bcf1d03f2f5R5) by another maintainer with no public discussion or PR. There was a reason it was disabled, but I don't see a reason it was turned enabled again, so really that shouldn't have happened.

An optimizer should be safe by default, and the consensus is that `removeViewBox` isn't safe. I've also been bitten by this previously. Even if SVGO supports different modes/presets in future, I'd expect the safest mode to be the default mode.

It's clear that among users and engineers, this setting should be off. Even large projects like [Docusaurus](https://github.com/facebook/docusaurus/blob/e17784effa2c370d81c7806c22ad19c6dce4a1cc/packages/docusaurus-utils/src/webpackUtils.ts#L127) and [SVGR](https://react-svgr.com/docs/migrate/#svgo) disable this by default for users, and I'd like to think Meta and [Greg Bergé](https://github.com/gregberge) know what they're doing.

On [Stack Overflow](https://stackoverflow.com/search?q=%5Bsvgo%5D+viewbox), a significant number of questions with the `svgo` tag are specifically looking for how to configure SVGO in the various ways to use it to disable `removeViewBox`. It's clear there is a demand for this.

As an open-source project making software for the community, we should value what the community has to say within reasonable measure. In this case, everyone has made overwhelming clear that SVGO is currently handling this wrong, so we should step up to address it.

mryechkin:
@SethFalco well said, and thank you for taking such a reasonable approach to this issue!

firefoxic:
@SethFalco Thanks a lot! ❤️‍🔥

> I think the problem stems from the mentality that we must reduce bytes at all costs. In my opinion, this isn't what an optimizer should be striving for. In fact, reducing bytes should be 4th or 5th priority.

Probably off-topic, but I hope this revision of the approach will someday touch the default of converting relative commands of the `d` attribute into absolute ones to save bytes. Now it breaks the ability to move the whole vector figure after optimization by shifting only the coordinates of the start point.

lzl124631x:
Well said. Thanks a lot! @SethFalco

SethFalco:
@firefoxic

> Probably off-topic, but I hope this revision of the approach will someday touch the default of converting relative commands of the `d` attribute into absolute ones to save bytes. Now it breaks the ability to move the whole vector figure after optimization by shifting only the coordinates of the start point.

Yeah, I'd say it's off-topic, but I'll respond here anyway. If you or anyone else wants to add more, I'd recommend opening a separate issue, and we'll continue the discussion there. (Or shoot me an [email](mailto:seth@falco.fun)!)

That isn't something I favor myself because that impacts maintainability (human-editability) of the file, not the usability (rendering/scaling) of the vector.

Most optimizations in SVGO impact the maintainability significantly, and that applies to most other optimizers too. Whether that's relative to absolute coordinates, converting shapes to paths, inlining styles, omitting spaces between attributes or transforms, the SVG will inevitably become a hassle to maintain.

The main difference compared to other optimizers is that we've encouraged overwriting the SVG by documenting that as the primary usage and making it the default behavior of the CLI. That workflow is flawed though, considering it's annoying to edit after and if the SVG breaks, the original file is lost. 😨

I was actually planning to add a warning to the [documentation](https://svgo.dev/), that the output may be difficult to maintain, so to encourage treating the original SVG as the project file, and the output of SVGO as the final product. That's effectively what users of Docusaurus and SVGR are inadvertently doing, so they don't have that problem.

I don't want to rush into making drastic changes, but in future I do want to get feedback on if we should change the default behavior of the CLI to output to a new file called `filename.min.svg` instead of overwriting to promote this separation. 🤔

Lucienest:
Well said, I've been tracking this issue for years now

JohnAlbin:
O M G 😭 

My fellow web developers, our long svgo configuration nightmare is over.

MoritzLost:
I love how this was just silently merged with no comment, no fanfare, no great upheaval. The end of an era heralded not by a bang, but by the click of a button. One man bravely stood firm, refusing to budge even as the entire internet told him that he's wrong. And now his watch is ended.

firefoxic:
@SethFalco, just thank you so much! 🙏

SethFalco:
It'll still be a bit of time before it's officially released, but it's coming sooner rather than later. I was planning to bring attention to the removeViewBox change, but in the v4.0.0 release notes rather than in this issue/PR.

After my little hiccup with v3.3.0 and v3.3.1 due to the ESM related changes, I'm being more careful with v4.0.0. Hoping to get a release candidate out next week for anyone who may want to test and report issues. ^-^'

Reference: https://github.com/svg/svgo/releases

Meanwhile, in my [comment earlier](https://github.com/svg/svgo/pull/1461#issuecomment-1808497923), I made note of what I believe the priorities of an optimizer should be. I've been working a lot on the documentation, and will add a dedicated page to make this more explicit. Welcome to check out [svgo.dev](https://svgo.dev/) and tell us content you think is missing, and I can work on this. 👍🏽

TL;DR: Usability > Optimization

Thank you everyone for being patient, though! I know you've been waiting years for this, and even since my last comment it'd been over 6 months.

Lucienest:
> O M G 😭
> 
> My fellow web developers, our long svgo configuration nightmare is over.

A massive victory ✌️, the endless battle between arrogance and facts is now ended.

GreLI:
Just encountered this topic, and want to answer this quote:

> After a https://github.com/svg/svgo/issues/139 in 2013, [Kir Belevich](https://github.com/deepsweet) (the original author of SVGO) disabled the plugin himself. It was then [silently enabled again](https://github.com/svg/svgo/commit/055e303607d53a3d539761d8727d86c03b2e88c1#diff-5e7553079e2475fcd0a5b3bc358706fe4fd8c58bfa8e8106d8843bcf1d03f2f5R5) 3.5 years later. There was a reason it was disabled, but I don't see a public discussion/reason it was enabled again, so that probably shouldn't have happened. ^-^'

AFAIK, `removeViewBox` directly contradicted the plugin which removed `width` and `height` attributes from SVG. (`removeSizes`?) That was much worse in my opinion, since it breaks image sizes when one uses SVG as an image or background image (unless its container has the same size set). That is why `removeViewBox` was disabled. You can't have both at the same time. And there were complains about this too. So, I reverted this to `removeViewBox` enabled instead of this plugin. Moreover, problems people complain about `removeViewBox` to are the same thing — SVG images somehow loses their sizes.

I found your arguments reasonable. However, I think their should applied to the default preset. Many people using SVGO for many reasons. But I myself, Kir, and others are mostly using SVGO for web development given images (mostly icons) from design department.  In this workflow there is no need to preserve copyright, SVG2 is probably ok, 2 fractional digits is enough for the most reasonable uses, and so on. As someone says “every byte matters”. So, there should be eager preset to squeeze images at most.

SethFalco:
Hey hey! Thanks for chiming in!

> AFAIK, removeViewBox directly contradicted the plugin which removed width and height attributes from SVG. (removeSizes?)

There is a plugin which is the implied opposite called [removeDimensions](https://svgo.dev/docs/plugins/removeDimensions/), but they don't contradict each other. They have different purposes.

Width and height are to set the intended size to display the image/document to a consumer such as an SVG viewer or browser, while viewBox is to set the virtual size within the document itself. Both are functionally distinct and just as important as the other.

It's a similar relationship to viewport size (in-game dimensions) vs. window size (operating system window dimensions) in the realm of game development.

Or in other words, `viewBox` is for the document's internal coordinate system, while `width` and `height` are for indicating how it'd like to be presented in an external/parent coordinate system.

> That was much worse in my opinion, since it breaks image sizes when one uses SVG as an image or background image (unless its container has the same size set).

Just my thoughts, but neither should get removed by default. Even in the case that both are equal:

* if we remove `viewBox`, then we lose scaling, which may be desirable and probably the very reason it was present in the SVG in the first place.
* if we remove `width` and `height`, we lose the original size, which may be desirable in certain cases like icons.

However, based on the direction modern web development taken, as well as from my experiences, scalability is a more common issue. Websites are typically developed mobile first and are designed with responsiveness in mind, where having a fixed size for assets isn't a common use case anymore, even for icons. Icon sizes are typically based on the device width or font-size, rather than the original `width` and `height` in the document. (A static `width` and `height` value can't be responsive.)

I think it's very possible that at the time the decision was made, it made sense to use removeViewBox, but I don't think it works out with how websites are developed **_today_**.

> That is why removeViewBox was disabled. You can't have both at the same time. And there were complains about this too.

You can have both at the same time!

If you are able to point me to the where having both lead to a problem, I'd be happy to investigate it. However, if both being defined causes issues, I could only imagine it was an issue in the SVG itself, and not SVGO's optimization of it.

> But I myself, Kir, and others are mostly using SVGO for web development given images (mostly icons) from design department.

This is also how I use SVGO, and I believe how many of our users too, who have encountered these issues. In fact, where I'm working atm, we had to disable removeViewBox too. ^-^'

Furthermore, we even have to disable removeViewBox for [svgo.dev](https://svgo.dev), otherwise it'd screw up the responsiveness of the website.

> In this workflow there is no need to preserve copyright

That's fair, in some workflows copyright information doesn't have to be preserved. But SVGO certainly shouldn't be putting users at risk of breaching licenses by default. For example, many are using SVGs from [Creative Commons Search](https://search.creativecommons.org/), downloading from [Font Awesome](https://fontawesome.com/), or from Wikipedia.

It's like proposing that websites/companies should delete `LICENSE` files by default before serving their software because it's bloat. But, it's not bloat, it's the license, and they're required to serve the license and attribute the copyright holders if they use the software/library.

Given we're open-source maintainers that distribute licenses ourselves, I'm sure it's easy to understand that others want their licenses respected too. Removing license information (which there are use cases for) should be a conscious decision, not an accident.

> So, there should be eager preset to squeeze images at most.

That's fair, in fact, I can even pick this up as part of the next release. I'll work on it next week.

I was thinking about making presets for two other purposes too:

* sprite sheets - a few people have raised that they wanted this
* inlined SVGs - it's such a popular use-case, and we have a few plugins that were made just for this

> That's why SVGO was created in the first place.

I'm not personally familiar with the motivation and mindset that went into creating SVGO when it started. However, I think I can comfortably describe what a modern take an optimizer should be and what the expectations are, most businesses value legal compliance and accessibility a lot more than shaving bytes. Reducing bytes is very important, but not if the alternative is losing users or even litigation.

Not sure about other places, but in the US and UK though, companies can get fined for discimination if they don't build software/websites with accessibility in mind.

GreLI:
> You can have both properties at the same time!

I meant that if you have both plugins enabled (especially by default) it tottally ruins your SVG image. That's why one of them had to be disable.

firefoxic:
> That's why one of them had to be disabled.

But that doesn't answer the question of why `removeViewBox` was left out of the two plugins. It's already been said more than once above, but I'll say it again:
- The `viewBox` attribute is a characteristic of the SVG internals and therefore should remain within the SVG file itself. Because it indicates which part of the infinite canvas to render. It's like a frame in Figma (which incidentally becomes the `viewBox` attribute when exporting to SVG).
- The `width` and `height` attributes are external (to the SVG file) because they are the dimensions of where the SVG file needs to be embedded, whether in markup (`` or `<svg>` tags) or in styles (`background-size` or `mask-size`).

And arguments like “every byte matters” are strange, because the string
 `viewBox="0 0 80 80"` is shorter than `width="80" height="80"` (I hope nobody will think that we are asking to replace `removeViewBox` with `removeDimensions` because of saving these 3 bytes 🤭).

SethFalco:
> But that doesn't answer the question of why `removeViewBox` was left out of the two plugins

I believe that was addressed, actually.

> That was much worse in my opinion, since it breaks image sizes when one uses SVG as an image or background image
> 
> — https://github.com/svg/svgo/pull/1461#issuecomment-2156871372

In a world where one of the two must be enabled to reduce bytes, there is a trade off:

1. Sacrifice scalability (removeViewBox)
2. Sacrifice the original dimensions the designer exported the SVG with, so it will always fill the parent element unless overridden otherwise (removeDimensions)

removeViewBox was disabled because it was decided that using the original dimensions was more valuable, or at least more correct, than the scalability. (Though this doesn't adhere to modern design paradigms.)

The reason I have a different perspective is that I think it's sensible to take the fourth option, which is not to enable either of them. Though if I were to enable one, I'd definitely go for `removeDimensions` personally, since that's overridable in CSS, while removeViewBox can't be worked around after the fact.

---

In any case, I don't think there's any need to drag this out too much. It sounds to me like Lev mostly wanted to share the motivations behind enabling it which I did value, and I'm happy to introduce an aggressive preset which includes plugins like removeViewBox and perhaps some of our newer plugins that haven't been promoted to preset-default yet.

firefoxic:
> the fourth option, which is not to enable either of them.

Now that's sad news. I thought that I would be able to delete the whole config file, which does nothing but replace one plugin with another. But it seems that only one line of this config file can be removed.

GreLI:
> …It sounds to me like Lev mostly wanted to share the motivations behind enabling it…

That's correct. I just wanted to clear what happened then, and what I realized at the time when I was digging it. Nothing else.

PR #1464, opened 2021-04-05
Improve docs for convertColors

JohnAlbin:
The docs for most of the plugins are… scant.

I was writing a custom plugin when I realized that the `convertColors` plugin already did what I wanted, but the `currentColor` param was completely undocumented.

Here's a PR to fix that part of the docs.


TrySound:
Please run "npm run fix"

SethFalco:
Thanks for opening the PR and proposing we document this behavior. This has been resolved through publishing documentation svgo.dev.

You can read more about `convertColor` and the `currentColor` param here:
https://svgo.dev/docs/plugins/convert-colors/#parameters

Contributions to improve our documentation are welcome! 👍🏽

PR #1471, opened 2021-04-17
Add hsl() support

### LINKED ISSUE #1468: convertColors: HSL support
Add support for HSL colors to the convertColors plugins.

vanillajonathan:
Add HSL support, closes #1468


omgovich:
Hi! `hsl()` function has different syntaxes and might contain different units.
The regexp you added doesn't process alpha values, `hsla` prefix, `deg` and `rad` units, modern slash syntax, etc.

PR #1472, opened 2021-04-17
Mark convertStyleToAttrs as `disabled` in README

fcaylus:
Hello !

The PR #1365 disabled `convertStyleToAttrs` plugin by default. However, it's still marked as `enabled` in the README.

Simply fixes that by marking it as `disabled` !


TrySound:
Thanks!

Issue #1476, opened 2021-05-19
Suboptimal output after upgrading to version 2.3


joseprio:
**Describe the bug**
Seems to be a regression of #1403; after updating from version 1 to 2, I noticed an increase in file size. I applied the additional configuration from #1353, and while it improved the situation, the there's still some superfluous spaces added to the arc command.

**To Reproduce**
Original SVG optimized with version 1:
```
<svg viewBox="0 0 18 18"><path fill-rule="evenodd" d="M8.153 3.352h6.669c.65 0 1.178.526 1.178 1.174v8.3c0 .648-.528 1.174-1.178 1.174H3.177C2.527 14 2 13.474
2 12.825V4.68c.014-.248.109-.484.269-.674l1.61-1.719c.18-.17.414-.272.662-.286h1.891c.248.014.484.113.667.28l1.054 1.072zm.835 1.977v.004a3.344 3.344 0 00-3.35
3.337 3.344 3.344 0 003.351 3.336 3.344 3.344 0 003.347-3.34c0-1.843-1.5-3.337-3.348-3.337zm.01.977a2.36 2.36 0 012.18 1.452c.365.879.163 1.89-.512 2.563a2.364
2.364 0 01-2.572.51 2.352 2.352 0 01-1.456-2.174 2.355 2.355 0 012.36-2.347v-.004z"/></svg>
```
After applying 2.3:
```
<svg viewBox="0 0 18 18"><path fill-rule="evenodd" d="M8.153 3.352h6.669c.65 0 1.178.526 1.178 1.174v8.3c0 .648-.528 1.174-1.178 1.174H3.177C2.527 14 2 13.474
2 12.825V4.68c.014-.248.109-.484.269-.674l1.61-1.719c.18-.17.414-.272.662-.286h1.891c.248.014.484.113.667.28l1.054 1.072zm.835 1.977v.004a3.344 3.344 0 00-3.35
3.337 3.344 3.344 0 0 0 3.351 3.336 3.344 3.344 0 0 0 3.347-3.34c0-1.843-1.5-3.337-3.348-3.337zm.01.977a2.36 2.36 0 012.18 1.452c.365.879.163 1.89-.512 2.563a2.
364 2.364 0 01-2.572.51 2.352 2.352 0 0 1-1.456-2.174 2.355 2.355 0 0 1 2.36-2.347v-.004z"/></svg>
```
`a3.344 3.344 0 00-3.35
[...]

PR #1484, opened 2021-06-04
Remove flag spaces for all arcs

joseprio:
This is a fix for issue #1476, which I reported some days ago. The issue is that a path command may have multiple arcs, and the space optimization was only applied to the first one. Modified a test to check it.


TrySound:
Thank you!

joseprio:
Thanks for this great utility @TrySound !

PR #1485, opened 2021-06-05
Upgrade css-select from `^3.1.2` to `^4.1.3`

### LINKED ISSUE #1488: "High" Severity Audit from dependency `css-select` and `css-what`
**Describe the bug**
Per [this advisory](https://www.npmjs.com/advisories/1754) installations that include `svgo` now cause an `npm audit` warning to appear because of the dependency on 3.x versions of `css-select`, which in turn depends on a version of `css-what` older than 5.0.1.

**To Reproduce**
1. Add `svgo` as a dependency.
2. Run `npm audit`

**Expected behavior**
No audit should appear
 
**Proposed fix**
Upgrade the dependency on `css-select` to be `^4.1.3` since 4.1.3 bumps _their_ dependency on `css-what` to 5.0.1 and fixes this issue.

ericcornelissen:
Resolves #1488

This updates [the `css-select` dependency](https://github.com/svg/svgo/blob/59876d894ba758814a224cffe26566104018130d/package.json#L98) from `^3.1.2` to `^4.1.2` in order to get [the transitive dependency `css-what`](https://github.com/svg/svgo/blob/59876d894ba758814a224cffe26566104018130d/package-lock.json#L627-L631) updated to `5.0.1`, which [was vulnerable to ReDoS up-to-and-including `5.0.0`](https://app.snyk.io/vuln/SNYK-JS-CSSWHAT-1298035). Given my understanding of SVGO, this means SVGO may be vulnerable as well when provided with a carefully crafted SVG containing CSS.

Alternatively, `css-select` can be set to as low as `^4.0.0` in order to get `css-what` updated to an appropriate version. I just set it to `^4.1.2` as that is the latest version as of opening this Pull Request.

As this concerns a vulnerable dependency that is also used in the latest SVGO v1, ideally SVGO v1 should also be patched. I'm not sure what the procedure is for updating SVGO v1 though...

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "@trysound/sax": "0.1.1",
     "chalk": "^4.1.0",
     "commander": "^7.1.0",
-    "css-select": "^4.1.2",
+    "css-select": "^4.1.3",
     "css-tree": "^1.1.2",
     "csso": "^4.2.0",
     "stable": "^0.1.8"
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "@trysound/sax": "0.1.1",
     "chalk": "^4.1.0",
     "commander": "^7.1.0",
-    "css-select": "^3.1.2",
+    "css-select": "^4.1.2",
     "css-tree": "^1.1.2",
     "csso": "^4.2.0",
     "stable": "^0.1.8"
--- a/package-lock.json
+++ b/package-lock.json
@@ -604,9 +604,9 @@
       }
     },
     "css-select": {
-      "version": "4.1.2",
-      "resolved": "https://registry.npmjs.org/css-select/-/css-select-4.1.2.tgz",
-      "integrity": "sha512-nu5ye2Hg/4ISq4XqdLY2bEatAcLIdt3OYGFc9Tm9n7VSlFBcfRv0gBNksHRgSdUDQGtN3XrZ94ztW+NfzkFSUw==",
+      "version": "4.1.3",
+      "resolved": "https://registry.npmjs.org/css-select/-/css-select-4.1.3.tgz",
+      "integrity": "sha512-gT3wBNd9Nj49rAbmtFHj1cljIAOLYSX1nZ8CB7TBO3INYckygm5B7LISU/szY//YmdiSLbJvDLOx9VnMVpMBxA==",
       "requires": {
         "boolbase": "^1.0.0",
         "css-what": "^5.0.0",
@@ -737,9 +737,9 @@
       }
     },
     "domutils": {
-      "version": "2.6.0",
-      "resolved": "https://registry.npmjs.org/domutils/-/domutils-2.6.0.tgz",
-      "integrity": "sha512-y0BezHuy4MDYxh6OvolXYsH+1EMGmFbwv5FKW7ovwMG6zTPWqNPq3WF9ayZssFq+UlKdffGLbOEaghNdaOm1WA==",
+      "version": "2.7.0",
+      "resolved": "https://registry.npmjs.org/domutils/-/domutils-2.7.0.tgz",
+      "integrity": "sha512-8eaHa17IwJUPAiB+SoTYBo5mCdeMgdcAoXJ59m6DT1vw+5iLS3gNoqYaRowaBKtGVrOF1Jz4yDTgYKLK2kvfJg==",
       "requires": {
         "dom-serializer": "^1.0.1",
         "domelementtype": "^2.2.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -604,14 +604,14 @@
       }
     },
     "css-select": {
-      "version": "3.1.2",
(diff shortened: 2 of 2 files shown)


sgerrand:
👋 @TrySound, are you able to review this change?

aleclarson (approved):

zackdotcomputer:
Noting that this "Resolves #1488" in the hopes that that magic phrase will close the issue when/if this get merged (I know that works if it's in the description of a PR, guess we'll find out if it works if its in a comment...)

zackdotcomputer:
FYI the breaking changes from css-select 3.1.2 -> 4.1.2 changelog (for someone with better knowledge of this codebase to be able to more easily review):
- Several built-in pseudos are now stricter. This aligns them with the CSS spec, but might lead to changes in results.
- In HTML, attributes are now automatically considered case-insensitive, based on the HTML spec. Some selectors might now match more elements.
- Removed strict option.

aleclarson:
@underoot This is a breaking change, as it affects CSS selectors in `<style>` elements, which requires `svgo` to bump to v3 major. Therefore, patching v1 is not viable.

aleclarson:
For Yarn users (in your package.json):
```json
"resolutions": {
  "**/css-what": "^5.0.0"
}
```
This should be safe to do, as it avoids any of the breaking changes introduced in `css-select@4`.

ericcornelissen:
> This is a breaking change, as it affects CSS selectors in <style> elements, which requires svgo to bump to v3 major. Therefore, patching v1 is not viable.

Given that [the changes are related to spec compliance](https://github.com/svg/svgo/pull/1485#issuecomment-857563882) (and the `strict` option doesn't seem to be used by SVGO) I'd be inclined to disagree. If you're using non-standard CSS selectors I think it's fair that SVGO doesn't support that out-of-the-box, and the fact that it did could be considered a bug. I think it's more important people on older versions of SVGO can easily update, than that a presumably small minority that uses non-standard selectors isn't bothered by an update.

Though of course it's up to the SVGO maintainers to decide.

> Edit: RIP me - it doesn't

Did it for you :wink:

package.json
```diff
@@ -95,7 +95,7 @@
     "@trysound/sax": "0.1.1",
     "chalk": "^4.1.0",
     "commander": "^7.1.0",
-    "css-select": "^3.1.2",
+    "css-select": "^4.1.2",
```
seahindeniz, line 98: Shouldn't it be 4.1.3 considering the conversation in here https://github.com/fb55/css-select/issues/468

silkfire: 4.1.3 was released a day after this PR was opened :)

seahindeniz: Oh, 🙂 well, it's still not too late to change it

ericcornelissen: Thanks for pointing this out @seahindeniz, I updated the Pull Request :+1:

seahindeniz (approved):

carboneater (approved):

silkfire (approved):

rejas (approved):

joelpittet (approved):

seahindeniz:
Christ. Enough with the approvals. Just merge it already! 😢

silkfire:
Looks like @TrySound is AWOL

JoKalliauer:
> Looks like @TrySound is AWOL

No I don't think so, I think we just need to be more patient/uncomplaining.

On Discord in SVGO: https://discord.gg/z8jX8NYxrE 

> @TrySound answered 21.06.21 07:31UTC (i.e. 09:31 CEST)
https://discord.com/channels/815166721315831868/815166722041839668/856436221604528128
>Hi, I will publish new versions today

And if you check https://github.com/TrySound , TrySound did 9 contriutions yesterday and 3 contributions today. I think it is just different priorities.

However @GreLI and @deepsweet  are imho also members of https://github.com/svg and might be able also be able to merge?

tlusk (approved):

felabr (approved):

seahindeniz:
Approving much?

sigveio:
In case it's not clear for everyone: you don't have to be an active participant to receive updates. Use the subscribe option instead. And please avoid doing reviews or posting comments just to "bump" the thread - as it can fire off unnecessary emails to a lot of people.

And for those who just want to catch the merge, you can update your preference like so:
(Found by clicking "Customize" next to the Subscribe button)

Screenshot 2021-06-23 at 13 18 35

Above all; please be patient and keep in mind that maintainers have lives/souls too :-)

To keep things clean I'd humbly suggest that we keep activity in this thread directly relevant to the nature of the PR (e.g. discussing potential issues with merging it), and that anything else goes in the linked issue (#1488) instead.

TrySound:
Thanks!

TrySound:
Released https://github.com/svg/svgo/releases/tag/v2.3.1

ericcornelissen:
> As this concerns a vulnerable dependency that is also used in the latest SVGO v1, ideally SVGO v1 should also be patched. I'm not sure what the procedure is for updating SVGO v1 though...

@TrySound, do you have any comment in this regard?

PR #1490, opened 2021-06-11
Update css-what

mischnic:
This updates `css-select` to use a version of the transitive dependency `css-what` that isn't marked as vulnerable: https://www.npmjs.com/advisories/1754

As far as I can tell, the breaking changes don't affect svgo: https://github.com/fb55/css-select/releases

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -604,14 +604,14 @@
       }
     },
     "css-select": {
-      "version": "3.1.2",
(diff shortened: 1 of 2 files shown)


midnight-wonderer:
Why do you need to create a new request?
What's wrong with https://github.com/svg/svgo/pull/1485?

mischnic:
Whoops, I didn't check beforehand 🙈

PR #1506, opened 2021-07-24
Enable sortAttrs plugin by default

sonnyp:
It was disabled by default in https://github.com/svg/svgo/commit/b9bbe7a61ba4666510226940fce0ec5c0df0dc83 without explanation - perhaps @deepsweet remembers?

Enabling it produces (slightly) smaller outputs when gzipped. You can verify with https://jakearchibald.github.io/svgomg/


TrySound:
I plan to enable it in v3. Btw need to add to default preset as well.

TrySound:
Fixed in https://github.com/svg/svgo/releases/tag/v3.0.0

PR #1508, opened 2021-07-25
Remove xmlns:xlink in removeXMLNS plugin

devongovett:
xmlns:xlink is also not required in SVGs embedded in HTML

### DIFF
--- a/plugins/removeXMLNS.js
+++ b/plugins/removeXMLNS.js
@@ -23,5 +23,6 @@ exports.description =
 exports.fn = function (item) {
   if (item.type === 'element' && item.name === 'svg') {
     delete item.attributes.xmlns;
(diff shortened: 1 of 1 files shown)


TrySound:
Hi @devongovett. Thanks for PR. I think [removeUnusedNS.js](https://github.com/svg/svgo/blob/master/plugins/removeUnusedNS.js) removes xmlns:xlink.

devongovett:
That only removes it if it's unused though. In this case, it's always safe to remove xlink:href in HTML. Since this plugin is opt-in, I think it makes sense to do it here as well.

TrySound:
I think it's better to create a separate plugin which removes xlink namespace from xlink:herf as well.
I can take care of it in a few days.

PR #1524, opened 2021-08-14
removeAttributesBySelector.js: fix wrong `map` usage

XhmikosR:
Nothing was returned so `map` was wrong. Switched to `const` while at it.


TrySound:
Thanks. I already refactored this plugin locally. Will close to not have conflicts.

PR #1531, opened 2021-08-20
Clarify how to enable/configure builtin plugins

oBusk:
I just spent quite a while trying to upgrade to svgo 2.4 because I was thinking that all the configuration I was passing to `extendDefaultPlugins()` previously should now be in `preset-default` overrides. 

Just adding some examples of enabling other builtin plugins could clarify it? 

Maybe I was just being stupid, but a thought :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ module.exports = {
       name: 'preset-default',
       params: {
         overrides: {
-          // customize options for plugins that are enabled by default
+          // customize options for plugins included in preset
           builtinPluginName: {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,7 +89,7 @@ module.exports = {
       name: 'preset-default',
       params: {
         overrides: {
-          // customize options
+          // customize options for plugins that are enabled by default
```
TrySound, line 92: "for plugins included in preset"

PR #1535, opened 2021-08-22
[new plugin] remove-xlink-ns

TrySound:
Ref https://github.com/svg/svgo/pull/1508

The new plugin removes xmlns:xlink from <svg> and replaces xlink:href
with href attribute.

xlink namespace is obsolete in SVG 2. Href attribute is recommended
replacement to xlink:href.

This plugin will be enabled by default in SVGO 3.

See https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/href

cc @devongovett

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,7 +63,7 @@ Our documentation is maintained in [MDX](https://mdxjs.com/), which is Markdown 
 
 To preview local changes, follow the steps to run the website locally in [svg/svgo.dev](https://github.com/svg/svgo.dev).
 
-New plugins, plugin parameters, and notable any features should be paired with documentation in the [`docs/`](./docs/) directory and pushed in the same pull request.
+New plugins, plugin parameters, and notable features should be paired with documentation in the [`docs/`](./docs/) directory and included in the same pull request.
 
 ## Funding
 
--- a/docs/03-plugins/convert-colors.mdx
+++ b/docs/03-plugins/convert-colors.mdx
@@ -8,22 +8,22 @@ svgo:
       description: If to convert all instances of a color to <code>currentColor</code>. This means to inherit the active foreground color, for example in HTML5 this would be the <a href="https://developer.mozilla.org/docs/Web/CSS/color" target="_blank"><code>color</code></a> property in CSS.
       default: false
     names2hex:
-      description: If to convert color names to the hex equivilent.
+      description: If to convert color names to the hex equivalent.
       default: true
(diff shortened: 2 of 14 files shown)


deepsweet (approved):

plugins/remove-xlink-ns.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+exports.fn = () => {
+  return {
+    element: {
+      enter: (node) => {
+        if (node.attributes['xmlns:xlink'] != null) {
+          delete node.attributes['xmlns:xlink'];
+        }
+        if (node.attributes['xlink:href'] != null) {
+          node.attributes['href'] = node.attributes['xlink:href'];
```
devongovett, line 27: I don't think this is always correct. Some older SVG elements don't support `href` and _only_ support `xlink:href`. Mostly elements that were removed in SVG2 but are still supported in engines. Examples include `cursor`, `filter`, `tref` and other SVG fonts related attributes.

You should either have a list of elements this is valid for, or don't do this transform. And therefore, removing `xmlns:xlink` is also not safe when one of those elements that require `xlink:href` is present. However, it should be an option because `xmlns:xlink` is _never_required in HTML.

SethFalco:
I took over this PR and implemented the plugin. It ended up being a bit more complicated, though! ^-^'

I did temporarily set `removeXlink` as a default plugin and can confirm regression tests 100% pass. 🎉

Differences:

* This migrates the whole XLink namespace rather than just the `xlink:href` attribute. For example, `xlink:show` and `xlink:title` too.
* This checks if the XLink namespace was added with other prefixes at some point in the document.

I took the equivalent SVG 2 syntax from MDN.

* `xlink:href` → `href`
* `xlink:show` → `target` (`new` → `_blank`, `replace` → `_self`)
* `xlink:title` → `<title>` in the child node

plugins/remove-xlink-ns.js
SethFalco: Thanks to both of you for the feedback. I've taken over and cleaned up the PR, so hopefully it should be good to merge.

* [devongovett](https://github.com/devongovett), I've done as you've proposed and made the plugin preserve XLink in that case. It applies for `<cursor>`, `<filter>`, `<font-face-uri>`, `<glyphRef>`, and `<tref>`, with a parameter to force it.
* [yisibl](https://github.com/yisibl), I do not plan to enable the plugin by default myself, but we'll see if any other maintainer would like to chime in.

Issue #951, opened 2018-04-24
No newline at end of file or Do not remove newline at end of file option


schalkneethling:
When using the SVGO-CLI on `.svg` files, the output removes the newline at the end of the file. Is there a way to tell SVGO to not remove this?

Currently using it as follows:

```
svgo jinja2/includes/icons/* --disable=removeViewBox
```

schalkneethling:
Enabling `--pretty` pretty much solves this problem.

adalinesimonian:
Support for enforcing a final newline (as well as choosing what EOL to use) has landed in e8321f0c27b563734d9c4d7240a1b5a0b904b4ec.

PR #1546, opened 2021-08-25
Allow choosing EOL and appending final newline

adalinesimonian:
Addresses #951

Adds the following options to js2svg and CLI:

- eol — can be set to `lf` or `crlf`. If unspecified, js2svg uses the platform EOL.
- finalNewline — defaults to false. If true, js2svg ensures any SVG output has a final newline.

Tests added to cover both options.

### DIFF
--- a/lib/svgo.test.js
+++ b/lib/svgo.test.js
@@ -1,9 +1,5 @@
 'use strict';
 
-jest.mock('os', () => ({
-  EOL: '\r\n',
-}));
-
 const { optimize } = require('./svgo.js');
 
 test('allow to setup default preset', () => {
@@ -69,31 +65,6 @@ test('allow to disable and customize plugins in preset', () => {
 });
 
 describe('allow to configure EOL', () => {
-  // See mock at top of test file
-  afterAll(() => {
-    jest.unmock('os');
-  });
-
-  test('should default to platform EOL', () => {
-    const svg = `
-      <?xml version="1.0" encoding="utf-8"?>
-      <svg viewBox="0 0 120 120">
-        <desc>
-          Not standard description
-        </desc>
-        <circle fill="#ff0000" cx="60" cy="60" r="50"/>
-      </svg>
-    `;
-    // using toEqual because line endings matter in these tests
-    expect(
-      optimize(svg, {
-        js2svg: { pretty: true, indent: 2 },
-      }).data
-    ).toEqual(
-      `<svg viewBox="0 0 120 120">\r\n  <circle fill="red" cx="60" cy="60" r="50"/>\r\n</svg>\r\n`
-    );
-  });
-
   test('should respect EOL set to LF', () => {
     const svg = `
       <?xml version="1.0" encoding="utf-8"?>
@@ -128,6 +99,26 @@ describe('allow to configure EOL', () => {
     expect(
       optimize(svg, {
         js2svg: { eol: 'crlf', pretty: true, indent: 2 },
+      }).data
+    ).toEqual(
+      `<svg viewBox="0 0 120 120">\r\n  <circle fill="red" cx="60" cy="60" r="50"/>\r\n</svg>\r\n`
+    );
+  });
+
+  test('should respect EOL set to line break', () => {
+    const svg = `
+      <?xml version="1.0" encoding="utf-8"?>
+      <svg viewBox="0 0 120 120">
+        <desc>
+          Not standard description
+        </desc>
+        <circle fill="#ff0000" cx="60" cy="60" r="50"/>
+      </svg>
+    `;
+    // using toEqual because line endings matter in these tests
+    expect(
+      optimize(svg, {
+        js2svg: { eol: '\r\n', pretty: true, indent: 2 },
       }).data
     ).toEqual(
       `<svg viewBox="0 0 120 120">\r\n  <circle fill="red" cx="60" cy="60" r="50"/>\r\n</svg>\r\n`
--- a/lib/svgo.test.js
+++ b/lib/svgo.test.js
@@ -1,5 +1,9 @@
 'use strict';
 
+jest.mock('os', () => ({
+  EOL: '\r\n',
+}));
+
 const { optimize } = require('./svgo.js');
 
 test('allow to setup default preset', () => {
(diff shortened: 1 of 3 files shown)


TrySound:
Hi, could you clarify what's the point of finalNewline option?

lib/svgo/js2svg.js
```diff
@@ -28,6 +28,8 @@ var defaults = {
   encodeEntity: encodeEntity,
   pretty: false,
   useShortTags: true,
+  eol: undefined,
```
TrySound, line 31: Let's just pass `\n` or `\r\n` here

lib/svgo.test.js
```diff
@@ -1,5 +1,9 @@
 'use strict';
 
+jest.mock('os', () => ({
+  EOL: '\r\n',
```
TrySound, line 4: I would prefer to avoid mocks and use only js2svg options

adalinesimonian:
> Hi, could you clarify what's the point of finalNewline option?

In a number of workspaces, files are required to have a final newline. For example, [`git diff`](https://i.stack.imgur.com/BgAyw.png), [GitHub's website](https://i.stack.imgur.com/0qC9r.png), and [other tools](https://stackoverflow.com/a/37684763) will either warn of missing newlines or behave improperly when encountering files that are missing a final newline.

lib/svgo/js2svg.js
adalinesimonian: In other words, `platformEOL` goes here?

TrySound: yep

lib/svgo.test.js
adalinesimonian: The point of the mock is to test whether or not js2svg is defaulting to using the platform EOL; without a mock, we'd just be testing js2svg's behaviour when given options and not its platform-specific behaviour.

We could get rid of it if the test isn't all that important, or perhaps you have a better idea that I didn't think of?

TrySound: Better get rid of it. I will move os import to svgo-node.js so svgo.js import could be platform agnostic

adalinesimonian: Addressed in latest push.

lib/svgo/js2svg.js
adalinesimonian: Addressed in latest push.

lib/svgo.test.js
```diff
@@ -64,6 +64,127 @@ test('allow to disable and customize plugins in preset', () => {
...
+      <svg viewBox="0 0 120 120">
+        <desc>
+          Not standard description
+        </desc>
+        <circle fill="#ff0000" cx="60" cy="60" r="50"/>
+      </svg>
+    `;
+    // using toEqual because line endings matter in these tests
+    expect(
+      optimize(svg, {
```
TrySound, line 80: Could you please split result into separate variable. This is how I refactored all cases in other branch
```
const { data } = optimize(...)
```

lib/svgo/js2svg.js
```diff
@@ -64,15 +66,26 @@ function JS2SVG(config) {
...
+  if (typeof this.config.eol === 'string' && this.config.eol) {
+    const configuredEOL = this.config.eol.toLowerCase();
+    if (configuredEOL === 'lf') {
+      this.eol = '\n';
+    } else if (configuredEOL === 'crlf') {
+      this.eol = '\r\n';
+    } else {
+      this.eol = this.config.eol;
+    }
+  }
```
TrySound, line 78: This conversion is no longer necessary since we pass newlines directly

adalinesimonian: I'm concerned that if we change the configuration option to only accept the actual line break characters themselves as opposed to supplying `--eol crlf` or `--eol lf`, line break characters may have unintended consequences in certain commands or configuration files. That's why I originally made the option take `crlf` or `lf`, similar to how other tools and configurations handle such options, such as git or editorconfig.

lib/svgo.test.js
adalinesimonian: Addressed in latest push.

lib/svgo/js2svg.js
TrySound: I see. Ok, we better avoid config complexity and stick with `lf`/`crlf` enum without else case. And in defaults please derive `lf`/`crlf` from os.eol.

lib/svgo/js2svg.js
```diff
@@ -64,15 +66,26 @@ function JS2SVG(config) {
     this.config.indent = '    ';
   }
 
+  if (typeof this.config.eol === 'string' && this.config.eol) {
```
TrySound, line 69: this.config.eol != null should be enough here

lib/svgo/js2svg.js
```diff
@@ -118,6 +131,15 @@ JS2SVG.prototype.convert = function (data) {
 
   this.indentLevel--;
 
+  if (
+    this.config.finalNewline &&
+    this.indentLevel === 0 &&
+    svg.length > 0 &&
+    svg[svg.length - 1] !== '\n'
```
TrySound, line 138: Is there a case where resulting svg may contain final newline?

lib/svgo/js2svg.js
```diff
@@ -64,15 +66,26 @@ function JS2SVG(config) {
     this.config.indent = '    ';
   }
 
+  if (typeof this.config.eol === 'string' && this.config.eol) {
+    const configuredEOL = this.config.eol.toLowerCase();
+    if (configuredEOL === 'lf') {
```
TrySound, line 71: toLowerCase probably also not necessary. Will conflict with types anyway. Btw type and null checks will not be necessary then at all

lib/svgo/js2svg.js
adalinesimonian: Yes, when `pretty` is true.

lib/svgo.test.js
```diff
@@ -64,6 +64,123 @@ test('allow to disable and customize plugins in preset', () => {
...
+        <desc>
+          Not standard description
+        </desc>
+        <circle fill="#ff0000" cx="60" cy="60" r="50"/>
+      </svg>
+    `;
+    const { data } = optimize(svg, {
+      js2svg: { finalNewline: true, pretty: true, indent: 2 },
+    });
+    expect(data).toMatchInlineSnapshot(`
```
TrySound, line 175: I'd use toEqual in all cases here to highlight final newlines as well

lib/svgo/js2svg.js
adalinesimonian: Addressed in 21bb65f.

lib/svgo/js2svg.js
adalinesimonian: Addressed in 21bb65f.

lib/svgo.test.js
adalinesimonian: Addressed in 21bb65f.

lib/svgo/js2svg.js
adalinesimonian: Addressed in 21bb65f.

lib/svgo/coa.js
```diff
@@ -112,6 +117,16 @@ async function action(args, opts, command) {
     }
   }
 
+  if (opts.eol != null) {
+    const eol = opts.eol.toLowerCase();
```
TrySound, line 121: Remove this too please. Better have configuration consistent.

adalinesimonian: Whoops, forgot about this file! Addressed in https://github.com/svg/svgo/pull/1546/commits/592c01fdf17ca38c87ce0b02fabb62dbd206225e.

lib/svgo/coa.js
```diff
@@ -112,6 +117,15 @@ async function action(args, opts, command) {
     }
   }
 
+  if (opts.eol != null) {
```
TrySound, line 120: Null check is not necessary anymore

TrySound: Ah, no. My bad. All good.

TrySound: if statements can be merged together

adalinesimonian: Addressed in https://github.com/svg/svgo/pull/1546/commits/c413409d3e68e53c78679dd1118cc5adffa145a7.

TrySound:
Thank you for working on this @adalinesimonian!

adalinesimonian:
> Thank you for working on this @adalinesimonian!

Наконец-то! It's been a pleasure. 🎉

PR #1553, opened 2021-09-01
Add better parser errors

TrySound:
Old SVGO errors were not very helpful. Packages like cssnano
(postcss-svgo) had to deal with a lot of issues which are hard to debug
with old errors.

```
Error: Error in parsing SVG: Unquoted attribute value
Line: 1
Column: 29
Char: 6
File: input.svg
```

New errors are more informative and may solve many struggles

```
Error: SvgoParserError: input.svg:2:29: Unquoted attribute value

  1 | <svg viewBox="0 0 120 120">
> 2 |   <circle fill="#ff0000" cx=60.444444" cy="60" r="50"/>
    |                             ^
  3 | </svg>
  4 |
```


deepsweet (approved):

matheus1lva (approved):

lib/svgo/coa.js
```diff
@@ -359,12 +359,9 @@ function processSVGData(config, info, data, output, input) {
...
   const result = optimize(data, { ...config, ...info });
-  if (result.error) {
-    let message = result.error;
-    if (result.path != null) {
-      message += `\nFile: ${result.path}`;
-    }
-    throw Error(message);
+  if (result.modernError) {
+    console.error(red(result.modernError.toString()));
+    return;
```
XhmikosR, line 364: Just wondering shouldn't you use `process.exit()` here?

TrySound: Good catch

PR #1556, opened 2021-09-01
Refactor cleanupIDs

TrySound:
I still did get how id generation works but data flow is clear now at
least.

- covered with types
- got rid of traverse api

### DIFF
--- a/plugins/cleanupIDs.js
+++ b/plugins/cleanupIDs.js
@@ -1,259 +1,103 @@
 'use strict';
 
-const { traverse, traverseBreak } = require('../lib/xast.js');
-const { parseName } = require('../lib/svgo/tools.js');
+/**
+ * @typedef {import('../lib/types').XastElement} XastElement
+ */
 
+const { visitSkip } = require('../lib/xast.js');
+const { referencesProps } = require('./_collections.js');
+
+exports.type = 'visitor';
 exports.name = 'cleanupIDs';
-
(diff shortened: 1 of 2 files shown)


plugins/cleanupIDs.js
```diff
@@ -264,15 +108,190 @@ function generateID(currentID) {
...
+                id = match[1]; // href reference
+              }
+            }
+            if (id != null) {
+              let refs = referencesById.get(id);
+              if (refs == null) {
+                refs = [];
+                referencesById.set(id, refs);
+              }
+              refs.push({ element: node, name, value });
```
mrazauskas, line 231: Might be I missed something. Isn’t this the same, but slightly shorter:

```js
const refs = referencesById.get(id) || [];
if (!refs.length) {
  referencesById.set(id, refs);
}
refs.push({ element: node, name, value });
```

TrySound: Yes, though I prefer more explicit null check, length check is less obvious

mrazauskas: Sure. I was just looking around for examples. Working on a custom plugin. Just to say, Visitor API is a pleasure to use.

TrySound: Thanks. All plugins will use visitor api. full, perItem and perItemReverse will be removed in v3.

matheus1lva (approved):

PR #1558, opened 2021-09-04
Avoid encoding script and style elements content

TrySound:
Ref https://github.com/svg/svgo/issues/176

@XhmikosR @GreLI


XhmikosR:
LGTM. Just so that we are safe, do you know what this comment was about? https://github.com/svg/svgo/issues/176#issuecomment-199947766

TrySound:
I actually not sure this is necessary change
https://github.com/svg/svgo/issues/176#issuecomment-913007825

matheus1lva (approved):

GreLI:
Seems good. But I'm not sure, that things like
```html
<style>
    style::after {
         content: "</style>"
    }
   /* other styles */
</style>
```
will not be broken. Though it's unlikely will be in SVG, and in JS usually there are ways to avoid that (like `'</' + 'script>'`).

TrySound:
Closing until somebody provide a repro with visual regression

PR #1559, opened 2021-09-05
Refactor removeUnusedNS plugin

TrySound:
- covered with types
- dropped traverse utility which is replaced by visitor

### DIFF
--- a/plugins/removeUnusedNS.js
+++ b/plugins/removeUnusedNS.js
@@ -6,7 +6,8 @@ exports.active = true;
 exports.description = 'removes unused namespaces declaration';
 
 /**
- * Remove unused namespaces declaration.
+ * Remove unused namespaces declaration from svg element
+ * which are not used in elements or attributes
  *
  * @author Kir Belevich
  *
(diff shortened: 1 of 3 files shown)


matheus1lva (approved):

plugins/removeUnusedNS.js
```diff
@@ -1,82 +1,59 @@
...
+              definedNamespaces.delete(ns);
+            }
+          }
+          // check each attr for the ns-attrs
+          for (const name of Object.keys(node.attributes)) {
+            if (name.includes(':')) {
+              const [ns] = name.split(':');
+              definedNamespaces.delete(ns);
+            }
+          }
```
omgovich, line 46: The code looks good, but it feels like it doesn't have enough explanation that allows other developers to understand it faster. For example, we remove some elements from `definedNamespaces` but there is a lack of explanation why we do that. Would be better to have reference to some specs or describe the goal: Examples:

 ```js
          // Collect all namespaces from the svg element
          // (such as `xmlns:xlink="http://www.w3.org/1999/xlink"`) 
          if (node.name === 'svg' && parentNode.type === 'root') {

          // We must preserve a namespace if there are nested elements that refer
          // to the same namespace in their names
          if (node.name.includes(':')) {
          // ...

          // Dropping a namespace from the root element might break the rendering of elements
          // that use this namespace in attribute names
          for (const name of Object.keys(node.attributes)) {
          // ...
```

omgovich (approved):

plugins/removeUnusedNS.js
TrySound: Tweaked comments

PR #1560, opened 2021-09-06
fix: if params.attrs = undefined, got a error

779102260:
error -> undefined.includes()

### DIFF
--- a/plugins/removeAttrs.js
+++ b/plugins/removeAttrs.js
@@ -73,7 +73,7 @@ const DEFAULT_SEPARATOR = ':';
  * @type {import('../lib/types').Plugin<{
  *   elemSeparator?: string,
  *   preserveCurrentColor?: boolean,
- *   attrs: string | Array<string>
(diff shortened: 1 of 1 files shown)


plugins/removeAttrs.js
```diff
@@ -86,7 +86,7 @@ exports.fn = (root, params) => {
     typeof params.preserveCurrentColor == 'boolean'
       ? params.preserveCurrentColor
       : false;
-  const attrs = Array.isArray(params.attrs) ? params.attrs : [params.attrs];
+  const attrs = Array.isArray(params.attrs) ? params.attrs : (params.attrs ? [params.attrs] : []);
```
TrySound, line 89: Please fix the type too

TrySound:
Actually what's the point of this plugin without attrs field? I'd log a message the option is missing

779102260:
> Actually what's the point of this plugin without attrs field? I'd log a message the option is missing

u r right，log is better。
i got the error when 2.3.0 -> 2.5.0

Issue #1499, opened 2021-07-03
Multiple runs using prefixIds will cause the id to be prefixed multiple times.


omerman:
Would be extremely helpful if prefixByIds plugin will check if id or class already begins with the result prefix, if so, it should ignore the class/id.

Great tool, thank you for your work!

omerman:
Help?

TrySound:
Fixed in https://github.com/svg/svgo/releases/tag/v2.6.0

Issue #1406, opened 2021-03-05
Enabling both `cleanupIDs` and `prefixIds` does not work as expected


dannobytes:
**Describe the bug**
When attempting to optimize with both `cleanupIDs` and `prefixIds` plugins enabled, I expect the IDs to be minified AND prefixed with the SVG filename. However, it instead only prefixes the IDs with the filename if `cleanupIDs.minify` is disabled.

There are valid reasons why minifying the ID and then prefixing it with the filename are necessary, mainly when inlining multiple SVGs on the same page. Without unique prefixes, duplicate IDs will conflict with one another.

**To Reproduce**
1. Enable both plugins
```js
// svgo.config.js
const { extendDefaultPlugins } = require('svgo')

module.exports = {
  multipass: true,
  plugins: extendDefaultPlugins([
    'cleanupIDs',
    'prefixIds'
  ])
}
```
2. Run the optimizer on an SVG that contains IDs. Here's a sample SVG:
```html
<!-- my-file.svg -->
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 46 46">
  <g clip-path="url(#test-aaa)">
    <path d="M37.375 42.167H4.792a.958.958 0 0 1-.959-.959V8.626a1.917 1.917 0 1 0-3.833 0v33.542A3.833 3.833 0 0 0 3.833 46h33.542a1.917 1.917 0 0 0 0-3.833z" fill="#99D099"/>
    <rect fill="#008A00" height="39.531" rx="3" width="39.531" x="6.469"/>
    <g clip-path="url(#test-bbb)">
      <path d="M23.15 27.192a1.168 1.168 0 0 0 1.168 1.168h3.114a1.168 1.168 0 0 0 1.168-1.168v-4.866a.195.195 0 0 1 .195-.195h4.866a1.168 1.168 0 0 0 1.169-1.168V17.85a1.168 1.168 0 0 0-1.169-1.168h-4.866a.195.195 0 0 1-.195-.195V11.62a1.168 1.168 0 0 0-1.168-1.168h-3.114a1.168 1.168 0 0 0-1.168 1.168v4.866a.195.195 0 0 1-.195.195H18.09a1.168 1.168 0 0 0-1.168 1.168v3.114a1.168 1.168 0 0 0 1.168 1.168h4.866a.195.195 0 0 1 .195.195v4.866z" fill="#CCE7CC"/>
    </g>
  </g>
  <defs>
    <clipPath id="test-aaa">
[...]

PR #1561, opened 2021-09-09
Refactor prefixIds

TrySound:
- covered with tsdoc
- fixed prefix when specified path has no slashes (https://github.com/svg/svgo/issues/1406)
- made the plugin idempotent as requested a few times
  Now even manually running svgo a few times will not duplicate
  prefix in ids and classes ( https://github.com/svg/svgo/issues/1499)
- run each plugin test twice to see which plugin need to run many times
  ideally idempotent plugins will allow to get rid of multipass option in v3

cc @strarsis


TrySound:
Looks like the error is caused by this
https://github.com/nodejs/node/issues/40030

TrySound:
Build is fixed!

matheus1lva (approved):

PR #1567, opened 2021-09-13
 jest: ignore test files from coverage

XhmikosR:


package.json
```diff
@@ -91,6 +91,11 @@
       }
     ]
   },
+  "jest": {
+    "coveragePathIgnorePatterns": [
+      "test"
```
TrySound, line 96: I'd ignore fixtures instead

XhmikosR: But fixtures aren't included by default AFAICT.

TrySound: image

XhmikosR: Yeah, which is what this PR ignores.

TrySound: All files here are in fixtures folders. So looks like they are not igored.

XhmikosR: They are 100% ignored... https://github.com/svg/svgo/pull/1567/checks?check_run_id=3586856059

TrySound: > But fixtures aren't included by default AFAICT.

"By default" they are included. And I'm asking to ignore fixtures instead of test

XhmikosR: Personally, I see no point in including the test folder in coverage. And fixtures right now are in the test folder which is ignored.

TrySound: Tests are ignored by default. We ignoring here not tests but fixtures specifically. They are not important to cover. Test utils are better be checked for example as a way to find dead code.

XhmikosR: Doesn't seem to be possible with `coverageIgnorePattern` or at least I can't make it work to ignore only the test files instead of the folder.

TrySound: "fixtures", not tests

package.json
```diff
@@ -47,7 +47,7 @@
     "lib",
     "plugins",
     "dist",
-    "!**/**.test.js"
+    "!**/*.test.js"
```
XhmikosR, line 50: Fixed this one too since I noticed the extra asterisk here

TrySound:
Thanks

Issue #1193, opened 2019-11-19
Should SVG 1.2 be supported? Specifically filters like "feDropShadow"


gdelmas:
Hi there,

I noticed filters like "feDropShadow" are removed by the "removeUnknownsAndDefaults" plugin, as it only covers the SVG 1.1 spec.

Is there a reason for only covering 1.1? Should SVG 1.2 be supported?

https://www.w3.org/TR/SVGFilter12
https://www.w3.org/TR/SVG12/

relates to #1045

GreLI:
SVG 1.2 is a dead-end. I doubt it has notable support, or supported at all.

Now, the actual upcoming standard is SVG 2 with corresponding CSS Modules, where some related stuff like transformations and filters were transferred.

gdelmas:
thanks for clarifying. i was not aware that these filters are also defined in SVG 2.

so to rephrase my question: how can filters like "feDropShadow" that are part of the new spec be supported? is there a plan to add them.

i am asking as i was about to add the filter to https://github.com/svg/svgo/blob/master/plugins/_collections.js when i noticed it was not in the SVG 1 spec.

TrySound:
Fixed in https://github.com/svg/svgo/releases/tag/v2.6.1

Issue #1045, opened 2018-09-27
"Collapse useless groups" collapsing not useless groups


utilmind:
Hello! This issue was noticed in svgo v1.0.5.

Option "collapseGroups" (or "Collapse useless groups" in SVGOMG) removing <g> tags and filter attributes to images, that does not work -- image doesn't displayed properly.

Attached file demonstrates the problem -- just try to compress it with and without "collapseGroups" option.

[Charivniy-Vechir.zip](https://github.com/svg/svgo/files/2423862/Charivniy-Vechir.zip)

tmorehouse:
I'm running into this same issue.

```xml
  <defs>
    <filter id="a">
      <feDropShadow dx="0" dy="25" stdDeviation="25" flood-opacity="0.65" />
    </filter>
</defs>
```

is being turned into this:

```xml
<defs><filter id='a'/></defs>
```

TrySound:
Fixed in https://github.com/svg/svgo/releases/tag/v2.6.1

PR #1571, opened 2021-09-14
Add missing filter primitives to collections

TrySound:
Ref https://github.com/svg/svgo/issues/1045#issuecomment-510208768 https://github.com/svg/svgo/issues/1193

This should prevent feDropShadow from removing.


omgovich (approved):

omgovich:
Don't you also want to add some specs testing these primitives?

TrySound:
Not really. Too much work to cover whole svg spec. I'd go another way and implement spec parser or just look into mdn-data as source of truth for "collections". But later.

PR #1573, opened 2021-09-15
Fix config type

thewilkybarkid:
https://github.com/svg/svgo/pull/1565 changed the signature when using SVGO programmatically so that the config argument is required. This breaks Parcel's usage where it might be `undefined` (https://github.com/parcel-bundler/parcel/blob/6f6759e5c2140aa8d7c3be70d56432258439ff19/packages/optimizers/svgo/src/SVGOOptimizer.js#L26-L32).

I've matched the original `optimize` function, and add some basic tests to cover it.

Refs https://github.com/parcel-bundler/parcel/pull/6481#discussion_r707864127

### DIFF
--- a/lib/svgo-node.js
+++ b/lib/svgo-node.js
@@ -54,6 +54,9 @@ const loadConfig = async (configFile, cwd = process.cwd()) => {
 exports.loadConfig = loadConfig;
 
 const optimize = (input, config) => {
+  if (config == null) {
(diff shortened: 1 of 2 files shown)


TrySound:
Thanks!

TrySound:
Released in https://github.com/svg/svgo/releases/tag/v2.6.1

PR #1575, opened 2021-09-16
Refactor mergeStyles

TrySound:
- covered with tsdoc
- replace another closestByName usage with visitSkip symbol to skip
  subtree instead of skipping element by ancestor
- added one more test case

Better review with hidden whitespaces as most code just got bigger indent.


omgovich (approved):
Nice!

matheus1lva (approved):

PR #1576, opened 2021-09-17
Avoid node.parentNode in style manager

TrySound:
node.parentNode will be removed in v3 along with JSAPI class wrapper on
each node.

Style manager uses it to find inherited styles. To workaround this
I collected all parents along with all styles. This constraints style
manager to work only with initial ast which should not be a problem as
each plagin execution is isolated.

Also covered style manager tests with tsdoc.
Better review with hidden whitespaces.

### DIFF
--- a/lib/style.js
+++ b/lib/style.js
@@ -3,6 +3,7 @@
 /**
  * @typedef {import('css-tree').Rule} CsstreeRule
  * @typedef {import('./types').Specificity} Specificity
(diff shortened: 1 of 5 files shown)


matheus1lva (approved):

renatorib (approved):
I love the simplicity of Map for these things

omgovich (approved):
Looks much better without `parent.parentNode` 👍

Issue #1579, opened 2021-09-20
Add svgo.config.cjs support to support ESM projects


ai:
**Describe the bug**

I have project which we plan to moved to ESM with `"type": "module"` in `package.json`.

But if I will rename `svgo.config.js` to `svgo.config.cjs`, SVGO stopped to see the config.

TrySound:
I was just thinking how to add es modules support. At the moment config is looking only for "svgo.config.js" though you can specify config file via `--config` option.

ai:
Having `.mjs` and `.cjs` config extension is always good for some edge cases.

TrySound:
Fixed in https://github.com/svg/svgo/releases/tag/v2.7.0

PR #1583, opened 2021-09-20
Support es modules

TrySound:
Ref https://github.com/svg/svgo/issues/1579

In config of course. Projects with type:module can now use modules to
export config

```js
export default {
  plugins: []
}
```

Also added support for resolving svgo.config.mjs and svgo.config.cjs.

Moved loadConfig tests to svgo-node tests.

mjs test is skipped for now in node 10, just don't use modules there

cc @ai


TrySound:
Wow, worked!

lib/svgo-node.js
```diff
@@ -13,7 +14,25 @@ exports.extendDefaultPlugins = extendDefaultPlugins;
 exports.createContentItem = createContentItem;
 
 const importConfig = async (configFile) => {
-  const config = require(configFile);
+  let config;
+  try {
+    // dynamic import expects file url instead of path and may fail
+    // when windows path is provided
+    const { default: imported } = await import(pathToFileURL(configFile));
```
matheus1lva, line 21: Just a comment here: ESM can be imported using dynamic import without any more work to get it working?

matheus1lva (approved):

lib/svgo-node.js
TrySound: With current lts yes. But we still supported node 10 until next major release so we fallback to require in case of error.

omgovich (approved):
Looks great to me!

Smart! 👨‍🔬 I haven't considered this kind of fallback.

PR #1586, opened 2021-09-23
Replace colorette with nanocolors

TrySound:
Ref https://github.com/ai/nanocolors

Nanocolors is the new player in color libraries. And it's perfectly
packaged as all @ai packages. Colorette is already replaced even
in transitive dependencies.


matheus1lva (approved):

deepsweet (approved):

omgovich (approved):
So good!

PR #1592, opened 2021-09-29
Fix prefixIds plugin to properly handle url()s in style="..."

IlyaSkriblovsky:
First of all, thanks for the great tool!

`prefixIds` plugin currently breaks url()-links inside `style` attributes:

```javascript
optimize(
  `<g style="fill:url(#brush-id);stroke:url(#pen-id)"/>`,
  { plugins: ['prefixIds'] }
).data
```
will generate `<g style=""/>`. Seems like `prefixIds` assumes that attribute's whole value might be `url()`, but this is not the case for the `style` attribute.

This fix solves the issue by preserving all attribute's content other than #id inside url(). It also adds some more tests for the `prefixIds` plugin.


test/plugins/prefixIds.test.js
```diff
@@ -22,3 +22,51 @@ test('should extract prefix from path basename', () => {
     }).data
   ).toEqual(`<svg id="input_svg__my-id"/>`);
 });
+
+test('should replace references in attributes', () => {
```
TrySound, line 26: This test module is used only for testing filename. All tests are in `test/plugins/prefixIds.*.svg`

plugins/prefixIds.js
```diff
@@ -206,15 +206,15 @@ exports.fn = (_root, params, info) => {
...
-            if (matches != null) {
-              const value = matches[1];
-              const prefixed = prefixReference(prefix, value);
-              if (prefixed != null) {
-                node.attributes[name] = `url(${prefixed})`;
+            node.attributes[name] = node.attributes[name].replace(
+              /url\((.*?)\)/gi,
+              (match, url) => {
+                const prefixed = prefixReference(prefix, url);
+                if (prefixed == null)
```
TrySound, line 213: brackets

test/plugins/prefixIds.test.js
IlyaSkriblovsky: Oh, I didn't found the right place for plugin tests. Writing tests in *.svg files is a bit... weird :)  So I thought that prefixIds plugin has only this single test and added some more. Now I removed my redundant tests and added one for testing `style="..."` replacement as `prefixIds.11.svg`.

plugins/prefixIds.js
IlyaSkriblovsky: Added curly braces for `if`

test/plugins/prefixIds.11.svg
```diff
@@ -0,0 +1,17 @@
+<svg width="120" height="120" xmlns="http://www.w3.org/2000/svg">
```
TrySound, line 1: Add description before "==="

IlyaSkriblovsky: added one

TrySound:
Thanks!

PR #1593, opened 2021-10-01
Refactor svg stringifier

TrySound:
- rewrote prototype class with functions
- covered with tsdoc
- added a few TODOs for v3

### DIFF
--- a/lib/stringifier.js
+++ b/lib/stringifier.js
@@ -1,8 +1,70 @@
 'use strict';
 
+/**
+ * @typedef {import('./types').XastParent} XastParent
+ * @typedef {import('./types').XastRoot} XastRoot
+ * @typedef {import('./types').XastElement} XastElement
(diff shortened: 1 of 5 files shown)


lib/stringifier.js
```diff
@@ -1,8 +1,70 @@
...
+ *   textEnd: string,
+ *   indent: string,
+ *   regEntities: RegExp,
+ *   regValEntities: RegExp,
+ *   encodeEntity: (char: string) => string,
+ *   pretty: boolean,
+ *   useShortTags: boolean
+ *   eol: 'lf' | 'crlf',
+ *   finalNewline: boolean,
+ * }} Options
```
omgovich, line 54: Can't we replace that with `@typedef {Required<StringifyOptions>} Options`?

TrySound: indent types are different but maybe easier to store it in state instead

PR #1599, opened 2021-10-02
Improve css-class-list

XhmikosR:
* use `const`
* use spread
* use `for...of`
* remove two internal functions only called once

### DIFF
--- a/lib/svgo/css-class-list.js
+++ b/lib/svgo/css-class-list.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var CSSClassList = function (node) {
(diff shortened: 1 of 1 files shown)


TrySound:
This code will be killed in v3. No need to touch it at all.

XhmikosR:
Is that documented somewhere? I wasn't aware of that, nor do I know when v3 will be out.

TrySound:
Whole JSAPI class is gonna be removed. SVGO plugins will work with plain object ast.

XhmikosR:
Regardless, your call. If v3 is close, feel free to close this.

PR #1600, opened 2021-10-02
Improve css-tools

XhmikosR:
* use const
* use arrow functions
* use `for...of`
* stop using bitwise operators
* group exports

The non-whitespace diff is better: https://github.com/svg/svgo/pull/1600/files?diff=unified&w=1


TrySound:
All css-tools usages was removed in https://github.com/svg/svgo/pull/1601. No need to touch it as well.

PR #1601, opened 2021-10-03
Refactor inlineStyles

TrySound:
This is a big one

- got rid from another closestByName usage
- delegated removing empty defs elements to removeEmptyContainers plugin
- got rid from all css-tools usages (most inlineStyles code was there
  for some reason)
- combined a few loops
- fixed useMqs option (I would remove it in v3 for simplicity as it
  seems nobody use it)

cc @strarsis

### DIFF
--- a/plugins/inlineStyles.js
+++ b/plugins/inlineStyles.js
@@ -1,23 +1,42 @@
 'use strict';
 
+/**
+ * @typedef {import('../lib/types').Specificity} Specificity
+ * @typedef {import('../lib/types').XastElement} XastElement
+ * @typedef {import('../lib/types').XastParent} XastParent
+ */
+
 const csstree = require('css-tree');
-const { querySelectorAll, closestByName } = require('../lib/xast.js');
-const cssTools = require('../lib/css-tools');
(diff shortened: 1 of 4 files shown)


plugins/inlineStyles.js
```diff
@@ -1,24 +1,43 @@
 'use strict';
 
-const csstree = require('css-tree');
-const { querySelectorAll, closestByName } = require('../lib/xast.js');
-const cssTools = require('../lib/css-tools');
+/**
+ * @typedef {import('../lib/types').Specificity} Specificity
+ * @typedef {import('../lib/types').XastElement} XastElement
```
matheus1lva, line 5: i'm really starting to see the benefits of jsdocs inside a ts project. You can define your types outside, as plain ts files and use on js files :D

strarsis:
Ah yes, in the good old times (1, 2 years ago) svgo didn't use any TS features and I tried to isolate the CSS-related stuff into a separate object.
LGTM, lots of code is deduplicated and simplified.

matheus1lva (approved):

PR #1602, opened 2021-10-03
Migrate to yarn

TrySound:
Yarn supports windows with custom shell, has better tools for
dependencies upgrade and more readable lockfile.

Migrated with preserving lockfile data.


TrySound:
Oops, looks like this have to wait til v3

TrySound:
It works!

PR #1624, opened 2021-12-06
Plugin definitive guide

TrySound:
This is just initial thoughts. Feel free to suggest any changes.

Here are described only the new visitor plugins api. Old "full", "perItem", "perItemReverse" will be removed in v3. Old JSAPI class on every node will be also removed so nodes could be plain objects.


glennosss:
Hi TrySound - I'm not sure I'd be in such a rush to remove 'full'. By removing 'full' you'll just get people who were using 'full' (like me) using the 'visitor' plugin, but never returning a visitor. In order to replace the 'ast' like you could with a plugin type of 'full', you'll get people (like me) replacing all the properties on the old ast object. Completely possible to use the 'visitor' plugin type to do these things, just messier.

Example usecase: https://github.com/glennosss/svgo-autocrop/blob/master/lib/AutocropUtils.js

TrySound:
Hi. I understand your usecase. Though the point of new plugin api is to cover all possible use cases and remove the necessity to choose.

Your case is easy to solve as well. Workers may accept root children or even root svg element. And you just need to assign new children to root.children. This fits into idea that ast is just mutable object.

glennosss:
Thanks TrySound - updated to 'visitor' plugin - https://github.com/glennosss/svgo-autocrop/blob/master/index.js

PR #1625, opened 2021-12-10
Adding convertAllShapeToPath plugin 

BaseMax:
Hi there,

We tried to create web font icon from SVGs and for that purpose what is matter is to convert all of shapes to Paths.

So I going to add a new plugin for someone who want to convert all of the shapes into paths. (without consider about optimization and long SVG)

I going to check bugs and problems.
Let me know if you find any kind of problems.


BaseMax:
A known- problem:

## input
input

## output (see border)
output

TrySound:
Hi, thanks for this. Though I meant you can have custom plugin in your config or published as a separate package. SVGO builtin plugin are only meant for optimisation.

PR #1627, opened 2021-12-16
Fix typo in regression.js

eltociear:
becase -> because


SethFalco:
Thanks for spotting this and opening a PR! Closing as it has actually been addressed in a recent commit, though.

PR #1629, opened 2021-12-21
Respect finalNewline setting in pretty mode

korki43:
Pretty mode will add a newline to the end of each 'element'. This means that a newline at the end is always created when setting `js2svg.pretty: true`. The setting `js2svg.finalNewLine` will be completely ignored because there already is a newline.

This change removes this unexpected behavior by stripping the last newline if `pretty: true` and `finalNewLine: false` are set. 

Ref: https://github.com/twbs/icons/pull/1148#discussion_r773155861

### DIFF
--- a/lib/stringifier.js
+++ b/lib/stringifier.js
@@ -124,14 +124,20 @@ const stringifySvg = (data, userOptions = {}) => {
     config.textEnd += eol;
   }
   let svg = stringifyNode(data, config, state);
-  if (config.finalNewline && svg.length > 0 && svg[svg.length - 1] !== eol) {
+
+  if (config.finalNewline && svg.length > 0 && svg[svg.length - 1] !== '\n') {
     svg += eol;
   } else if (
-    config.pretty &&
     !config.finalNewline &&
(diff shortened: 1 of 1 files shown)


lib/stringifier.js
```diff
@@ -124,8 +124,14 @@ const stringifySvg = (data, userOptions = {}) => {
     config.textEnd += eol;
   }
   let svg = stringifyNode(data, config, state);
-  if (config.finalNewline && svg.length > 0 && svg[svg.length - 1] !== '\n') {
+  if (config.finalNewline && svg.length > 0 && svg[svg.length - 1] !== eol) {
```
TrySound, line 127: This will never work with crlf as there are 2 characters.

korki43: You're right, I forgot about that. Will adjust it.

korki43: Should be fixed by 95b0122caef622a83f24a71af3a2c3018bffc7cc.

TrySound:
This logic looks quite complex. I would keep it as is and use .trim twbs/icons repo. Thanks anyway!

PR #1632, opened 2021-12-28
[Documentation] Update how-it-works docs to remove .yml reference

michaeljaltamirano:
I am working on a project that updates `svgo` from v1 to v2 and I noticed that `.yml` files were deprecated in https://github.com/svg/svgo/pull/1328.

This PR removes a floating reference in the repo to `.yml` configurations, favoring just pointing to the newer docs. It's pretty minor; the only other possible change I can think of is making the information that only `.js` files are valid config a bit clearer in the README. Feel free to edit as you please!

### DIFF
--- a/docs/how-it-works/en.md
+++ b/docs/how-it-works/en.md
@@ -18,25 +18,7 @@ Plugins can remove and modify SVG elements, collapse contents, move attributes a
 
 ## How it works
 ### 1. config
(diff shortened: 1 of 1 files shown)


TrySound:
Hi! how-it-works is actually very outdated. I'm writing new docs. Any suggestions are welcome.
https://github.com/svg/svgo/pull/1624

TrySound:
Removed how-it-works

PR #1633, opened 2021-12-28
Correct typo in comment in removeDimensions plugin

rogierslag:
This addresses a very minor typo, but came across this while reading up on the plugin itself.


TrySound:
Thanks!

PR #1635, opened 2022-01-06
Refactor convertStyleToAttrs

TrySound:
Did not rewrite much, just migrated to tsdoc and visitor plugin api.
This plugin will be obsolete eventually.

### DIFF
--- a/plugins/convertStyleToAttrs.js
+++ b/plugins/convertStyleToAttrs.js
@@ -1,54 +1,56 @@
 'use strict';
 
+const { attrsGroups } = require('./_collections');
+
+exports.type = 'visitor';
 exports.name = 'convertStyleToAttrs';
-
-exports.type = 'perItem';
-
 exports.active = false;
-
 exports.description = 'converts style to attributes';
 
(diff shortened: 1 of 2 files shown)


plugins/convertStyleToAttrs.js
```diff
@@ -63,70 +65,77 @@ var stylingProps = require('./_collections').attrsGroups.presentation,
...
-          ? ''
-          : match[0] == '\\' && /[-g-z]/i.test(match[1])
-          ? match[1]
-          : match;
-      }
-    );
+          if (styles.length) {
+            styles = styles.filter(function (style) {
+              if (style[0]) {
+                var prop = style[0].toLowerCase(),
```
matheus1lva, line 110: Just curiosity here. Why the need to use var?

matheus1lva (approved):

plugins/convertStyleToAttrs.js
TrySound: I just kept old code here

> Note: review with hidden whitespaces

PR #1645, opened 2022-02-04
More helpful preset warnings

liamcmitchell-sc:
Preset documentation and warning [here](https://github.com/svg/svgo/commit/cb7e9be623b6e2fbbfcb9b67c4c85131e1477925) got me confused. I think these changes should reduce others confusion.

* warn **any** override that is not in the default list
* use actual plugin names in the warning message

## How I got here
For context, I'm migrating from SVGO v1 and I came straight to the config docs here (https://github.com/svg/svgo#configuration) when I realised there were config changes.

I read the `preset-default` docs and saw the `overrides` option, quickly copy pasted my previous config over and changed to the new API:
```js
name: 'preset-default',
params: {
  overrides: {
    removeXMLNS: true,
    removeTitle: true,
    removeViewBox: false,
    addAttributesToSVGElement: {/* ... */},
  },
},
```
And see the new warning:
```
You are trying to enable removeXMLNS which is not part of preset.
Try to put it before or after preset, for example

plugins: [
  {
    name: 'preset-default',
  },
  'cleanupListOfValues'
]
```
My first thought is that `removeXMLNS` was no longer part of the built-in plugins. The warning suggests using `cleanupListOfValues`, I thought maybe that functionality was moved into this new plugin.

I go back to the documentation and see no, the plugin is still there, it just needs to be enabled in a different way now.

### DIFF
--- a/lib/svgo/plugins.js
+++ b/lib/svgo/plugins.js
@@ -87,8 +87,9 @@ const createPreset = ({ name, plugins }) => {
         globalOverrides.floatPrecision = floatPrecision;
       }
       if (overrides) {
+        const pluginNames = plugins.map(({ name }) => name);
         for (const pluginName of Object.keys(overrides)) {
-          if (!plugins.find(({name}) => name === pluginName)) {
+          if (!pluginNames.includes(pluginName)) {
             console.warn(
               `You are trying to configure ${pluginName} which is not part of ${name}.\n` +
                 `Try to put it before or after, for example\n\n` +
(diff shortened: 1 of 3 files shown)


lib/svgo/plugins.js
```diff
@@ -87,16 +87,16 @@ const createPreset = ({ name, plugins }) => {
         globalOverrides.floatPrecision = floatPrecision;
       }
       if (overrides) {
-        for (const [pluginName, override] of Object.entries(overrides)) {
-          if (override === true) {
+        for (const pluginName of Object.keys(overrides)) {
+          if (!plugins.find(({name}) => name === pluginName)) {
```
TrySound, line 91: Better create array with builtin plugin names and use includes

liamcmitchell-sc: Done

TrySound:
Good improvement. Thanks!

PR #976, opened 2018-06-06
inline definitions

marinewater:
this inlines definitions
e.g. turns
```
<svg id="test01" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 100 100">
    <defs>
        <rect width="100" height="100" id="rect1"/>
    </defs>
    <use xlink:href="#rect1"></use>
</svg>
```
into
```
<svg id="test01" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 100 100">
    <rect width="100" height="100" id="rect1"/>
</svg>
```

By default, only definitions which are used once are inline. If you set the *onlyUnique* parameter to `false`, all definitions will be inlined.
Should solve #563

PR #1655, opened 2022-02-21
[new plugin] reimplement inlineDefs

TrySound:
The code is taken from https://github.com/svg/svgo/pull/976, refactored
with new api, covered types and simplified.

Plugin has no dependencies so can be used without changing.

```
const inlineDefs = require('./inlineDefs.js');

module.exports = {
  plugins: [
    'preset-default',
    inlineDefs
  ]
};
```

### DIFF
--- a/plugins/inlineDefs.js
+++ b/plugins/inlineDefs.js
@@ -1,0 +1,166 @@
+'use strict';
+
+/**
+ * @typedef {import('../lib/types').XastParent} XastParent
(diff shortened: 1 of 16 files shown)


itayganor:
Will this ever happen?

strarsis:
@TrySound, @itayganor: Yes, dereference `<use>` plugin: https://github.com/svg/svgo/pull/1279

PR #1658, opened 2022-03-03
fix: begin's attribute incorrectly modification

Kreeg:
Hello!
The problem is with `<animate>'s` `begin` clearing. For example if both `id="2"` and `<animate begin="2.5s">` are present in svg, it changes to `id="a"` and `<animate begin="a.5s">` which is incorrect value.

This bug was originally found here https://github.com/svg-sprite/svg-sprite/issues/526

### DIFF
--- a/plugins/cleanupIds.js
+++ b/plugins/cleanupIds.js
@@ -10,9 +10,9 @@ const { referencesProps } = require('./_collections.js');
 exports.name = 'cleanupIds';
 exports.description = 'removes unused IDs and minifies used';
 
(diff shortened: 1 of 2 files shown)


XhmikosR:
@Kreeg could you rebase your branch? @TrySound would it be possible to include this fix in the next patch release?

Kreeg:
@XhmikosR I updated the branch. @TrySound please take a look.

TrySound:
Thanks!

TrySound:
Published in https://github.com/svg/svgo/releases/tag/v3.0.1

PR #1660, opened 2022-03-06
removeXMLSpace plugin: remove xml:space attribute for inline svg usage

IngyuTae:
`xml:space` attribute has been deprecated according to MDN.

SVG supports the built-in XML xml:space attribute to handle whitespace characters inside elements. Child elements inside an element may also have an xml:space attribute that overrides the parent's one.

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xml:space#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xml:space


mogrady88:
Is this something that will be implemented in a near release?

alvinometric:
@TrySound Any plans to merge this?

SethFalco:
We have a dedicated plugin for removing deprecated attributes in `main`, so if you're removing it because it's deprecated, you can use that instead. It will be part of the next release.

It will be enabled by default, but will only remove safe attributes. Deprecated attributes that may impact rendering will only be removed if the `removeAny` parameter is passed.

See:

* https://github.com/svg/svgo/pull/1869

It could be worth having a safe plugin that removes `xml:space="preserve"` when it's safe to do (i.e. the text/cdata doesn't contain content that would be impacted by the attribute).

PR #1662, opened 2022-03-15
Add `funding` key to package.json

mondeja:
This will add a notification about how to fund the package to the npmjs.com page (example from webpack):

image

See [package-json#funding](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#funding).


TrySound:
Thanks!

Issue #1050, opened 2018-10-09
Here is a hack for svgo to work in a browser


fr33z00:
Hi all.
Like some others, I needed svgo to work in a browser, so, here is a hack I used.
In lib/svgo/config.js, add the following code under the 'use strict' (line 1): 
```
var all_plugins = {
  cleanupAttrs: require('../../plugins/cleanupAttrs.js'),
  removeDoctype: require('../../plugins/removeDoctype.js'),
  removeXMLProcInst: require('../../plugins/removeXMLProcInst.js'),
  removeComments: require('../../plugins/removeComments.js'),
  removeMetadata: require('../../plugins/removeMetadata.js'),
  removeTitle: require('../../plugins/removeTitle.js'),
  removeDesc: require('../../plugins/removeDesc.js'),
  removeUselessDefs: require('../../plugins/removeUselessDefs.js'),
  removeEditorsNSData: require('../../plugins/removeEditorsNSData.js'),
  removeEmptyAttrs: require('../../plugins/removeEmptyAttrs.js'),
  removeHiddenElems: require('../../plugins/removeHiddenElems.js'),
  removeEmptyText: require('../../plugins/removeEmptyText.js'),
  removeEmptyContainers: require('../../plugins/removeEmptyContainers.js'),
  removeViewBox: require('../../plugins/removeViewBox.js'),
  cleanupEnableBackground: require('../../plugins/cleanupEnableBackground.js'),
  convertStyleToAttrs: require('../../plugins/convertStyleToAttrs.js'),
  convertColors: require('../../plugins/convertColors.js'),
  convertPathData: require('../../plugins/convertPathData.js'),
  convertTransform: require('../../plugins/convertTransform.js'),
  removeUnknownsAndDefaults: require('../../plugins/removeUnknownsAndDefaults.js'),
  removeNonInheritableGroupAttrs: require('../../plugins/removeNonInheritableGroupAttrs.js'),
  removeUselessStrokeAndFill: require('../../plugins/removeUselessStrokeAndFill.js'),
  removeUnusedNS: require('../../plugins/removeUnusedNS.js'),
  cleanupIDs: require('../../plugins/cleanupIDs.js'),
  cleanupNumericValues: require('../../plugins/cleanupNumericValues.js'),
  moveElemsAttrsToGroup: require('../../plugins/moveElemsAttrsToGroup.js'),
  moveGroupAttrsToElems: require('../../plugins/moveGroupAttrsToElems.js'),
[...]

PR #1671, opened 2022-04-28
Add documentation & sample usage for `svgo.browser.js`

dr-js:
Reworded based on https://github.com/svg/svgo/issues/1050#issuecomment-1111706921


dr-js:
@Glandos I think to use the `es module` file with webpack or rollup would need one sample,
and to import this file from TypeScript with correct type would be another sample config,
both can be complex depending on how everyone's repo/IDE is configured.

For your use case in TypeScript the types from [`@types/svgo`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/svgo) is for the main `svgo` code (usage like `import .. from 'svgo'`).
'svgo/dist/svgo.browser.js' is plain bundled JavaScript and without corresponding `.d.ts`, so some type patch from the user side should be required (maybe borrow types from `@types/svgo` can work).

Glandos:
That's what I did for now:
```ts
declare module 'svgo/dist/svgo.browser.js' {
    import svgo from 'svgo'
    export const optimize: typeof svgo.optimize
}
```

PR #1675, opened 2022-05-13
Allow --input to be a single file when --output is a folder

burntcustard:
The README currently has example CLI usage, where you can either:
a) process one or more single input files into the same number of output files:  
`svgo one.svg two.svg -o one.min.svg two.min.svg`  
or  
b) process a directory of files, into an output directory:  
`svgo -f ./path/to/folder/with/svg/files -o ./path/to/folder/with/svg/output`

I had hoped that I'd be able to do an in-between, with a single file as the input and a directory as the --output:  
`svgo one.svg -o ./path/to/folder/with/svg/output`

That would let me use chokidar in my project's package.json, to watch any SVGs that are updated, and re-run SVGO on them:  
`"svg:watch": "chokidar src/assets/svg/*.svg -c 'svgo --input {path} --output dist/assets/svg'"`

Unfortunately, I ran into an issue where if any layer of the output directory did not exist before SVGO ran, it would have unexpected results, like `Error: ENOENT: no such file or directory`, or creating a weird `assets/svg` file, that should have been `assets/svg/mySvg.svg`.

My fix replaces the check to see if a single --output string matches a pre-existing _directory_, with a check that it's simply a directory string (rather than a filename string with a file extension). That way, if you intend to put files into an output directory, but the directory doesn't yet exist, the surrounding code in coa.js successfully creates both the intended directory and output file inside it.

If this PR is accepted, it might be helpful to put a relevant single-input folder-output example in the CLI usage section of the README, to show that this can now be done?

### DIFF
--- a/lib/svgo/coa.js
+++ b/lib/svgo/coa.js
@@ -215,7 +215,7 @@ async function action(args, opts, command) {
   // --output
   if (output) {
     if (input.length && input[0] != '-') {
-      if (output.length == 1 && checkIsDir(output[0])) {
+      if (output.length == 1 && !path.extname(output[0])) {
         var dir = output[0];
         for (var i = 0; i < input.length; i++) {
           output[i] = checkIsDir(input[i])
--- a/lib/svgo/coa.js
+++ b/lib/svgo/coa.js
@@ -215,7 +215,7 @@ async function action(args, opts, command) {
   // --output
   if (output) {
     if (input.length && input[0] != '-') {
-      if (output.length == 1 && !path.extname(output[0])) {
+      if (output.length == 1 && output[0] != '-' && !path.extname(output[0])) {
         var dir = output[0];
         for (var i = 0; i < input.length; i++) {
           output[i] = checkIsDir(input[i])


burntcustard:
Apologies for not catching the output-as-stream test failure. I've put in an extra check to revert to previous behaviour when output[0] is `-`, and it passes locally.

I'm not really a fan of the non-strict comparisons being used in the line I'm changing, but it is at least consistent with the rest of the code, and doesn't negatively effect anything AFAIK.

lib/svgo/coa.js
```diff
@@ -215,7 +215,7 @@ async function action(args, opts, command) {
   // --output
   if (output) {
     if (input.length && input[0] != '-') {
-      if (output.length == 1 && checkIsDir(output[0])) {
+      if (output.length == 1 && output[0] != '-' && !path.extname(output[0])) {
```
SethFalco, line 218: I disagree with this approach because it leads to unexpected behavior in other cases, for example:

```sh
svgo original.svg -o minified
```

I would expect this to create a file, not a directory, called `minified`.

Imo, a better solution would be to update `checkIsDir` to return `true` when the path ends with a trailing `path.sep`?

```js
/**
 * Synchronously check if path is a directory. Tolerant to errors like ENOENT.
 *
 * @param {string} filePath
 */
export function checkIsDir(filePath) {
  try {
    return fs.lstatSync(filePath).isDirectory();
  } catch (e) {
    return filePath.endsWith(path.sep);
  }
}
```

```sh
# output: minified
svgo original.svg -o minified

# output: minified.svg
svgo original.svg -o minified.svg

# output: dist/minified.svg
svgo original.svg -o dist/minified.svg

# output: dist/original.svg
svgo original.svg -o dist/
```

Now, when the output path ends with an explicit `/` (or `\` on Windows) we know you mean a directory. Backward compatibility is less of a concern because SVGO previously crashed in this scenario anyway. (When the path ends with a trailing slash.)

Does this sound like a suitable solution to you?

SethFalco:
This has been handled with the approach proposed in my comment. It'll be released in the next version of SVGO. 👍🏽

PR #1679, opened 2022-06-15
fix SVGO ignoring config for `datauri` and `path`

dlueth:
The currently implemented code for multipass leads SVGO to completely ignore config options for `datauri` and `path` in most cases so that it will never base64-encode to datauri.


dlueth:
@TrySound Would you mind having a look eventually?

TrySound:
Fixed in https://github.com/svg/svgo/releases/tag/v3.0.0

PR #1688, opened 2022-07-07
Remove extension from prefixIds

### LINKED ISSUE #1687: prefixIds should probably not append the extension in the prefix.

`prefixIds` plugin works well, it will transform `id="path-1"` and `class="path-1"` to `id="test_svg__path-1"` and `class="test_svg__path-1"` for a svg named `test.svg`

Since `svgo` only process svg, shouldn't it just append the name of the file and not the name + extension ?
For exemple:
transform `id="path-1"` and `class="path-1"` to `id="test__path-1"` and `class="test__path-1"` for a svg named `test.svg`

mrleblanc101:
Fix #1687

### DIFF
--- a/plugins/prefixIds.js
+++ b/plugins/prefixIds.js
@@ -21,7 +21,7 @@ const getBasename = (path) => {
   // extract everything after latest slash or backslash
   const matched = path.match(/[/\\]?([^/\\]+)$/);
   if (matched) {
-    return matched[1];
+    return matched[1].indexOf('.') ? matched[1].split('.')[0] : matched[1];
   }
   return '';
 };


mrleblanc101:
Tested with npm link and it work great 👍

mrleblanc101:
Why not add the extension only for non-SVG file ? That would make much more sense to me

SethFalco:
For the same reason, for example, files named `original` and `original.svg` would conflict. We want the default behavior to avoid these kinds of things.

It may be an edge-case, but better to achieve the goal of the plugin properly than to potentially produce the same issue that most users enable this plugin to solve.

If you're sure you won't encounter this, you can pass the `prefix` param, and configure it to suit you. However, it's not SVGO's place to assume the directory structure or file naming conventions of users, hence my thoughts on the matter.

PR #1692, opened 2022-08-23
Add option to convert hex colors to a common case

JayLeininger:
This adds an option that allows the convertCase parameter to be passed with 'lower' or 'upper' in order to covert all hex colors to a common case. This allows for consistent use of case in converted SVGs. The parameter defaults to false so no changes to the case is applied.


dilyanpalauzov:
This would allow for better compression, so the default shall be: do convert all hexcodes to a common case.

dilyanpalauzov:
This would allow for better compression, e.g. when the same color appears several times, but with different caSEs, or different colour share the same prefix (the same first hex-digits).

SethFalco (approved):
Hey! Thanks for submitting the PR. I've rebased it for you and made the following changes:

* We can't always lower case the color, we have to make sure it isn't a URL reference first. i.e. `url(#Aa)`
* Enable this by default with `lower` as the default value, it's a good optimization.
* Update the type to explicitly allow only `false`, `'lower'`, and `'upper'`, rather than any `boolean`/`string`.
* Add documentation for the new param.

On the side, I also fixed a bug with the `removeAttrs` `preserveCurrentColor` param, which previously checked for `currentColor` case sensitively when it should've been case-insensitive.

I have tested this with the default compression settings of NGINX, and it does improve compression. I used one of our test SVGs to measure the difference, which went from 4.03 KB to 4.01 KB transferred.

PR #1707, opened 2022-11-08
[Docs] Added Remark plugin as another way to use SVGO

alvinometric:
Hello,
Thank you for the great work! I made a remark plugin to inline SVG image files within Markdown and optimise them with SVGO. This PR adds a link to the plugin in `Other Ways to Use SVGO` in the Readme.


ADTC:
@alvinometric please change to the new tabular structure. Thanks.

alvinometric:
@ADTC Done 😃

PR #1712, opened 2022-11-17
fix: types in tsconfig

daniil4udo:


TrySound:
Hi, what are you trying to fix here? It's not clear without description.

daniil4udo:
@TrySound  here

image

daniil4udo:
I see you have removed `csso` from types, ignore my PR the 👍🏻

PR #1716, opened 2022-11-27
Speed up strongRound by avoiding string casting

marvinhagemeister:
I'm currently profiling my build pipeline and noticed that the `strongRound` function showed up a couple of times, followed immediately with a bit of GC cleanup shortly after. We can speed up that function by avoiding the string casting caused by `Number.prototype.toFixed()`.

In my project's build pipeline with lots of svg icons this saves about 1.4s in total.

js-tools-strongRound

### DIFF
--- a/plugins/convertPathData.js
+++ b/plugins/convertPathData.js
@@ -35,10 +35,6 @@ let arcThreshold;
  * @type {number}
  */
 let arcTolerance;
-/**
- * @type {number[]}
- */
-const powCache = [1, 10, 100, 1000, 10000, 100000, 1000000];
 
 /**
  * @typedef {{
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
Alright CI is now passing here too. Didn't account for the `powCache` needing to expand to `floatPrecision + 1` instead of just `floatPrecision` due to the narrowing logic in `strongRound`.

marvinhagemeister:
Totally forgot about the `**` operator. I think that's a cleaner approach compared to caching it. Confirmed with a local benchmark that it's about the same.

plugins/convertPathData.js
```diff
@@ -951,6 +951,16 @@ function getIntersection(coords) {
...
 }
 
+/**
+ * Does the same as `Number.prototype.toFixed` but without casting
+ * the return value to a string.
+ * @type {(num: number, precision: number) => number}
+ */
+function toFixed(num, precision) {
+  const pow = 10 ** precision;
+  return Math.round(num * pow) / pow;
```
bnoordhuis, line 961: Hi @marvinhagemeister, I was reading your blog post and I noticed this potential issue:

`precision` can go up to 19 if I read this file right (but seems to default to 3?) so you can lose some... ah, er... precision when multiplying it with another large number:
```js
Math.round(Number.MAX_SAFE_INTEGER * 10**19) / 10**19  // off by one
Math.round(4294967295 * 10**19) / 10**19  // 4294967295.0000005
Math.round(16777215 * 10**19) / 10**19  // 16777214.999999998
```
I assume that's okay here because `num` won't be that big but I can't say for sure from context.

marvinhagemeister: Good catch, yeah with higher precision values this can become an issue. The `precision` setting is typically used to _reduce_ precision to make the SVG smaller. So far I've only seen precision settings of `2` or `3` in the wild.

PR #1717, opened 2022-11-27
Speed up stringifyNumber

marvinhagemeister:
I'm currently profiling my build setup and noticed that `stringifyNumber` pops up here and there. Instead of going with the double regex replace approach, we'll go with the faster approach in `removeLeadingZero`.

In total the changes in this PR speed up the build by about `0.9s` in my project.

stringifyNumber

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const { removeLeadingZero } = require('./svgo/tools');
+
 /**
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
Alright, tests are now passing. Apologies, was only checking the `test-browser` and `test-regression` commands before making a PR 🙈

TrySound:
Hey, no worries. Thanks for looking into this!

PR #1736, opened 2023-01-09
Include all configuration in the <style /> element

john-neptune:
This fixes a bug in the prefixIds plugin.  It assumed that the `<style />` element will have a single text fragment underneath.

So this works:
```
<style>
<!-- Leading comment ignored -->
.class1 {}
...
.classN {}
</style>
```

The above parses so that node.children[0] is a "text" node with all the CSS classes where "node" is the style element.

And this fails:
```
<style>
<!-- Ignored comment -->
.class1 {}
<!-- comment -->
.class2 {}
</style>
```

This fails because only node.children[0] is processed and node.children looks something like:
[
.class1 (type = "text")
comment (type = "comment")
.class2 (type = "text")
]

### DIFF
--- a/plugins/prefixIds.js
+++ b/plugins/prefixIds.js
@@ -17,7 +17,7 @@ exports.description = 'prefix IDs';
  */
 const getBasename = (path) => {
   // extract everything after latest slash or backslash
-  const matched = path.match(/[/\\]?([^/\\]+)$/);
+  const matched = /[/\\]?([^/\\]+)$/.exec(path);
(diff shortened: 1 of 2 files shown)


john-neptune:
Any issues with this PR?

john-neptune:
Any issues with this PR?

SethFalco:
Hey! Thanks for reporting the issue and pull request.
I've rebased this for you and added a test. This will be included in the next release.

PR #1738, opened 2023-01-14
Type fix for custom / external plugins

Hecatron:
I've included a small type fix for using custom / external plugins with svgo
If you're trying to define a plugin which takes parameters it tends to complain
as the type checking assumes the parameter has to be void

The type warning also shows up with the example already provided within the Readme without this fix
Since there's no way to know the type of parameters being passed through a custom plugin and the CustomPlugin type isn't exported, it's why I've used the any type here.

Currently I'm having to use `// @ts-ignore` above where the plugin is defined in the config
or add 'as any' to the end of the line

### DIFF
--- a/lib/svgo.d.ts
+++ b/lib/svgo.d.ts
@@ -6,7 +6,8 @@ import type {
 
 type CustomPlugin = {
   name: string;
-  fn: PluginFn<void>;
+  fn: PluginFn<any>;
+  params: any;
 };
 
 type PluginConfig =


SethFalco (changes requested):
Hey! Thanks for raising the issue and creating a pull request.

I think a better solution would be to make the following changes:

```js
type CustomPlugin<T = any> = {
  name: string;
  fn: Plugin<T>;
  params?: T
};
```

```js
export type Plugin<Params> = (
  root: XastRoot,
  params: Params,
  info: PluginInfo,
) => Visitor | null | void;
```

The benefits:

* `fn` does not require returns, we don't do this internally either. Drops the need to `return null;`.
* If you choose to populate the generic, you can have good code completion as normal.
  * Allows omitting the `params`, like when the custom plugin is being defined.
  * Offers code completion when populating params.

Do you see any problems with this approach?

PR #1742, opened 2023-01-27
Add support for image/jpeg mimetype

abejfehr:
In practice, this plugin does not work to remove embedded jpg images.

On further inspection, it's because the plugin is looking for `image/jpg`, which I'm not even sure is a valid mimetype. Inkscape uses `image/jpeg` as the mimetype for embedded JPG images.

### DIFF
--- a/plugins/removeRasterImages.js
+++ b/plugins/removeRasterImages.js
@@ -21,7 +21,7 @@ exports.fn = () => {
         if (
           node.name === 'image' &&
           node.attributes['xlink:href'] != null &&
-          /(\.|image\/)(jpg|png|gif)/.test(node.attributes['xlink:href'])
+          /(\.|image\/)(jpe?g|png|gif)/.test(node.attributes['xlink:href'])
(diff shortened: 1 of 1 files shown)


plugins/removeRasterImages.js
```diff
@@ -21,7 +21,7 @@ exports.fn = () => {
         if (
           node.name === 'image' &&
           node.attributes['xlink:href'] != null &&
-          /(\.|image\/)(jpg|png|gif)/.test(node.attributes['xlink:href'])
+          /(\.|image\/)(jpg|jpeg|png|gif)/.test(node.attributes['xlink:href'])
```
SethFalco, line 24:
```suggestion
          /(\.|image\/)(jpe?g|png|gif)/.test(node.attributes['xlink:href'])
```

Very minor suggestion, but figured it just keeps things tidier.
Thanks for spotting this, will happily approve after.

SethFalco (approved):

PR #1745, opened 2023-01-27
Correct documentation of cleanupIDs

GeorgeTaveras1231:
Basing this documentation clean up on the fact that referencing the plugin via `cleanupIds` doesn't work but `cleanupIDs` does work

### This doesn't work
```js
svgo.optimize(source, {
    path: dirname,
    plugins: [
      {
        name: 'preset-default',
        params: {
          overrides: {
            cleanupIds: {
              minify: false,
            },
          },
        },
      },
    ],
  });
```

### This works
```js
svgo.optimize(source, {
    path: dirname,
    plugins: [
      {
        name: 'preset-default',
        params: {
          overrides: {
            cleanupIDs: {
              minify: false,
            },
          },
        },
      },
    ],
  });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Default preset includes the following list of plugins:
 - mergeStyles
 - inlineStyles
 - minifyStyles
-- cleanupIds
+- cleanupIDs
 - removeUselessDefs
 - cleanupNumericValues
 - convertColors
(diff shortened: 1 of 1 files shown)


ADTC:
Should I integrate this fix in my PR @TrySound @deepsweet ?

ADTC:
@GeorgeTaveras1231 are you sure about this btw? Searching the code, I can only find `cleanupIds`.

GeorgeTaveras1231:
@ADTC I've looked further into it and it seems I ran into this problem because I was using an older version of SVGO. I was using v2.8.0, and the documentation in that git tag seems to be correct. Sorry for that miss 😅 . I will go ahead and close this.

PR #1747, opened 2023-02-06
Don't remove an empty container if parent is "switch"

### LINKED ISSUE #1410: "Viewer does not support full SVG 1.1" added to image during optimization
**Describe the bug**
When I run an SVG 1.1 document through svgo (version 2.2.1 installed via `yarn global`), I get the text "Viewer does not support full SVG 1.1" added to the bottom of the image. I don't see anything else wrong with the image. Checking the output SVG source shows that the text is added by svgo, not the viewer.

**To Reproduce**

1. Use diagrams.net / draw.io to create a simple diagram.
2. Export the image to SVG with default settings.
3. Run `svgo my.svg -o my2.svg`
4. View the resulting image.

**Expected behavior**

Assuming that the warning is legitimate, I would expect it to either appear on stdout or else to fail processing and be on stderr. Showing up silently on the output image was unexpected.

**Screenshots**

image

 - SVGO Version - 2.2.1
 - NodeJs Version - 15.10.0
 - OS: MacOS Catalina 10.15.7

vincentbernat:
Fix #1410

### DIFF
--- a/plugins/removeEmptyContainers.js
+++ b/plugins/removeEmptyContainers.js
@@ -49,6 +49,9 @@ exports.fn = () => {
         if (node.name === 'mask' && node.attributes.id != null) {
           return;
         }
+        if (parentNode.type === 'element' && parentNode.name === 'switch') {
+          return;
+        }
         detachNodeFromParent(node, parentNode);
       },
     },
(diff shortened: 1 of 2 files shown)


SethFalco (changes requested):
The fix looks fine, thanks for investigating the issue.

If you have time, could you please:

* Rebase with `main` (to run CI)
* Move the new test into the directory with plugin-specific tests

Please let me know if you don't have time, and I can take it over (preserving authorship) to get it merged. 👍🏽

PR #1750, opened 2023-02-16
Lossless optimization of SVGO logos

gucr:
A nit-picky PR to optimize SVGO logos.
That's a mean 30% reduction by using SVGO itself and a bit of manual cleanup


SethFalco:
Reviewing the result, this isn't actually lossless. 😱
In particular, the gradients of the colored variant have changed quite a bit. Not sure if those were from SVGO or the manual clean up.

Since there's no urgency for this change and SVGO is continuing to improve, I'll close it for now. I would be happy to approve a truly lossless PR, though it might be best to wait for the next release, as we're ironing out a lot of these smaller issues.

PR #1751, opened 2023-02-16
feat: remove useless closing path segments

jguddas:
The difference between those 2, is that the left side has has no close path command but is still closed due to ending at the start point.
```diff
- M12 22c6.23-.05 7.87-5.57 7.5-10-.36-4.34-3.95-9.96-7.5-10-3.55.04-7.14 5.66-7.5 10-.37 4.43 1.27 9.95 7.5 10z
+ M12 22c6.23-.05 7.87-5.57 7.5-10-.36-4.34-3.95-9.96-7.5-10-3.55.04-7.14 5.66-7.5 10-.37 4.43 1.27 9.95 7.5 10
```

### DIFF
--- a/plugins/convertPathData.js
+++ b/plugins/convertPathData.js
@@ -811,6 +811,7 @@ function filters(path, params, { maybeHasStrokeAndLinecap, hasRoundLinejoinAndLi
       if (
         params.removeUseless &&
         hasRoundLinejoinAndLinecap &&
+        !hasMarkerMid &&
         relSubpoint[0] === pathBase[0] &&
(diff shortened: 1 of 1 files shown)


SethFalco:
Sorry that this wasn't reviewed in a reasonable time. The maintainers were busy and I just joined recently!

This has been tackled in a recent PR and should be resolved in `main`.

If you have any qualms with the implementation, or can identify an edge case (with a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example)) that isn't handled how you'd expect, then feel free to open another issue/PR!

Issue #1748, opened 2023-02-15
[convertPath]: convert path segments to arcs


jguddas:
**Is your feature request related to a problem? Please describe.**
Path segments are sometimes shorter when written as arcs, but more importantly, once converted the existing [arc merge logic](https://github.com/svg/svgo/blob/aa5d667fcb5a33be9e51f958bcd20d2b719f8345/plugins/convertPathData.js#L475-L580) can be used to merge multiple segments into one.

**Describe the solution you'd like**
1. Convert the segment to arc. [C# Implementation](https://dlacko.org/blog/2016/10/19/approximating-bezier-curves-by-biarcs/#the-biarc-fitting)
2. Merge the arcs if they have the same center. [SVGO Implementation](https://github.com/svg/svgo/blob/aa5d667fcb5a33be9e51f958bcd20d2b719f8345/plugins/convertPathData.js#L475-L580)
3. Convert the merged arc back to a curve if doing so is shorter.

jguddas:
@SethFalco I have a working version [here](https://github.com/lucide-icons/lucide/blob/b2032aa225233dba8bfa93f58f993b6700f4f996/site/src/pages/api/edit/optimize.ts#L554-L566), code is not in the shape svgo expects and a bit of a mess tho.

Example: [https://lucide-site-git-feat-added-edit-page-jguddas.vercel.app/edit](https://lucide-site-git-feat-added-edit-page-jguddas.vercel.app/edit?PHN2ZwogIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIKICB3aWR0aD0iMjQiCiAgaGVpZ2h0PSIyNCIKICB2aWV3Qm94PSIwIDAgMjQgMjQiCiAgZmlsbD0ibm9uZSIKICBzdHJva2U9ImN1cnJlbnRDb2xvciIKICBzdHJva2Utd2lkdGg9IjIiCiAgc3Ryb2tlLWxpbmVjYXA9InJvdW5kIgogIHN0cm9rZS1saW5lam9pbj0icm91bmQiCj4KICA8cGF0aCBkPSJNOCAxMGEyIDIgMCAxIDAgMC00IDIgMiAwIDAgMCAwIDRaIiAvPgo8L3N2Zz4=)

KTibow:
Is this intended?
![image](https://github.com/svg/svgo/assets/10727862/5d986db1-364b-410a-b2a5-bc81bcd1ed6f)
[...]

PR #1754, opened 2023-02-27
feat: combine arcs #1748

jguddas:


plugins/convertPathData.js
```diff
@@ -403,180 +404,44 @@ function filters(path, params, { maybeHasStrokeAndLinecap, hasMarkerMid }) {
...
-          data = arc.args;
-          // @ts-ignore
-          item.coords = arc.coords;
-        }
+        // TODO check that stringify is shorter
+        command = 'a';
+        data = arc.item.args;
+        // @ts-ignore
+        item.coords = arc.item.coords;
+        // TODO: filter out consumed next items
```
jguddas, line 444: This currently creates useless 0 width path segments.

PR #1755, opened 2023-03-03
fix: useless path collapsing for segments that are not start and end

jguddas:


plugins/convertPathData.js
```diff
@@ -766,11 +766,18 @@ function filters(path, params, { maybeHasStrokeAndLinecap, hasMarkerMid }) {
...
-      if (params.removeUseless && !maybeHasStrokeAndLinecap) {
-        // l 0,0 / h 0 / v 0 / q 0,0 0,0 / t 0,0 / c 0,0 0,0 0,0 / s 0,0 0,0
+      // remove useless non-start/end path segments
+      var isStart = !path[index - 1] || path[index - 1].command === 'm';
+      var isEnd = !path[index + 1] || path[index + 1].command === 'm';
+      if (
+        params.removeUseless &&
+        !hasMarkerMid &&
+        (!maybeHasStrokeAndLinecap || !(isStart || isEnd))
+      ) {
```
jguddas, line 776: `hasRoundLinejoinAndLinecap` from https://github.com/svg/svgo/pull/1751 could also be included.

```suggestion
      if (
        params.removeUseless &&
        !hasMarkerMid &&
        (hasRoundLinejoinAndLinecap || !maybeHasStrokeAndLinecap || !(isStart || isEnd))
      ) {
```

plugins/convertPathData.js
```diff
@@ -766,11 +766,18 @@ function filters(path, params, { maybeHasStrokeAndLinecap, hasMarkerMid }) {
         }
       }
 
-      // remove useless non-first path segments
-      if (params.removeUseless && !maybeHasStrokeAndLinecap) {
-        // l 0,0 / h 0 / v 0 / q 0,0 0,0 / t 0,0 / c 0,0 0,0 0,0 / s 0,0 0,0
+      // remove useless non-start/end path segments
+      var isStart = !path[index - 1] || path[index - 1].command === 'm';
```
KTibow, line 770: Shouldn't this be `command === 'm'`, not testing if the previous one is `m`?

plugins/convertPathData.js
KTibow: This doesn't make sense as these do fundamentally different things.
If it has a linecap, we can't remove segments off the end. We can only remove segments in between.
If it doesn't have a linecap, we can remove all useless segments.
What you're proposing is that if it has a round linejoin, we can change `M5 5h0` to nothing. That would cause incorrect behavior, because the original shows up as a dot.

*I would personally write this as `maybeHasStrokeAndLinecap ? !isStart && !isEnd : true` as I find that way easier to read and understand, but style is personal.

plugins/convertPathData.js
```diff
@@ -766,11 +766,18 @@ function filters(path, params, { maybeHasStrokeAndLinecap, hasMarkerMid }) {
...
+      var isEnd = !path[index + 1] || path[index + 1].command === 'm';
```
KTibow, line 771: If I understand what I'm reading correctly, this current version doesn't remove move statements at the very end of the path, which are completely useless. It should.

PR #1757, opened 2023-03-07
feat: Add suffixPattern and doesPreserveClasses parameters to addClassesToSVGElement plugin

anselmbradford:
This PR adds a `suffixPattern` parameter to the `addClassesToSVGElement` plugin, which allows adding of a suffix to the custom classes added on the SVG element.

This PR also adds a `doesPreserveClasses` parameter, which when `true` will retain the existing classes on the SVG element, otherwise it will clear the existing classes and only use those from any `className` and `suffixPattern` parameters passed into the plugin.

### DIFF
--- a/plugins/addClassesToSVGElement.js
+++ b/plugins/addClassesToSVGElement.js
@@ -8,6 +8,7 @@ exports.description = 'adds classnames to an outer <svg> element';
 var ENOCLS = `Error in plugin "addClassesToSVGElement": absent parameters.
 It should have a list of classes in "classNames" or one "className".
 Config example:
+
 plugins: [
   {
     name: "addClassesToSVGElement",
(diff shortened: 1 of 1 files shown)


SethFalco:
A better way to do this would be to have the `addClassesToSVGElement` plugin accept `Array<string|function>` instead of just a `string[]`. The function can have the same signature as the function for `prefixIds` (`(node, info) => string`). 

Then you could do the following:

```js
module.exports = {
  plugins: [
    {
      name: "addClassesToSVGElement",
      params: {
        classNames: [
          'my-svg-icon',
          (_, info) => `my-svg-icon_${info.path.split('.')[0]}`,
        ]
      }
    }
  ]
}
```

The benefits to this approach is that:

1. It's consistent with the `prefixIds` plugin.
2. We don't add one-off special handling unique to this plugin, like what you did with `$FILENAME`.
3. Users have much more control in general, rather than specifically a suffix.
4. The way you've currently implemented this, it seems very specific to your personal workflow, while the new approach works out is easier to adapt for others.

Would you care to give this approach a shot and update your PR?

PR #1759, opened 2023-03-24
Fix tar.gz extraction for regressions testsuite file

strarsis:
The SVG test suite `tar.gz` file appears to be also gunzip-compressed, hence `tar-stream` alone used in the current regression test setup fails to extract the file, hence the whole regressions test fails:
`Download W3C SVG 1.1 Test Suite and extract svg files
Error: Invalid tar header. Maybe the tar is corrupted or it needs to be gunzipped?`

This PR adds [`gunzip-maybe`](https://github.com/mafintosh/gunzip-maybe) in between that can handle gunzip-compressed archives ([as recommended by the `tar-stream` docs](https://github.com/mafintosh/tar-stream#:~:text=Note%20that%20you%20still%20need%20to%20gunzip%20your%20data%20if%20you%20have%20a%20.tar.gz.%20We%20recommend%20using%20gunzip%2Dmaybe%20in%20conjunction%20with%20this.)).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4872 @@
+{
+  "name": "svgo",
+  "version": "3.0.2",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "svgo",
+      "version": "3.0.2",
+      "license": "MIT",
+      "dependencies": {
+        "@trysound/sax": "0.2.0",
+        "commander": "^7.2.0",
+        "css-select": "^5.1.0",
+        "css-tree": "^2.2.1",
+        "csso": "^5.0.5",
+        "gunzip-maybe": "^1.4.2",
+        "picocolors": "^1.0.0"
+      },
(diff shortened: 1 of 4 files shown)


SethFalco (commented):
I think instead of adding the dev dependency, it'd be better for us to just take the one function call that actually matters.

> ```js
>         swap(null, pumpify(zlib.createGunzip(), gunzip(maxRecursion - 1)))
> ```
> 
> — https://github.com/mafintosh/gunzip-maybe/blob/master/index.js#L21

The rest is just various checks or determining whether calling [`zlib#createGunzip`](https://nodejs.org/api/zlib.html#zlibcreategunzipoptions) is required or not. As we're hard-coding the URL, I think we can handle things ourselves.

Based on this, both review comments are redundant, but I thought it was worth sharing the FYI anyway! As I'd already tested this approach locally, I've created a separate PR for it here:

* https://github.com/svg/svgo/pull/1788

Issue #1746, opened 2023-01-27
Unique id is broken since v3


yoriiis:
**Describe the bug**

I need unique ids when multiple SVGs have `<defs>`. With `v2.8` I was using the following code and it works well.

```js
{
    name: 'preset-default',
    params: {
        overrides: {
            cleanupIDs: {
                prefix: {
                    toString() {
                        return crypto.randomBytes(20).toString('hex').slice(0, 4)
                    }
                }
            }
        }
    }
}
```

But in `v3` the `prefix` parameter of `cleanupIDs` is removed replacing `prefixIds`. I can't get the same rendering for the IDs.
I've tested this code but it breaks the `<defs>`.

```js
{
    name: 'prefixIds',
    params: {
        delim: '',
        prefix: function (node, info) {
            return crypto.randomBytes(20).toString('hex').slice(0, 4)
        }
    }
}
```

In input there are 2 svgs

**gradient.svg**

```html
<svgxmlns="http://www.w3.org/2000/svg"><defs><linearGradient id="MyGradient"><stop offset="5%" stop-color="green"/><stop offset="95%" stop-color="gold"/></linearGradient></defs><rect fill="url(#MyGradient)" width="100%" height="100%"/></svg>
```

**delete-alert.svg**

```html
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 19"><defs><linearGradient id="a" x1="0%" x2="100%" y1="9.5%" y2="90.5%"><stop offset="0%" stop-color="#FF0053"/><stop offset="100%" stop-color="#EE5B35"/></linearGradient></defs><g fill="none"><path fill="url(#a)" d="M15 8a5 5 0 110 10 5 5 0 010-10zm-9.81 7L9 16.36c-.48 1.27-1.72 1.9-2.77 1.54-1.04-.36-1.52-1.63-1.04-2.9zM11.2.84c.2-.68.92-.98 1.63-.78.71.2 1.02.88.82 1.56l-.41.98A5.87 5.87 0 0116 6.4h-.2c-3.98 0-7.14 3.03-7.14 6.84 0 .59 0 1.17.2 1.76L0 11.88l.3-.79 2.25-.97 1.73-4.69a6.1 6.1 0 016.62-3.71z" transform="translate(0 .5)"/><path fill="#FFF" d="M12 13h6v1h-6z"/></g></svg>
```

In output, with the `prefixIds` plugin, it produced

```html
[...]

PR #1763, opened 2023-04-08
Re-add prefix parameter on cleanupIds to fix unique ids

yoriiis:
Fixed: #1746

SVGO [v3.0.0](https://github.com/svg/svgo/releases/tag/v3.0.0) removed the `prefix` parameter in the `cleanupIds` plugin in favor of `prefixIds` plugin. This created a regression with id conflicts with multiple SVGs.

The PR propose a revert of the https://github.com/svg/svgo/commit/bc07c483c3b986c66834b5fe019bb75597dd45ab commit to re-add the parameter and let user add a prefix with a custom hash.

Exemple implementation to resolve the id conflicts.

```js
const crypto = require('crypto');

module.exports = {
    plugins: [
        {
            name: 'preset-default',
            params: {
                overrides: {
                    cleanupIDs: {
                        prefix: {
                            toString() {
                                return crypto.randomBytes(20).toString('hex').slice(0, 4);
                            }
                        }
                    }
                }
            }
        }
    ]
}
```

### DIFF
--- a/plugins/cleanupIds.js
+++ b/plugins/cleanupIds.js
@@ -111,10 +111,10 @@ const generateId = (currentId) => {
 /**
  * Get string from generated ID array.
  *
- * @type {(arr: Array<number>) => string}
+ * @type {(arr: Array<number>, prefix: string) => string}
  */
-const getIdString = (arr) => {
-  return arr.map((i) => generateIdChars[i]).join('');
+const getIdString = (arr, prefix) => {
+  return prefix + arr.map((i) => generateIdChars[i]).join('');
 };
 
 /**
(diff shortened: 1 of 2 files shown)


yoriiis:
@TrySound and @GreLI could you take a look to this PR please? Thanks 🙏

lusarz (approved):

PurpleTape:
Any updates? 🥺🥺🥺

yoriiis:
FYI @SethFalco

SethFalco:
@yoriiis I'm currently working on improving our documentation. As soon as I'm done with that, I'll give this a look!

While reviewing this though, I want to check out the `prefixIds` plugin and see if we can solve the issue there rather than maintain the functionality twice, or merge the plugins again.

There is one obvious bug I can see already in **Prefix IDs**, which is that the plugin modifies IDs and references to IDs separately, basically always breaking vectors when a function with random output is passed.

Regarding docs, can see progress here:
https://svgojs.org/docs/plugins/cleanup-ids/

PR #1770, opened 2023-06-03
chore: update postcss-svgo link

Lehoczky:
Hi 👋🏿 

The old link was broken, so I updated it^^


SethFalco:
Thanks for spotting this and providing the fix!

PR #1771, opened 2023-06-06
README: image-webpack-loader -> image-minimizer-webpack-plugin

tanyabouman:
[image-webpack-loader](https://github.com/tcoopman/image-webpack-loader) is deprecated and replaced with [image-minimizer-webpack-plugin](https://github.com/webpack-contrib/image-minimizer-webpack-plugin/#optimize-with-svgo)


SethFalco:
Thanks for spotting this and providing the fix!

PR #1772, opened 2023-06-20
Link to svgo.nvim in README.md

bennypowers:
add svgo.nvim


SethFalco:
Hey!

Sorry, but we'll no longer be adding clients to the list in the README.

1. We're finding that more of them are becoming unmaintained or shipping outdated versions of SVGO.
2. It takes some time to review the plugins to make sure they're fine to list, or to look back at them to check if they're still maintained.
3. We shouldn't just not review them, as in some cases, users have bad experiences or file bug reports for issues that were already resolved in a more recent version.

While those issues don't apply to all plugins, it's just not a good use of maintainer time to vet third-party clients or handle the issues above.

In future, we'll be removing that and updating the website with some of the more notable and well maintained third-party clients/libraries, and linking to [github.com/topics/svgo](https://github.com/topics/svgo) or similar for users to discover other plugins like yours.

PR #1788, opened 2023-09-22
test: fix regression testing action

SethFalco:
The regression-test GitHub Action was failing for all pull requests due to not handling the gzipped archive properly.

This resolved it by piping the response body to [`zlib#createGunzip`](https://nodejs.org/api/zlib.html#zlibcreategunzipoptions) first. Also updates the relevant dependencies.

## Related

* Closes https://github.com/svg/svgo/pull/1759

## Note

* I've noticed sometimes when running all tests locally or on CI, it hangs at the end. Will investigate this later. 🤔

### DIFF
--- a/test/regression-extract.js
+++ b/test/regression-extract.js
@@ -5,11 +5,16 @@ const path = require('path');
 const util = require('util');
 const zlib = require('zlib');
 const stream = require('stream');
-const fetch = require('node-fetch');
+const { default: fetch } = require('node-fetch');
 const tarStream = require('tar-stream');
 
 const pipeline = util.promisify(stream.pipeline);
 
+/**
+ * @param {string} url
+ * @param {string} baseDir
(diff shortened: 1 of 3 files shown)


test/regression-extract.js
```diff
@@ -39,7 +43,7 @@ const extractTarGz = async (url, baseDir, include) => {
     next();
   });
   const response = await fetch(url);
-  await pipeline(response.body, extract);
+  await pipeline(response.body.pipe(zlib.createGunzip()), extract);
```
TrySound, line 46: Since this is a pipeline I think you can pass zlib.createGunzip() as a second argument instead of calling .pipe.

TrySound (approved):

test/regression-extract.js
SethFalco: Ahh, thanks for the heads-up! You're right.

PR #1792, opened 2023-09-24
test: update jest

SethFalco:
Locally and on CI the tests sometimes hang at the end, but I don't see any obvious errors in our test suite. 🤔
 
Before investigating further, I think it'd be best to just update Jest and see how it goes. So far, I've been unable to reproduce the issue with this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
   },
   "eslintConfig": {
     "parserOptions": {
-      "ecmaVersion": "2021"
+      "ecmaVersion": 2021
     },
     "env": {
       "node": true,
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -122,10 +122,10 @@
     "@rollup/plugin-node-resolve": "^14.1.0",
     "@types/css-tree": "^2.0.0",
     "@types/csso": "^5.0.0",
-    "@types/jest": "^29.1.1",
+    "@types/jest": "^29.5.5",
     "del": "^6.0.0",
     "eslint": "^8.24.0",
-    "jest": "^29.1.2",
+    "jest": "^29.5.5",
```
TrySound, line 128: You can import from `@jest/globals` to get types instead of `@types/jest`

SethFalco: It's unclear to me how to use `@jest/globals` properly when the package isn't declared as a module. When I use the `require`, nothing works as expected, and I can't import the types with `@typedef`. 🤔 

Despite this, the tests do pass, so the library works. I'm just struggling with the TypeScript integration.

TrySound: Ah, right. Forgot we don't use ts syntax.

TrySound (approved):

PR #1805, opened 2023-09-30
chore: use external xast type declarations

SethFalco:
Removes our custom type declarations for Xast in favor of the ones suggested by [xast](https://github.com/syntax-tree/xast).

> If you are using TypeScript, you can use the xast types by installing them with npm:
> 
> ```sh
> npm install @types/xast
> ```
> 
> — https://github.com/syntax-tree/xast#types

## Motivation

While it makes sense to declare types of Xast for our plugin API, I think it'd be better to avoid maintaining them as part of this library/application if there are conventional solutions available already.

* Imo it's better to leave maintaining/redistributing them to @DefinitelyTyped, which [Titus Wormer](https://github.com/wooorm) and [Christian Murphy](https://github.com/ChristianMurphy) from syntax/xast helped define.
* The new types more strictly adhere to the Xast spec, which should simplify integrating with Xast libraries in future.

My hope was actually to explore some libraries surrounding Xast, in particular, [xast-util-from-xml](https://github.com/syntax-tree/xast-util-from-xml) and [xast-util-select](https://github.com/TrialAndErrorOrg/xast-util-select). Unfortunately, neither of these meet fulfil our needs atm.

---

Not 100% sure if this is a worthwhile change, so feedback welcome.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -1,16 +1,17 @@
 'use strict';
 
 /**
- * @typedef {import('./types').XastNode} XastNode
- * @typedef {import('./types').XastInstruction} XastInstruction
- * @typedef {import('./types').XastDoctype} XastDoctype
- * @typedef {import('./types').XastComment} XastComment
- * @typedef {import('./types').XastRoot} XastRoot
- * @typedef {import('./types').XastElement} XastElement
- * @typedef {import('./types').XastCdata} XastCdata
- * @typedef {import('./types').XastText} XastText
- * @typedef {import('./types').XastParent} XastParent
- * @typedef {import('./types').XastChild} XastChild
+ * @typedef {import('xast').Cdata} Cdata
+ * @typedef {import('xast').Comment} Comment
+ * @typedef {import('xast').Doctype} Doctype
+ * @typedef {import('xast').Element} Element
+ * @typedef {import('xast').ElementContent} ElementContent
+ * @typedef {import('xast').Instruction} Instruction
+ * @typedef {import('xast').Node} Node
+ * @typedef {import('xast').Parents} Parents
+ * @typedef {import('xast').Root} Root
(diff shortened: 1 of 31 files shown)


TrySound:
Working with attributes with these types is less trivial though parser produce only string values. I'd prefer to stay with old types.

This change also introduce 2 more packages. xast-util-from-xml would bring even more.
Last couple months I had very bad Internet connection and loading myriads of packages in node_modules made it worse. So I'd encourage you to keep svgo more concentrated.

Underlying xml parser seem interesting though. We could write own wrapper around it and replace sax.

SethFalco:
Sounds good, I'll shelve this then!

> Underlying xml parser seem interesting though.

Unfortunately, one thing that will stop us from using it for now is the lack of support for DTD entities.

https://github.com/rgrove/parse-xml#not-features

Not sure if that's planned, or something we'd be able to work around. I'll review both later, but right now, I'm more focused on documentation.

Sharing a WIP for the sake of it:

TrySound:
Awesome!

SethFalco:
@TrySound Seems we're unable to exchange email. Yandex thinks I'm a spammer. 😭

The documentation isn't ready yet, but I'm making good progress and think it will be ready for other maintainers to review soon. My goal is by 6 October 2023!

It's already hosted on GitHub pages, but from my namespace:
* https://svgojs.org/
* The [Cleanup IDs](https://svgojs.org/docs/plugins/cleanup-ids/) page is a pretty good showcase for info, parameters, and practical demo.

Also willing to transfer the domain to Kir when or if requested.

Would it be alright for you to create a new repo called **svg/svgojs.org** or **svg/website**, so I can open a PR there when it's ready?

TrySound:
Wrote to you in matrix

Issue #1811, opened 2023-10-19
removeComments Plugin not working as expected.


fmotion1:
**Describe the bug**
I want to remove comments from my SVG files:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512">
<!--! Font Awesome Free 6.4.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license Copyright 2023 Fonticons, Inc.-->
  <path d="M384 64c17.7 0 32 14.3 32 32v320c0 17.7-14.3 32-32 32H64c-17.7 0-32-14.3-32-32V96c0-17.7 14.3-32 32-32h320zM64 32C28.7 32 0 60.7 0 96v320c0 35.3 28.7 64 64 64h320c35.3 0 64-28.7 64-64V96c0-35.3-28.7-64-64-64H64zm160 96a128 128 0 1 1 0 256 128 128 0 1 1 0-256zm0 288a160 160 0 1 0 0-320 160 160 0 1 0 0 320zm0-136a24 24 0 1 0 0-48 24 24 0 1 0 0 48z"/>
</svg>
```

The `removeComments` plugin is supposedly enabled by default, but the comments remain after processing. This is what I want to remove:

```svg
<!--! Font Awesome Free 6.4.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license Copyright 2023 Fonticons, Inc.-->
```

**To Reproduce**

1. Create svgo.config.js as outlined:

```js
// svgo.config.js
module.exports = {
    multipass: true,
    js2svg: {
      indent: 2, 
      pretty: true,
    },
    plugins: [  
        { name: 'removeComments' }
    ]
  };
```

2. Run `svgo --config=svgo.config.js "FontAwesome Free 6 Light Album Circle Plus.svg"`
3. Open the resulting SVG and notice that the comment is still present.

**Expected behavior**
All comments to be nuked from the SVG file.

**Desktop (please complete the following information):**
 - SVGO Version 3.0.2
 - NodeJs Version v20.8.0
 - OS: Windows 10 Pro x64 Version 22H2 (Build 19045.3448)
 - 
**Additional context**
Not sure what I'm doing wrong. All I want to do is strip comments.

fmotion1:
**Update:** 

It works if I open `removeComments.js` in `Roaming\nvm\v20.8.0\node_modules\svgo\plugins` and then comment out the conditional: `if (node.value.charAt(0) !== '!')`

[...]

PR #1812, opened 2023-10-22
fix(removeComments): make preserving comments a param

SethFalco:
Converts the hard-coded exception to a param so users can override this behavior to preserve custom patterns or drop the logic altogether.

We retain the existing exception as the default value, as this convention is interpreted across minifiers as important. For example, Font Awesome prefixes their comment with a `!`.

```
<!--! Font Awesome Free 6.4.2 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2023 Fonticons, Inc. -->
```

However, we also allow users to pass custom patterns, or either `preservePatterns: []` or `preservePatterns: false` to indiscriminately remove comments.

This is completely backward compatible.

I also replaced the `ignored comment` in the tests with something that matches the kind of comment we're expecting to encounter with this prefix.

## Related

* Closes https://github.com/svg/svgo/issues/1811

### DIFF
--- a/plugins/removeComments.js
+++ b/plugins/removeComments.js
@@ -6,6 +6,12 @@ exports.name = 'removeComments';
 exports.description = 'removes comments';
 
 /**
+ * If a comment matches one of the following patterns, it will be
+ * preserved by default. Particularly for copyright/license information.
+ */
+const DEFAULT_PRESERVE_PATTERNS = [/^!/];
+
+/**
  * Remove comments.
  *
  * @example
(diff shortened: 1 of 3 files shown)


plugins/removeComments.js
```diff
@@ -5,6 +5,12 @@ const { detachNodeFromParent } = require('../lib/xast.js');
 exports.name = 'removeComments';
 exports.description = 'removes comments';
 
+/**
+ * If a comment matches one of the following patterns, it will be
+ * preserved by default. Particularly for copyright/license information.
+ */
+const DEFAULT_PRESERVE_PATTERNS = [/^!/, /LICENSE/i, /COPYRIGHT/i];
```
SethFalco, line 12: Does anyone here know the motivation for the `!` check?
With the new preserve patterns param, perhaps that one can be removed in v4?

plugins/removeComments.js
```diff
@@ -5,6 +5,17 @@ const { detachNodeFromParent } = require('../lib/xast.js');
 exports.name = 'removeComments';
 exports.description = 'removes comments';
 
+/**
+ * If a comment matches one of the following patterns, it will be
+ * preserved by default. Particularly for copyright/license information.
+ */
+const DEFAULT_PRESERVE_PATTERNS = [
+  /^!/,
```
SethFalco, line 13: Does anyone here know the motivation for the `/^!/` check?
With the new preserve patterns param, perhaps that one can be removed in v4?

TrySound: `!` is a convention for preservable comments like license. All minifiers rely on it.
No need to check for keywords with it.

SethFalco: Interesting, is this convention defined anywhere, even if only informally? I'd just like to make a note and reference it in the documentation and JSDocs? I've tried searching for this, but don't see a resource that makes note of this. 🤔

In my opinion, it makes sense to retain the keywords as well if this isn't a defined standard. In the worst-case scenario, it just adds a bit of redundancy while processing, what do you think?

PR #1816, opened 2023-11-03
feat(convertShapeToPath): convert rect with rounded corners using arcs 

JoCa96:
Convert `rect` shape with a defined `rx` and/or `ry` to path when the `convertArcs` parameter is true.

E.g.
```svg
<rect x="40" y="50" width="30" height="20" rx="2" ry="5"/>
```
will become
```svg
<path d="M68 50A2 5 0 0 1 70 55L70 65A2 5 0 0 1 68 70L42 70A2 5 0 0 1 40 65L40 55A2 5 0 0 1 42 50z"/>
```

When `rx` and `ry` are invalid or `0` the conversion will be performed without arcs as there are no rounded corners in that case.


KTibow:
why would you want to do this? this seems like the opposite of optimization

JoCa96:
> why would you want to do this? this seems like the opposite of optimization

That's true for all conversions that are enabled by the `convertArcs` flag.

The need for this is use-case dependant. 
If the flag is enabled it should do all the arc conversions imho. We want to reduce all shapes to a single path. You can also see the need for this in PR  #1767.

jguddas:
This is basically the opposite of #1748, there are definitely cases where one is closer than the other.

PR #1822, opened 2023-11-05
replace with z and remove z when optimal

### LINKED ISSUE #1821: Convert lines home to z
Paths like `m 0 0 h 5 v 5 h -5 v -5` should be converted to `m 0 0 h 5 v 5 h -5 z` for better minification. That is, convert the line home to `z`.
(To follow up on that, you could also convert `m 0 0 a 2.5 2.5 0 0 0 0 5 a 2.5 2.5 0 0 0 0 -5 z` to `m 0 0 a 2.5 2.5 0 0 0 0 5 a 2.5 2.5 0 0 0 0 -5`. That is, remove a redundant `z`.)

I already made a plugin that does the first one, but it should probably be rewritten to fit in to the existing path transformer better. And the second one requires disabling on stroke. Anyway, code of the plugin, in case it helps:
```ts
	const processPath: Visitor = {
		element: {
			enter(element) {
				const path =
					'pathJS' in element && (element.pathJS as PathDataItem[]);
				if (!path) return;

				const start = [0, 0];
				const cursor = [0, 0];

				for (let i = 0; i < path.length; i += 1) {
					const pathItem = path[i];
					const { command, args } = pathItem;

					// moveto (x y)
					if (command === 'm') {
						cursor[0] += args[0];
						cursor[1] += args[1];
						start[0] = cursor[0];
						start[1] = cursor[1];
					}
					if (command === 'M') {
						cursor[0] = args[0];
						cursor[1] = args[1];
						start[0] = cursor[0];
						start[1] = cursor[1];
					}

					// lineto (x y)
					if (command === 'l') {
						cursor[0] += args[0];
						cursor[1] += args[1];
					}
					if (command === 'L') {
						cursor[0] = args[0];
						cursor[1] = args[1];
					}

					// horizontal lineto (x)
					if (command === 'h') {
						cursor[0] += args[0];
					}
					if (command === 'H') {
						cursor[0] = args[0];
					}

					// vertical lineto (y)
					if (command === 'v') {
						cursor[1] += args[0];
					}
					if (command === 'V') {
						cursor[1] = args[0];
					}

					if (['l', 'L', 'h', 'H', 'v', 'V'].includes(command)) { // !! the part that does stuff
						if (start[0] == cursor[0] && start[1] == cursor[1])
							path[i] = { command: 'z', args: [] };
					}

					// curveto (x1 y1 x2 y2 x y)
					if (command === 'c') {
						cursor[0] += args[4];
						cursor[1] += args[5];
					}
					if (command === 'C') {
						cursor[0] = args[4];
						cursor[1] = args[5];
					}

					// smooth curveto (x2 y2 x y)
					if (command === 's') {
						cursor[0] += args[2];
						cursor[1] += args[3];
					}
					if (command === 'S') {
						cursor[0] = args[2];
						cursor[1] = args[3];
					}

					// quadratic Bézier curveto (x1 y1 x y)
					if (command === 'q') {
						cursor[0] += args[2];
						cursor[1] += args[3];
					}
					if (command === 'Q') {
						cursor[0] = args[2];
						cursor[1] = args[3];
					}

					// smooth quadratic Bézier curveto (x y)
					if (command === 't') {
						cursor[0] += args[0];
						cursor[1] += args[1];
					}
					if (command === 'T') {
						cursor[0] = args[0];
						cursor[1] = args[1];
					}

					// elliptical arc (rx ry x-axis-rotation large-arc-flag sweep-flag x y)
					if (command === 'a') {
						cursor[0] += args[5];
						cursor[1] += args[6];
					}
					if (command === 'A') {
						cursor[0] = args[5];
						cursor[1] = args[6];
					}

					// closepath
					if (command === 'Z' || command === 'z') {
						// reset cursor
						cursor[0] = start[0];
						cursor[1] = start[1];
					}
				}
				element.attributes.d = stringifyPathData({
					pathData: path,
				});
			},
		},
	};
```

KTibow:
## What
closes #1821 by adding a mechanism that tries to prefer z over other methods of going home and removes z when it's redundant.

how svg handles `z` is complicated but i think i added enough safeguards. [the tests pass](https://github.com/KTibow/svgo/actions/runs/6762610404) so hopefully it doesn't break anything.
  - the line-z pattern can always be converted to z. this is accounted for by checking if the next command is z.
  - when we detect either no stroke or completely round strokes, line is the same as z. this lets us convert line to z, and if you disabled converttoz we can still convert line-z to line.
  - when we detect a stroke that isn't round, the optimizations are limited, as i observed with the svg test suite. if you convert ITEM-z to ITEM, it'll change a closed shape to a non-closed shape. if you convert LINE to z, it'll change a non-closed shape to a closed shape. but i added safeguards so it shouldn't do that.
  - closes #1751

## Results
Material Design Icons: 98% current svgo 2,651kb -> this 2,601kb
some random vector 95% current svgo 12.2kb -> this 11.7kb


SethFalco:
Hey, thanks for opening the issue and submitting a pull request. This is a pretty nice optimization.

Also, thanks for being on top of this PR, especially regarding documentation. 👍🏽
Just wanted to apologize for keeping you waiting. I'm going to release v3.0.3 soon, and then I'll review this in the upcoming days.

For now, can you just confirm, this would close https://github.com/svg/svgo/pull/1751 and https://github.com/svg/svgo/pull/1755 in theory, right? If this is the case, would you mind adding their example in #1751 as a test case in your PR?

KTibow:
> For now, can you just confirm, this would close https://github.com/svg/svgo/pull/1751 and https://github.com/svg/svgo/pull/1755 in theory, right?

Definitely closes #1751. I think 1755 can still stay open.
> If this is the case, would you mind adding their example in https://github.com/svg/svgo/pull/1751 as a test case in your PR?

Just to confirm, you mean as in adding it to the automated tests?

SethFalco:
> Definitely closes https://github.com/svg/svgo/pull/1751. I think 1755 can still stay open.

Sure, for now, I'll go with that. Once this is merged, I'll check the other PR to see whether it's resolved or not, and hit up the contributor.

> Just to confirm, you mean as in adding it to the automated tests?

Yeah! So, creating a `convertPathData.xx.svg` with that scenario. Doesn't have to the same path, just the same optimization that was being demonstrated there.

plugins/convertPathData.js
```diff
@@ -167,6 +170,14 @@ exports.fn = (root, params) => {
             (computedStyle['stroke-linecap'].type === 'dynamic' ||
               computedStyle['stroke-linecap'].value !== 'butt');
           const maybeHasStrokeAndLinecap = maybeHasStroke && maybeHasLinecap;
+          // TODO: when stroke is used, prefer z more
+          // Z is the same as going home at 90 and 0 degrees, but not at other degrees
```
SethFalco, line 174:
```suggestion
          // Z is the same as going home at 90 and 0 degrees, but not at other degrees
```

Can we drop the TODO?
When or if we have a solution for preferring `z` in more cases involving stroke, we can create an issue and work on it. I'd prefer not to pollute the code with tasks, though.

If you already have an idea of some cases we could prefer `z` but currently don't, I'd appreciate it if you could open an issue with the scenarios and I can look at it if you don't have time.

SethFalco (commented):
This looks great, thanks for opening the issue and working on this. I've left 3 minor comments, once addressed I'd be happy to merge.

plugins/convertPathData.js
KTibow: I removed the whole thing, given that since I implemented always converting LINE-z to z, this isn't as much of a concern as it was when I first wrote the comment

SethFalco:
Awesome, thanks again for working on this. 👍🏽
This will be released in v3.0.4.

PR #1835, opened 2023-11-14
chore: revamp readme

SethFalco:
Revamps the README to tidy things up, reference the new documentation, and drop some unmaintained or deprecated ways to use SVGO.

## Hero

I changed the `alt` attribute of the hero image because it wasn't fulfilling it's purpose. The `alt` attribute is what should be used instead if the user is unable to see the image for whatever reason. However, this image is purely decorative and so can be skipped over in the scenario it doesn't load.

See: https://www.w3.org/WAI/tutorials/images/decision-tree/

### Shields

Adds a badge guiding users to the SVGO documentation. I also modified the alt to match the text we'd expect to see if the badge loaded.

### Content

* Proofread the README, revising grammar and making some subjective changes.
* SVG is an acronym, so all uppercase. (`svg` → `SVG`)
* Reduce verbosity in general, particularly when describing the config.
* Instead of displaying the plugin list in a table here, link to the SVGO documentation.

### Troubleshooting

I've dropped the troubleshooting section, as caveats of SVGO and each plugin are explained in the documentation. The particular example with `removeViewBox` has also been covered to hell across mediums.

### Ways to use SVGO

I've deleted some entries, either due to planned deprecation, or because it still uses SVGO v0 or SVGO v1.

I will be working with third-party tools to try to get everyone upgraded to SVGO v3, but once we have released SVGO v4, I would like to remove tools still on SVGO v2 from this list as well.

* SVGO Action - planned deprecation
* mimosa-minify-svg - SVGO pre-v2 and unmaintained
* svgo-osx-folder-action - Repo is archived
* svgo_bot - SVGO pre-v2 and unmaintained
* Image Shrinker - SVGO pre-v2 and unmaintained

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,78 +1,83 @@
 <div align="center">
-  <img src="./logo/logo-web.svg" width="348.61" height="100" alt="SVGO logo"/>
+  <img src="./logo/logo-web.svg" width="348.61" height="100" alt=""/>
 </div>
 
-## SVGO [![npm version](https://img.shields.io/npm/v/svgo)](https://npmjs.org/package/svgo) [![Discord](https://img.shields.io/discord/815166721315831868)](https://discord.gg/z8jX8NYxrE)
+# SVGO [![npm](https://img.shields.io/npm/v/svgo)](https://npmjs.org/package/svgo) [![chat](https://img.shields.io/discord/815166721315831868)](https://discord.gg/z8jX8NYxrE) [![docs](https://img.shields.io/badge/docs-svgo.dev-blue)](https://svgo.dev/)
 
-**SVG O**ptimizer is a Node.js-based tool for optimizing SVG vector graphics files.
+SVGO, short for **SVG O**ptimizer, is a Node.js library and command-line application for optimizing SVG files.
 
 ## Why?
 
-SVG files, especially those exported from various editors, usually contain a lot of redundant and useless information. This can include editor metadata, comments, hidden elements, default or non-optimal values and other stuff that can be safely removed or converted without affecting the SVG rendering result.
+SVG files, especially those exported from vector editors, usually contain a lot of redundant information. This includes editor metadata, comments, hidden elements, default or suboptimal values, and other stuff that can be safely removed or converted without impacting rendering.
 
 ## Installation
 
-Via npm:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,118 +1,128 @@
...
-Via npm:
-```sh
-npm -g install svgo
-```
-Via yarn:
+You can install SVGO globablly through npm, yarn, or pnpm. Alternatively, drop the global flag (`global`/`-g`) to use it in your Node.js project.
+
 ```sh
+# npm
+npm install -g svgo
```
TrySound, line 19: These days npx, yarn dlx or pnpx are more popular ways to run something globally.

TrySound (approved):

README.md
SethFalco: Sorry for the quick merge! I should'be left this open clearly. ^-^'

I actually opted for the individual commands intentionally because while looking `npx`, I've found it's very slow. This checks out with the fact that typically when I see it, it's for running one-time tasks like `create-react-app`. 

Every time one does `npx svgo`, it will check through node modules etc, taking 2 seconds just to run `npx svgo --help`. Not the ideal UX, so I thought it'd be best to document them separately, which is also handy to explain how to depend on it in a Node.js project anyway.

README.md
```diff
@@ -1,118 +1,128 @@
...
 svgo --help
 ```
 
 ## Configuration
 
-SVGO has a plugin-based architecture, separate plugins allows various xml svg optimizations. See [built-in plugins](#built-in-plugins).
-SVGO automatically loads configuration from `svgo.config.js` or from `--config ./path/myconfig.js`. Some general options can be configured via CLI.
+SVGO has a plugin architecture. You can read more about all plugins in [Plugins | SVGO Documentation](https://svgo.dev/docs/plugins/), and the default plugins in [Preset Default | SVGO Documentation](https://svgo.dev/docs/preset-default/).
 
+SVGO reads the configuration from `svgo.config.js` or the `--config {{path/to/config.js}}` command-line option. Some other parameters can be configured though command-line options too.
```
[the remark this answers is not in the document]
SethFalco: Sure, I just borrowed the syntax from [tldr-pages](https://github.com/tldr-pages/tldr).  
I'll update this locally and include it as a chore in my next PR.

PR #1842, opened 2023-11-19
fix(convertPathData): account for error in z conversions

KTibow:
You know how 0.1 + 0.2 = 0.30000000000000004? I was trying to prove that an icon can be optimized more, got curious why it wasn't getting converted to z, and found the culprit: `===`.
This PR fixes it using `error` to check for closeness.

## Results
-130b from Arch Linux
-450b from Isometric Madness
-7b from Wikipedia


KTibow:
Facing weird test behavior on convertPathData.11. When I change the expected output to one thing it produces the unexpected. When I change it to that it starts producing the other output.

KTibow:
It's because it tests it both with multipass and without. The thing is that the matrix transform leads to it being 0.001106 away (not enough to convert to z) but once it's stringified it comes out to 0.001 away (enough to convert to z). I'll solve this by just making sure it can convert in all the cases.

test/plugins/convertPathData.11.svg
```diff
@@ -28,7 +28,7 @@
     <path d="M0 300a1 2 0 1 0 200 0a1 2 0 1 0 -200 0" transform="rotate(15 100 300) scale(.8 1.2)"/>
     <path d="M700 300a1 2 0 1 0 200 0a1 2 0 1 0 -200 0" transform="rotate(-75 700 300) scale(.8 1.2)"/>
     <path d="M12.6 8.6l-3.1-3.2-3.1 3.2-.8-.7 3.9-3.9 3.9 3.9zM9 5h1v10h-1z" transform="rotate(-90 9.5 9.5)"/>
-    <path d="M637.43 482.753a43.516 94.083 0 1 1-87.033 0 43.516 94.083 0 1 1 87.032 0z" transform="matrix(1.081 .234 -.187 .993 -37.573 -235.766)"/>
+    <path d="M637.43 482.753a43.516 94.083 0 1 1-87.033 0 43.516 94.083 0 1 1 87.033 0z" transform="matrix(1.081 .234 -.187 .993 -37.573 -235.766)"/>
```
SethFalco, line 31: Could you clarify why you changed this line in the existing test?

SethFalco (commented):
The change makes sense, nice catch!

And yeah, the multipass tests tripped me up as well lately, I experienced the same behavior. ^-^'

That's in place since we want to make sure the plugin does everything it could want to do in a single execution, i.e. running SVGO with `multipass` but just a single plugin should be redundant.

Just a question regarding the modified test, others happy to merge!

test/plugins/convertPathData.11.svg
KTibow: >>> The thing is that the matrix transform leads to it being 0.001106 away (not enough to convert to z) but once it's stringified it comes out to 0.001 away (enough to convert to z). I'll solve this by just making sure it can convert in all the cases *(that is, making it so that the original is a perfect ellipse, the original didn't quite close properly)*.

SethFalco: > I'll solve this by just making sure it can convert in all the cases

Alright, I just reviewed the individual commits following that comment, so I follow now. I thought you were referring to your issue with `multipass`, and fixed the implementation, not an issue with the test case.

I think it's best to leave inputs for existing tests as-is rather than update them to fit new code. If the existing test case can't utilize a certain optimization, then that's what the test should enforce. Alternatively, if it should be able to, then that's a bug.

If I understand the issue correctly, then rather than update the test, the implementation could round the figure to the desired precision before comparing it to `error`, i.e.

```js
    if (
      (command === 'Z' || command === 'z') &&
      params.removeUseless &&
      isSafeToUseZ &&
      // @ts-ignore
      toFixed(Math.abs(item.base[0] - item.coords[0]), precision) <= error &&
      // @ts-ignore
      toFixed(Math.abs(item.base[1] - item.coords[1]), precision) <= error
    )
      return false;
```

**Original**
```svg
<path d="M637.43 482.753a43.516 94.083 0 1 1-87.033 0 43.516 94.083 0 1 1 87.032 0z" transform="matrix(1.081 .234 -.187 .993 -37.573 -235.766)"/>
```

→ 

**Optimized**
```svg
<path d="M561.214 392.766a48.107 95.08 10.132 1 1-94.083-20.365 48.107 95.079 10.132 1 1 94.082 20.365"/>
```

See any issues with this? With the proposed change, tests pass, regression passes, and `z` is dropped in the original test case.

KTibow: That might cause other issues though - let's look at numbers with a difference of `.4`, or rounded to `0`. While `1.0` and `1.4` both round to `1`, `1.3` rounds to `1` while `1.7` rounds to `2`. It's not always consistent. This essentially uses 5x the error which might cause things to look bad too. I'm just going to revert the input for now because it was fixed by other changes.

SethFalco: Ahh my bad, you're right. 🤔
In any case, just wanted to avoid changing inputs for previous tests for now, so thanks for reverting that.

SethFalco:
Looks good to me, thanks for working on this!
This will be released in SVGO v3.0.5. (ETA, 1 December 2023)

PR #1844, opened 2023-11-21
feat: Improve loadConfig types

nuintun:
Improve loadConfig types, fixed like:

```ts

await loadConfig(null, 'webpack context');
await loadConfig(undefined, 'webpack context');
```


SethFalco:
Hey! Thanks for reporting and fixing the issue.
This will be released in SVGO v3.0.5.

PR #1848, opened 2023-11-23
New plugin: optimize path order

### LINKED ISSUE #78: Closed path micro-optimization: substitute z for the most wasteful segment
A simple closed path(*) terminated with a single `z` (the logic gets hairier when you have multiple `z`s) such as

```
<path d="m 10 0 v 10 l -10 -10 z">
```

…can be rotated around until its `z` close-path segment (here used in place of a short `h 10` segment) replaces the most space-consuming of the line segments (here `l -10 -10`):

```
<path d="m 0 0 h 10 v 10 z">
```

(*) the example is a synonym to `<polygon points="0,10 10,10 0,0">` – but the optimization can be applied to any path with curves, arcs and other things in it, as long as there is at least a line segment in there too

### LINKED ISSUE #1847: Plugin: optimize polylines/other segments
*This could be added to convertPaths or made into a new plugin*

**Describe what the plugin should do**
Paths can often be optimized by changing their direction.
Base path: `M16 14v-2H8v2Z` (14b)
Optimized by changing the start: `M8 14h8v-2H8Z` (13b)
Optimized by reversing direction and changing the start: `M8 12h8v2H8z` (12b)

**Implementation**
- [ ] Are you volunteering to work on this plugin?
*I will try but I might not manage to make something that works properly - this would require somehow separating each segment, trying out each possibility, and converting to mixed.*

KTibow:
Closes #1847
Closes #78 (damn thats old)
To understand the plugin read #1847
Right now it only does polylines, might add arcs later

| Name | Standard | Multipass |
|--------|--------|--------|
| Arch Linux logo | -30b | -19b |
| Blobs | 0b | 0b |
| Isometric Madness | -588b | -797b |
| tldr-pages banner | 0b | 0b |
| Wikipedia logo | -32b | -25b |

### DIFF
--- a/lib/builtin.js
+++ b/lib/builtin.js
@@ -23,6 +23,7 @@ exports.builtin = [
   require('../plugins/minifyStyles.js'),
   require('../plugins/moveElemsAttrsToGroup.js'),
   require('../plugins/moveGroupAttrsToElems.js'),
+  require('../plugins/optimizePathOrder.js'),
   require('../plugins/prefixIds.js'),
   require('../plugins/removeAttributesBySelector.js'),
   require('../plugins/removeAttrs.js'),
--- a/plugins/optimizePathOrder.js
+++ b/plugins/optimizePathOrder.js
@@ -1,6 +1,6 @@
 const { stringifyPathData } = require('../lib/path.js');
 const { computeStyle, collectStylesheet } = require('../lib/style.js');
-const { hasScripts } = require('../lib/svgo/tools.js');
+const { hasScripts, cleanupOutData } = require('../lib/svgo/tools.js');
 const { pathElems } = require('./_collections.js');
 const { path2js, js2path } = require('./_path.js');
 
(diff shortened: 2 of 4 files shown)


plugins/mergePaths.js
```diff
@@ -17,7 +17,7 @@ exports.description = 'merges multiple paths in one if possible';
 exports.fn = (root, params) => {
   const {
     force = false,
-    floatPrecision,
+    floatPrecision = 3,
```
KTibow, line 20: If I don't do this, whenever I set an exact value for an argument it leaks out, despite my plugin having a default set.

anthonyryan1:
I just thought I'd chime in. This is the first plugin thing I've seen in years that could improve the compression of a few of my hand-optimized ultra small files.

I had previously thought there was little more optimization potential in them.

Great idea

xerc:
@KTibow you have time/motiv. to merge? https://github.com/svg/svgo/pull/2118
```bash
git mv docs/03-plugins/optimize-path-order.mdx docs/04-plugins/optimizePathOrder.mdx

PR #1850, opened 2023-11-23
test failed on Windows

### LINKED ISSUE #1849: svgo-node.test.js fails on Windows
'should respect line-ending config' fails on Windows, apparently because attributes on the expected value are in incorrect order.

1. **yarn run test** on Windows

All tests pass except the above.

- SVGO Version 3.0.4
- NodeJs Version 20.9.0
- OS: Windows 11

johnkenny54:
'should respect line-ending config' failed because parameters were out of order on expected result. CRLF version of the tests was out of sync with LF version.

Fixes #1849

### DIFF
--- a/lib/svgo-node.test.js
+++ b/lib/svgo-node.test.js
@@ -120,7 +120,7 @@ describeCRLF('with CRLF line-endings', () => {
     });
     // using toEqual because line endings matter in these tests
     expect(data).toEqual(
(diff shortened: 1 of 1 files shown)


SethFalco:
Thanks for reporting and fixing this for us!

This one was my fault, those tests get skipped both for me and on our CI environment. ^-^'
Later I'll evaluate if we can include a Windows runner in our pipeline to catch this in future.

Issue #1858, opened 2023-11-29
altered paths


nschloe:
Running

```
svgo in.svg
```

on

![in](https://github.com/svg/svgo/assets/181628/cc9f91cf-f826-43c1-81ac-ccf364c0ce16)

gives

![in-opti](https://github.com/svg/svgo/assets/181628/8ee093e8-8137-4ada-93a7-38156eaab7d0)

(Note how the axes moved.)

This is with svgo 3.0.4.

SethFalco:
Hey! Thanks for reporting the issue.

The issue is caused due to rounding. By default, SVGO reduces the precision of floating-point numbers to 5 decimal places.

In most cases, this is fine. However, in this SVG some of the numbers are _super_ inflated, with matrixes applied that are really miniscule.

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 673 205">
  <path d="M0 0L6235704.138785 2260595.537351" transform="matrix(0.0000431728,0,0,0.0000431865,60.47463,20.8221)" stroke-width="12700" stroke="black"/>
</svg>
```

Here is a trimmed down version of the SVG that showcases the issue. See how the path commands use numbers like 6,235,704, and then to scale it down to fit in the viewport a transform is applied using numbers like `0.0000431728`. The `0.0000431728` is being rounded down to `0.00004`. With numbers at this scale, this ends up making a significant difference.

The optimal solution to this would be to implement https://github.com/svg/svgo/issues/624, more specifically to apply the `matrix` to the path before rounding.

For now, a workaround you can do is increase the transform precision of the `convertTransform` plugin. You can create the following config in the directory you're running the command in, then try again.

**`svgo.config.js`**
```svg
module.exports = {
  plugins: [
    {
      name: "preset-default",
      params: {
        overrides: {
          convertTransform: {
            transformPrecision: 10
          }
        }
      }
    }
  ]
}
```

nschloe:
Interesting! Is there a way to disable the float clipping entirely?

Issue #1810, opened 2023-10-19
Plugin 'convertTransform' corrupts image


phpcoder2022:
**Step to reproduce:**

1. Clone repo from my [bug test project](https://github.com/phpcoder2022/svgo-bug)
1. Go to repo directory
1. `npm i`
1. `npm run build`
1. Run `serve` in `dist` directory
1. Visit server page
1. See that image was showed correctly
1. Stop `serve`
1. Delete value in `excludedSvgoPluginsBecauseItCorruptImages` in `webpack.config.js`
1. `npm run build`
1. Run `serve` in `dist` directory
1. Visit server page
1. See that image was corrupted

**Actual result:**

Image was corrupted

**Expected result:**

Image on second view must be equal image on first view

**Svgo version**: `^3.0.2`
**NodeJS version**: `v18.13.0`
**OS version**: `Ubuntu 23.04` Lunar Lobster

SethFalco:
The problem is ultimately rounding. The `convertTransform` plugin reduces the number of decimal places, which changes in output to `-.125` to `-.13`.

We make the assumption that there's no need to add more precision than there already is. So since it started with 2 decimal places, we assume that we can round to two decimal places at the end.

We should make this do one of the following:

* Use significant digits instead of decimal places to determine floating-point precision.
* Take the existing number of decimal places + 1 to calculate the desired precision, the plus 1 offset being what we consider a non-negligible change.
* Drop that logic altogether.

There may be edge cases where this still screws up the SVG, but this should certainly improve the situation for now.

phpcoder2022:
I created additional branch [`overrides`](https://github.com/phpcoder2022/svgo-bug/tree/overrides), where I overrided svgo plugins options. Three commits was placed on this branch and in any ones `webpack.config.js` have different settings. I discover that disabling of option `collapseIntoOne` solved this problem, but increasing of `...Precision` settings do not.

Thus enabling of plugin `convertTransform` have some risk. Will you may descript it in documentation at least short?

SethFalco:
I'd actually consider this a bug based on my findings above. Rather than update the documentation, I intend to resolve the precision issue.

You're correct, updating the `precision` settings wouldn't help due to the assumptions that are made. For some reason, that takes precedence over the parameters, which I'd also consider a bug tbh. 🤔

charlyoleg2:
I also get a corrupted image when using 'convertTransform'. I upload my files here to give you an other example of corruption:
- abc_original.svg
[...]

Issue #1363, opened 2021-02-22
[Test] Add regression testing


sk-:
As reported by #1360, `SGVO` should guarantee that at least the default settings produce a lossless optimization. On the other hand for the rest of the plugins, we should still require that the changes do not introduce visual differences that are  too noticeable.

In #1357 @TrySound started the work for adding such regression testing using the W3C SVG 1.1 Test Suite. And @sk- has been working on a similar effort using a dataset taken from Wikimedia Commons.

I think we should have both a minimal dataset for doing some regression testing on CI (as the operation is rather expensive) an an extensive collection to be used offline before releasing. (To reduce the chance of errors we could instruct contributors to run that test suite on their machines before sending a PR, it could be a checkbox in the PR template).

In my testing the files from Wikimedia commons are quite a good source as they unveil quite a few errors. Also we should use the Oxygen icon them as at least `svgcleaner` [claims](https://github.com/RazrFalcon/svgcleaner#correctness) svgo breaks many files.

For comparing the images I think [reg-suit](https://github.com/reg-viz/reg-suit) is a good alternative, as it integrates with github and has a visual ui, which makes it easy to see the differences (see these [screenshots](https://github.com/svg/svgo/pull/1357#issuecomment-783430236)).

TrySound:
I found oxygen library is quite huge. 400mb gzipped and 1.2gb unpacked. Not a good case for running on CI.

PR #1867, opened 2023-12-02
fix(convertPathData): fix some weird behavior

KTibow:
- Move converting to z to the end of the loop. This makes sure the removing useless paths thing runs before it tries to convert to z, preventing extra zs from being added.
- Round relative data before using it to get a length estimate. This fixes some cases where absolute was used instead of relative, leading to possible size increasing and data loss.
- Use error when checking if curves are convertible or not. This fixes some false negatives caused by `0.1 + 0.2`-esque situations.

See tests for examples of expected behavior

## Results
This PR is just fixing some unexpected behavior that might be already resolved with other methods. However, with Isometric Madness it subtracts 927b without multipass and 915b with.


KTibow:
@SethFalco would you mind leaving a review?

SethFalco (approved):
@KTibow Yeah sure, sorry to keep you waiting.

All the changes make sense, but the reason I wanted more time with this is because of a test like this:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 0.05 0.05">
    <path d="M 0 0 h 0.0005" stroke="black"/>
</svg>

@@@

<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 0.05 0.05">
    <path d="M0 0" stroke="black"/>
</svg>
```

This PR does not cause this behavior, it's always been there, even in v3.0.2. However, I was thinking it could be worth improving our regression testing first, which may catch this kind of thing.

But we can come back to that later, we have a ticket for it already too.

> However, with Isometric Madness it subtracts 927b without multipass and 915b with.

Plugins should execute all optimizations it would want upfront without the need of multipass.

In other words, a pipeline with a single plugin should not benefit from consequent executions.

However, I can see that this was already doing this prior to this PR, so this should be addressed separately. For now, happy to approve. 👍🏽

KTibow:
> Plugins should execute all optimizations it would want upfront without the need of multipass.

Not as in additional, I was comparing previous SVGO to current SVGO with and without multipass.

KTibow:
> This PR does not cause this behavior, it's always been there, even in v3.0.2.

Actually the default behavior wouldn't do that, it checks for stroke

SethFalco:
> Actually the default behavior wouldn't do that, it's always been [checking for stroke](https://github.com/svg/svgo/blob/8e7cb86d5ed5dec1bd5970fc33ea9efbcfd8bd16/plugins/convertPathData.js#L784).

The default behavior does remove it, I tested manually. I didn't check which optimization within `convertPathData` did it, though.

> ticket we have already

https://github.com/svg/svgo/issues/1363

> consider merging

It will be merged soon, similarly to other pull requests you've made that I reviewed.

I was just saving the merge for next time I'm actively working on SVGO and reviewing pull requests. Just leaves more time for others to chime or spot potential issues.

In general, we'll get around to reviewing/merging/triaging when we can. Keep in mind that none of the maintainers work on SVGO full-time, so the pace just varies. The occasional nudge is fine, but I'd appreciate it if you'd be more patient between interactions.

KTibow:
> The default behavior does remove it, I tested manually. I didn't check which optimization within convertPathData did it, though.

I just realized that `maybeHasStrokeAndLinecap` also required the linecap - excuse my brain fart.

> In general, we'll get around to reviewing/merging/triaging when we can. Keep in mind that none of the maintainers work on SVGO full-time, so the pace just varies. The occasional nudge is fine, but I'd appreciate it if you'd be more patient between interactions.

Right, I understand that. The motivation for "could you leave a review" was "it had been more than a week since this PR was last updated and nothing acknowledged that you had seen this at all", and the motivation for "could you merge" was a bit of confusion as there were some opposing signals (while it was an approval, it wasn't merged and it was talking about some blockers), but I'll take this into account in the future.

KTibow:
On the problem with the test you sent, I think a better approach would be to change the method for calculating the default floatPrecision. eg, if this SVG was upscaled to [size], what's the minimum precision needed so that any change is less than [real error]? This lets us fix bugs with small viewboxes and get better reductions on large viewboxes.

A precision of 3 (error of 0.001) on a 48x48 SVG shows up as an error as only 0.038px if you upscaled the SVG to fill the height of my laptop. This lets us to reduce the precision to 2, while still staying under 0.5px error.
A precision of 3 (error of 0.001) on a 0.05x0.05 SVG shows up as an error of 36px on my screen (upscaled to fill the height). This means we should increase the precision to 5 to get it down to 0.36px error.

(my laptop's display is 2736 x 1824 (267 PPI), which means pixels are pretty small.)

Thoughts on this idea? Should I open a issue/PR for it?

SethFalco:
> I just realized that `maybeHasStrokeAndLinecap` also required the linecap

No worries, we'll figure all of it out later. SVGO has done that since <=v3.0.2, so while it'd be nice to address, there's no rush compared to the myriad of other long-standing issues.

> as there were some opposing signals (while it was an approval, it wasn't merged and it was talking about some blockers), but I'll take this into account in the future.

Ahh, I see! I'll be clearer in future, sorry for the confusion. And for explicitness, I'm happy to merge this.

I was just making notes on issues I noticed while reviewing this PR that and wanted to be mindful of later.

* The issue is an actual problem, but not caused by this PR. It's likely related to how the default value for `floatPrecision` is calculated.
* The second issue is just a misunderstanding, which you've cleared up already.

Just to be clear, I had no qualms with a nudge when this went unreviewed for two weeks. While it does add pressure, I can appreciate it's frustrating when you invest your time to open a PR, and it doesn't get reviewed. I was just busy, so I kept to simpler tasks the past 1–2 weeks.

My message namely applied to the follow-up requesting a merge.

---

> I think a better approach would be to change the method for calculating the default floatPrecision

Agreed, though I'm not sure what the best approach for this is yet. We have a similar issue with other numbers in SVGO too:

* https://github.com/svg/svgo/issues/1810
* https://github.com/svg/svgo/issues/1858

I'd like if the default value was determined based on the SVG content rather than something we arbitrarily set, for example, maybe something like the highest between:

* 3
* The number of decimal places or significant digits in the `viewBox`, or if it's not present, the `height` and `width` of the closest parent.
* The number of decimal places or significant digits in the `transform` applied, if applicable.

That definition probably has flaws, so it needs more thought. It's why I was thinking that maybe it was worth improving our regression tests first.

> Should I open a issue/PR for it?

If you have an idea, contributions are welcome!

SethFalco:
The code looks good, and I ran regression tests with the larger dataset locally with no impact on mismatches. Happy to merge, and thanks for being patient!

PR #1869, opened 2023-12-03
feat(removeDeprecatedAttrs): new removeDeprecatedAttrs plugin

### LINKED ISSUE #1701: Remove deprecated `version` attribute from top `svg` element
Per MDN docs on the `<svg version>` attribute
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/version

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/version#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

This text suggests to me it is safe to remove when optimizing for fewer total characters.

jdufresne:
The new removeDeprecatedAttributes removes deprecated attributes from
the SVG document. For example, the "version" attribute on the "svg"
element is deprecated and not recommended as documented on MDN:

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/version

> Deprecated: This feature is no longer recommended. Though some
> browsers might still support it, it may have already been removed from
> the relevant web standards, may be in the process of being dropped, or
> may only be kept for compatibility purposes. Avoid using it, and
> update existing code if possible; see the compatibility table at the
> bottom of this page to guide your decision. Be aware that this feature
> may cease to work at any time.

The document:

```
<svg version="1.1" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
  <rect x="10" y="10" width="80" height="80"/>
</svg>
```

Becomes:

```
  <rect x="10" y="10" width="80" height="80"/>
</svg>
```

The plugin has a concept of "safe" and "unsafe" deprecated attributes. A
safe attribute is on that, when removed, does not alter the rendering.
An example is the "version" attribute described above. An unsafe
attribute is one that does change the rendering. By default, the plugin
only removes the safe attributes. The param "removeUnsafe" can be used
to also remove the unsafe ones.

Deprecated attributes may also appear in the SVG stylesheet as a
selector. When this occurs, the attribute won't be removed from the
document to avoid additional rendering changes.

The "xml:lang" attribute is deprecated but also has a replacement,
"lang", so this attribute is treated as a special case and is safe when
its replacement exists.

The plugin is built for easy expansion as new deprecated attributes are
discovered or announced. Simply add to ones of the "deprecated" sets in
plugins/_collections.js.

Fixes #1701

### DIFF
--- a/docs/03-plugins/remove-deprecated-attrs.mdx
+++ b/docs/03-plugins/remove-deprecated-attrs.mdx
@@ -1,0 +1,27 @@
+---
+title: Remove Deprecated Attributes
+svgo:
+  pluginId: removeDeprecatedAttrs
+  defaultPlugin: true
+  parameters:
+    removeAny:
+      description: By default, this plugin only removes safe deprecated attributes that do not change the rendered image. Enabling this will remove all deprecated attributes which may impact rendering.
+      type: boolean
+      default: false
+---
+
+Removes deprecated attributes from elements in the document.
+
+This plugin does not remove attributes from the deprecated XLink namespace. To remove them, use the [Remove XLink](/docs/plugins/remove-xlink/) plugin.
+
+## Usage
+
+<PluginUsage />
+
+## Demo
+
+<PluginDemo />
+
+## Implementation
+
+- https://github.com/svg/svgo/blob/main/plugins/removeDeprecatedAttrs.js
--- a/plugins/_collections.js
+++ b/plugins/_collections.js
@@ -376,10 +376,34 @@ export const attrsGroupsDefaults = {
 };
 
 /**
+ * @type {Record<string, { safe?: Set<string>, unsafe?: Set<string> }>}
+ * @see https://www.w3.org/TR/SVG11/intro.html#Definitions
+ */
+export const attrsGroupsDeprecated = {
+  animationAttributeTarget: { unsafe: new Set(['attributeType']) },
+  conditionalProcessing: { unsafe: new Set(['requiredFeatures']) },
+  core: { unsafe: new Set(['xml:base', 'xml:lang', 'xml:space']) },
+  presentation: {
+    unsafe: new Set([
+      'clip',
+      'color-profile',
+      'enable-background',
+      'glyph-orientation-horizontal',
+      'glyph-orientation-vertical',
+      'kerning',
+    ]),
+  },
+};
+
+/**
  * @type {Record<string, {
  *   attrsGroups: Set<string>,
  *   attrs?: Set<string>,
  *   defaults?: Record<string, string>,
+ *   deprecated?: {
+ *     safe?: Set<string>,
+ *     unsafe?: Set<string>,
+ *   },
  *   contentGroups?: Set<string>,
  *   content?: Set<string>,
  * }>}
@@ -574,6 +598,7 @@ export const elems = {
       name: 'sRGB',
       'rendering-intent': 'auto',
     },
+    deprecated: { unsafe: new Set(['name']) },
     contentGroups: new Set(['descriptive']),
   },
   cursor: {
@@ -958,6 +983,7 @@ export const elems = {
       width: '120%',
       height: '120%',
     },
+    deprecated: { unsafe: new Set(['filterRes']) },
     contentGroups: new Set(['descriptive', 'filterPrimitive']),
     content: new Set(['animate', 'set']),
   },
@@ -977,6 +1003,15 @@ export const elems = {
     defaults: {
       'horiz-origin-x': '0',
       'horiz-origin-y': '0',
+    },
+    deprecated: {
+      unsafe: new Set([
+        'horiz-origin-x',
+        'horiz-origin-y',
+        'vert-adv-y',
+        'vert-origin-x',
+        'vert-origin-y',
+      ]),
     },
     contentGroups: new Set(['descriptive']),
     content: new Set(['font-face', 'glyph', 'hkern', 'missing-glyph', 'vkern']),
@@ -1028,6 +1063,31 @@ export const elems = {
       'panose-1': '0 0 0 0 0 0 0 0 0 0',
       slope: '0',
     },
+    deprecated: {
+      unsafe: new Set([
+        'accent-height',
+        'alphabetic',
+        'ascent',
+        'bbox',
+        'cap-height',
+        'descent',
+        'hanging',
+        'ideographic',
+        'mathematical',
+        'panose-1',
+        'slope',
+        'stemh',
+        'stemv',
+        'unicode-range',
+        'units-per-em',
+        'v-alphabetic',
+        'v-hanging',
+        'v-ideographic',
+        'v-mathematical',
+        'widths',
+        'x-height',
+      ]),
+    },
     contentGroups: new Set(['descriptive']),
     content: new Set([
       // TODO: "at most one 'font-face-src' element"
@@ -1038,10 +1098,12 @@ export const elems = {
   'font-face-format': {
     attrsGroups: new Set(['core']),
     attrs: new Set(['string']),
+    deprecated: { unsafe: new Set(['string']) },
   },
   'font-face-name': {
     attrsGroups: new Set(['core']),
     attrs: new Set(['name']),
+    deprecated: { unsafe: new Set(['name']) },
   },
   'font-face-src': {
     attrsGroups: new Set(['core']),
@@ -1134,6 +1196,18 @@ export const elems = {
     defaults: {
       'arabic-form': 'initial',
     },
+    deprecated: {
+      unsafe: new Set([
+        'arabic-form',
+        'glyph-name',
+        'horiz-adv-x',
+        'orientation',
+        'unicode',
+        'vert-adv-y',
+        'vert-origin-x',
+        'vert-origin-y',
+      ]),
+    },
     contentGroups: new Set([
       'animation',
       'descriptive',
@@ -1173,6 +1247,14 @@ export const elems = {
       'vert-origin-x',
       'vert-origin-y',
     ]),
+    deprecated: {
+      unsafe: new Set([
+        'horiz-adv-x',
+        'vert-adv-y',
+        'vert-origin-x',
+        'vert-origin-y',
+      ]),
+    },
     contentGroups: new Set([
       'animation',
       'descriptive',
@@ -1236,6 +1318,7 @@ export const elems = {
   hkern: {
     attrsGroups: new Set(['core']),
     attrs: new Set(['u1', 'g1', 'u2', 'g2', 'k']),
+    deprecated: { unsafe: new Set(['g1', 'g2', 'k', 'u1', 'u2']) },
   },
   image: {
     attrsGroups: new Set([
@@ -1430,6 +1513,14 @@ export const elems = {
       'vert-origin-x',
       'vert-origin-y',
     ]),
+    deprecated: {
+      unsafe: new Set([
+        'horiz-adv-x',
+        'vert-adv-y',
+        'vert-origin-x',
+        'vert-origin-y',
+      ]),
+    },
     contentGroups: new Set([
       'animation',
       'descriptive',
@@ -1710,6 +1801,15 @@ export const elems = {
       contentScriptType: 'application/ecmascript',
       contentStyleType: 'text/css',
     },
+    deprecated: {
+      safe: new Set(['version']),
+      unsafe: new Set([
+        'baseProfile',
+        'contentScriptType',
+        'contentStyleType',
+        'zoomAndPan',
+      ]),
+    },
     contentGroups: new Set([
       'animation',
       'descriptive',
@@ -1956,11 +2056,13 @@ export const elems = {
       'viewTarget',
       'zoomAndPan',
     ]),
+    deprecated: { unsafe: new Set(['viewTarget', 'zoomAndPan']) },
     contentGroups: new Set(['descriptive']),
   },
   vkern: {
     attrsGroups: new Set(['core']),
     attrs: new Set(['u1', 'g1', 'u2', 'g2', 'k']),
+    deprecated: { unsafe: new Set(['g1', 'g2', 'k', 'u1', 'u2']) },
   },
 };
 
--- a/plugins/removeDeprecatedAttrs.js
+++ b/plugins/removeDeprecatedAttrs.js
@@ -1,0 +1,120 @@
+import * as csswhat from 'css-what';
+import { attrsGroupsDeprecated, elems } from './_collections.js';
+import { collectStylesheet } from '../lib/style.js';
+
+export const name = 'removeDeprecatedAttrs';
+export const description = 'removes deprecated attributes';
+
+/**
+ * @typedef {{ safe?: Set<string>; unsafe?: Set<string> }} DeprecatedAttrs
+ * @typedef {import('../lib/types.js').XastElement} XastElement
+ */
+
+/**
+ * @param {import('../lib/types.js').Stylesheet} stylesheet
+ * @returns {Set<string>}
+ */
+function extractAttributesInStylesheet(stylesheet) {
+  const attributesInStylesheet = new Set();
+
+  stylesheet.rules.forEach((rule) => {
+    const selectors = csswhat.parse(rule.selector);
+    selectors.forEach((subselector) => {
+      subselector.forEach((segment) => {
+        if (segment.type !== 'attribute') {
+          return;
+        }
+
+        attributesInStylesheet.add(segment.name);
+      });
+    });
+  });
+
+  return attributesInStylesheet;
+}
+
+/**
+ * @param {XastElement} node
+ * @param {DeprecatedAttrs | undefined} deprecatedAttrs
+ * @param {import('./plugins-types.js').DefaultPlugins['removeDeprecatedAttrs']} params
+ * @param {Set<string>} attributesInStylesheet
+ */
+function processAttributes(
+  node,
+  deprecatedAttrs,
+  params,
+  attributesInStylesheet,
+) {
+  if (!deprecatedAttrs) {
+    return;
+  }
+
+  if (deprecatedAttrs.safe) {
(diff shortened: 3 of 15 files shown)


jdufresne:
Thanks!

Are you suggesting they too be included in the deprecated list? I'm happy to do that if you'd like.

Any that affect rendering can be omitted as unsafe.

jdufresne:
> Yeah I would suggest adding all ones marked as deprecated by MDN that are verified to not affect rendering for multiple renderers

I pushed a new revision that includes all attributes marked deprecated on MDN.

plugins/removeDeprecatedAttrs.js
```diff
@@ -0,0 +1,29 @@
...
+      enter: (node) => {
+        const elemConfig = elems[node.name];
+        if (elemConfig) {
+          const deprecated = elemConfig.deprecated;
+          if (deprecated) {
+            deprecated.forEach((name) => {
+              delete node.attributes[name];
+            });
+          }
+        }
```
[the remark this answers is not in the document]
jdufresne: The two forms are equivalent and come down to personal preference, so I'll hold off until a maintainer weighs in before making a change here.

plugins/_collections.js
```diff
@@ -1679,6 +1680,69 @@ exports.elems = {
       contentScriptType: 'application/ecmascript',
       contentStyleType: 'text/css',
     },
+    deprecated: [
```
SethFalco, line 1683: This list is hard-coded for the `<svg>` node only, but most of these attributes aren't for the `<svg>` node at all, and were already being removed by SVGO as unknown attributes.

For example, `accent-height` is used by the `<font-face>` node, not `<svg>`.

The deprecated attributes should be split into the separate nodes they were usable in. i.e. `accent-height` in `font-face` and `viewTarget` in `view`, etc.

Your current test only passes because it happens to test on an `<svg>` node, but a test on a `<view>` node fails:

```svg
Removes deprecated viewTarget attribute from view node.

===

  <view id="one" viewBox="0 0 100 100" viewTarget=""/>
</svg>

@@@

    <view id="one" viewBox="0 0 100 100"/>
</svg>
```

plugins/removeDeprecatedAttrs.js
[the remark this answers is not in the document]
SethFalco: Personally, I try not to enforce style too much unless it's enforced by the project, i.e. the linter. Any personal qualms I have, I can just look at it when or if we do a refactor.

In general, I do much prefer early returns, though. If I were to be nitpicky, I'd favor scoping the return.

```js
if (!elemConfig) {
  return;
}
```

https://invidious.falco.fun/watch?v=CFRhGnuXG-4

I wouldn't worry too much about it, though!

SethFalco (changes requested):
Hey! Thanks for opening the PR.

The plugin makes sense, but I wouldn't make it a default.
This will likely break many SVGs in non-browser clients, so it should be opt-in.

I think it'd be cool to merge though if disabled by default, and the issue with how the deprecated attributes are stored in `_collections.js` is resolved.

jdufresne:
Thanks for the review! I believe I have addressed all feedback:

1. Removed the plugin from default
2. Dropped `xlink` attribute from the "deprecated" array
3. Switch style to early return
4. Moved the array away from "svg". As I understand the documentation, these attributes are universally deprecated, so I've created a single "deprecatedAttrs" array that applies to all elements. Let me know how you feel about this approach. If you still think it should be broken down per element, I can that path too.

docs/03-plugins/remove-deprecated-attrs.mdx
```diff
@@ -0,0 +1,22 @@
+---
+title: Remove Deprecated Attributes
+svgo:
+  pluginId: removeDeprecatedAttrs
+  defaultPlugin: true
```
SethFalco, line 5:
```suggestion
```

SethFalco, line 5:
```suggestion
```

docs/03-plugins/remove-deprecated-attrs.mdx
```diff
@@ -0,0 +1,22 @@
...
+---
+
+Remove deprecated attributes from elements in the document.
+
+This plugin does not remove the deprecated `xlink:*` attributes. To remove them, use the `removeXlink` plugin.
```
SethFalco, line 10:
```suggestion
This plugin does not remove attributes from the deprecated XLink namespace. To remove them, use the [Remove XLink](/docs/plugins/remove-xlink/) plugin.
```

Can we refer to the namespace name rather than a particular prefix? (Since it can have any prefix!)

Also it'd be nicer to link to the plugins docs.

plugins/_collections.js
```diff
@@ -2289,3 +2289,66 @@ exports.pseudoClasses = {
   userAction: ['hover', 'active', 'focus', 'focus-visible', 'focus-within'],
   functional: ['is', 'not', 'where', 'has'],
 };
+
+/**
+ * @type {string[]}
+ * @see https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute
+ */
+exports.deprecatedAttrs = [
```
SethFalco, line 2297: I'd prefer if we had these in the respective nodes, and if any of them to apply to all nodes, we can have a `deprecatedCoreAttrs` outside which is destructed into the relevant nodes.

It's because `removeUnknownsAndDefaults` already handles this for all other nodes, so I don't think there's any need to iterate through every item on _every_ node. In my opinion, it also makes more sense since it's technically only deprecated in the places it was valid before, it can't be deprecated in a node it was never supported in.

plugins/_collections.js
```diff
@@ -2289,3 +2289,66 @@ exports.pseudoClasses = {
   userAction: ['hover', 'active', 'focus', 'focus-visible', 'focus-within'],
   functional: ['is', 'not', 'where', 'has'],
 };
+
+/**
+ * @type {string[]}
+ * @see https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute
```
SethFalco, line 2295:
```suggestion
 * @see https://developer.mozilla.org/docs/Web/SVG/Attribute
```

Can we remove local specific paths from the link?
It's just so when someone who doesn't want English clicks it, they don't get English.

plugins/_collections.js
KTibow: > For example, the clip attribute should probably be migrated to the clip-path CSS property rather than removed.

This kind of setup would be unexpected for cases where the previous attribute wasn't recognized by most SVG renderers.
+ clip doesn't do the same thing as clip-path in my experience with CSS, you would need to do some complicated conversions.

SethFalco: > This kind of setup would be unexpected for cases where the previous attribute wasn't recognized by most SVG renderers, but the new one is.

While that's technically valid, I think it's safe to assume that if an SVG uses `clip`, they were intending to use the feature. Plus, that argument can be made for any modification we do, though I'll concede in cases where it really does apply to the majority of clients. We'll see after looking into each migratable attribute individually.

The rest of this will cover `clip` exclusively, since that's the only one I've looked at so far.

I've only tested `clip` and `clip-path` in two clients so far, Firefox and Eye of GNOME. Both work in Firefox, and neither work in Eye of GNOME, so in both cases rendering isn't affected.

I'll need to investigate this further, but the scenario you've presented is more like a rendering bug or unimplemented feature in certain clients, in which case this would actually work around the bug and increase compatibility.

In the worst case, if it may result in changes to rendering, it wouldn't be a default plugin anyway. It's up to users to decide whether that's worthwhile for them or not based on the warning in the documentation. (Non-default plugins ideally won't, but have more leeway to make changes that impact rendering or compatibility, as long as they're logical and adhere to whatever spec they're following, usually SVG or HTML.)

> Looks like SVG's clip only accepts "rect", which could be migrated to "inset"

I just took the alternative from MDN, but didn't check further:

> Warning: This property is deprecated. Use [clip-path](https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path) instead.
> 
> — https://developer.mozilla.org/docs/Web/SVG/Attribute/clip

Here is what I tested with:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10">
  <svg x="0" width="10" height="10" clip="rect(1, 9, 8, 2)">
    <circle cx="5" cy="5" r="4" stroke="green" />
  </svg>
</svg>
```

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10">
  <svg x="0" width="10" height="10" style="clip-path:inset(0px 1px 2px 1px)">
    <circle cx="5" cy="5" r="4" stroke="green" />
  </svg>
</svg>
```

---

This weekend I'll look into each attribute and their compatibility table individually, then I'll drop a line with what I think.

Meanwhile, other opinions welcome, especially if anyone else wants to do the same.

SethFalco:
I've looked a bit into this, and there are four properties that could be handled differently, however taking a migration path seems to be complicated for these plugins, so it might be better to shelve that idea for now.

While investigating this, I made some notes, so I'm just leaving them here for future reference.

### `clip`

Convert to the CSS `clip-path` property. As [KTibow](https://github.com/KTibow) pointed out in https://github.com/svg/svgo/pull/1869#discussion_r1419067835, this can be translated to the `inset`. However, I have found some weirdness that makes this annoying…

| Client | Supports `clip` | Supports equivalent | Impacts rendering |
|---|---|---|---|
| Eye of GNOME | ❌ | ❌ | No |
| Inkscape | ❌ | ❌ | No |
| Firefox (SVG) | ✅ | ➖ | Yes but…¹ |
| Firefox (Inline) | ✅ | ➖ | Yes but…¹ |
| Chromium (SVG) | ❌ | ✅ | Yes |
| Chromium (Inline) | ❌ | ✅ | Yes |

¹ `clip-path` appears to have a bug in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1869456

### `requiredFeatures`

If the same node doesn't have `requiredExtensions` or `systemLanguage`, and the parent is a `<switch>`, move it to the parent node and drop the `<switch>` along with the rest of its children.

### `xml:lang`

Convert to SVG `lang` attribute. It can be moved as-is. Important for accessibility.

```js
node.attributes.lang = node.attributes['xml:lang'];
delete node.attributes['xml:lang'];
```

### `xml:space`

Convert to CSS `white-space` property. But this also has some messiness…

MDN suggests using `white-space` instead. From what I can see, the equivalent would be replacing all whitespace characters with a single space (`/\s/g` → ` `), then using `white-space:pre`.

Here is the support, though:

| Client | Supports `xml:space` | Supports equivalent | Impacts rendering |
|---|---|---|---|
| Eye of GNOME | ✅ | ❌ | Yes |
| Inkscape | ❌ | ✅ | Yes |
| Firefox (SVG) | ✅ | ✅ | No |
| Firefox (Inline) | ❌ | ✅ | Yes |
| Chromium (SVG) | ❌ | ❌ | No |
| Chromium (Inline) | ❌ | ❌ | No |

---

I was thinking that migrating would be good to avoid issues with breaking modern clients… but nothing is that simple, I'll investigate those more later but as separate plugins. ^-^'

Let's keep this plugin focused on removing params, but could you please add a parameter that separates the safe and unsafe attributes? Even if we ignore migrating, we can see in the tables above that dropping some of these attributes can impact rendering on modern clients/browsers. I'd prefer if the plugin ran in a "safe mode" by default. Otherwise, it wouldn't be very practical to use in most cases.

I'd be happy to help you decide which is which in a separate comment later this week. Two clear examples are that `version` is safe to remove, while `clip` and `xml:space` are unsafe to remove.

Then by default, this plugin can perform safe operations only, and the parameter can be passed to include unsafe operations for those that want it. The docs for the `unsafe` parameter should clarify that it may impact rendering, even on modern clients and browser versions.

jdufresne:
@SethFalco I believe I have addressed the feedback in the latest revision.

The deprecated attributes are now split into safe/unsafe. By default, only safe attributes are removed but through a param, unsafe can also be removed. At this time, the only safe attribute is "version". 

Do you think this change would allow making this plugin included by default now?

Thanks for the continued reviews.

plugins/removeDeprecatedAttrs.js
```diff
@@ -0,0 +1,62 @@
...
+
+exports.name = 'removeDeprecatedAttrs';
+exports.description = 'removes deprecated attributes';
+
+/**
+ * Remove deprecated attributes.
+ *
+ * @type {import('./plugins-types').Plugin<'removeDeprecatedAttrs'>}
+ */
+exports.fn = (root, params) => {
```
SethFalco, line 13: To play it extra safe, could you please collect the stylesheet, and only remove the attribute if a selector doesn't traverse it?

For example, let's not remove `clip` if there is a CSS selector that targets `[clip]` or `[clip=…]`. Welcome to introduce a `force` param to bypass this, but I don't think there's any need.

I think the best way to approach this would be:

* Initialize the stylesheet rules with `collectStylesheet(root)`.
* Collect a collection of attribute selectors found in the stylesheet upfront. You can reference [`#includesAttrSelector`](https://github.com/svg/svgo/blob/main/lib/style.js#L297) for how to check for attributes.
* Either skip checking for the attributes in the collection above, or continue checking but on match, also verify it's not in the collection before actually deleting the attribute.

If this is confusing, please let me know and I can elaborate or just handle it for in a commit after other conversations have been resolved.

plugins/removeDeprecatedAttrs.js
```diff
@@ -0,0 +1,62 @@
...
+          deprecated.safe.forEach((name) => {
+            delete node.attributes[name];
+          });
+        }
+
+        if (removeUnsafe && deprecated.unsafe) {
+          deprecated.unsafe.forEach((name) => {
+            delete node.attributes[name];
+          });
+        }
```
SethFalco, line 58: Might be better to have a function like:

```js
function processAttributes(node, deprecatedAttrs, params) {
  if (!deprecatedAttrs) {
    return;
  }

  if (deprecatedAttrs.safe) {
    deprecatedAttrs.safe.forEach((name) => {
      delete node.attributes[name];
    });
  }

  if (params.removeUnsafe && deprecatedAttrs.unsafe) {
    deprecatedAttrs.unsafe.forEach((name) => {
      delete node.attributes[name];
    });
  }
}
```


```suggestion
        elemConfig.attrsGroups.forEach((attrsGroup) => {
          processAttributes(node, attrsGroupsDeprecated[attrsGroup], params);
        });
        
        processAttributes(node, elemConfig.deprecated, params);
```

SethFalco (changes requested):
Hey! Thanks for continuing to work on this.

I've left a few more comments, but this is looking pretty good now. Very happy to see us open a path to remove `version` by default too.

> At this time, the only safe attribute is "version".

We'll figure out if we can do any of the others post-merge. Would be nice if we could at least tackle `xml:lang` in this PR too though if `lang` is defined.

> Do you think this change would allow making this plugin included by default now?

Yeah! I think this would be good for a default plugin. Happy to have it enabled now that it's safe by default.

---

Welcome to tell me if any of my comments confuse you, if you're short on time. Happy to help out. 👍🏽

jdufresne:
Thanks again for the review. I believe I have addressed all comments including the lang attribute and CSS check. All feedback is welcome.

plugins/_collections.js
```diff
@@ -381,11 +381,35 @@ exports.attrsGroupsDefaults = {
...
 
+/**
+ * @type {Record<string, { safe?: Set<string>, unsafe?: Set<string> }>}
+ * @see https://www.w3.org/TR/SVG11/intro.html#Definitions
+ */
+exports.attrsGroupsDeprecated = {
+  animationAttributeTarget: { unsafe: new Set(['attributeType']) },
+  conditionalProcessing: { unsafe: new Set(['requiredFeatures']) },
+  core: { unsafe: new Set(['xml:base', 'xml:lang', 'xml:space']) },
+  presentation: {
```
SethFalco, line 392: Presentation attributes can be used in styles as well. We should probably be removing them too.

Could you please add a check where if the attribute being removed is a presentation attribute, also parse the `style` attribute and remove instances from there too. If the style attribute becomes empty, delete the style attribute entirely.

I worked on something like this recently in the `cleanupEnableBackground` plugin, so you can use it as a reference:

* https://github.com/svg/svgo/pull/1866/files?short_path=3741d79#diff-0805bdcf601afdd4c7674c3595fee8ea9da61677f6f36a4fdff25bd3f39fb840R45-R68

Basically, if the node defines the `style` attribute, parse the styles and if it's a `DeclarationList`, iterate the list and removing unwanted styles. Finally you can generate the new styles with `csstree.generate` and save it back, or delete the attribute if it's become empty.

Important to note that the same attribute can redundantly defined more than once in CSS, so it's worth looping through all styles and not exit early if the attribute is found.

It could be worth updating the stylesheet (`<style>` tags) too.

SethFalco (approved):
Hey! Thanks for handling all the feedback, this looks great.

Some other ways we could improve the plugin, but I'd be happy to merge it as is and file issues to tackle in future:

1. The write-up above regarding handling styles. So, to remove deprecated presentation attributes from inline styles and stylesheets.
2. Investigate if it's worth conditionally removing `enable-background` as safe by duplicating or referencing the logic from the `cleanupEnableBackground` plugin.
3. Investigate all the other deprecated attributes and see if we can identify when some of them can be safely removed.

If you are willing and have time to tackle styles in this PR, that would be very welcome, but no pressure. I'm not sure what the best approach for handling `enable-background` is, so we can figure that out later (ideas welcome), same with investigating other deprecated parameters as that will take too long, so we'll work on it iteratively.

jdufresne:
Thanks @SethFalco!

I like the idea of merging the PR as-is and then documenting the future enhancements as issues. Especially if you believe this is in a usable state.

jdufresne:
Rebased and switched to the new ESM.

SethFalco:
Sorry for the delay with merging this. This looks good, thanks for working on it and being so receptive to feedback.

PR #1870, opened 2023-12-03
chore: apply prettier to all supported file types

jdufresne:
Let prettier format Markdown, YAML, TypeScript, and JSON files. This makes contributing easier as contributors don't need to learn the project's particular formatting style and can instead let the tool handle it.

This is a followup to using prettier for JavaScript in commit a99cc08e4f6a1bc9b6d61b65e14c138195ba14c1.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -4,7 +4,6 @@ about: Create a report to help us improve
 title: ''
 labels: bug
 assignees: ''
-
 ---
 
 **Describe the bug**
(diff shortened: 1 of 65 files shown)


KTibow (approved):

SethFalco:
Hey! Thanks for proposing this, makes sense and happy to merge.

Later I'll also update the eslint/prettier dependencies and pull the configs out of `package.json` since there's no need to ship those in npm.

I'd also like to review the rules, but I'll do that in a separate pull request. I'll ping you both on that one if you want to share thoughts on it. 👍🏽

PR #1876, opened 2023-12-09
chore: Add CodeQL workflow

XhmikosR:
Feel free to adapt the triggers/schedule. This should help catch security issues like ReDoS. You can also opt to scan only for security issues; I went with `security-and-quality` since that's what I use myself.


XhmikosR:
@SethFalco Can you look at this please?

SethFalco:
Thanks for proposing this. I just had to find time to look into CodeQL, surrounding tools, and what kind of problems will be flagged. I was also wary since unless we write our own queries, it does look like a bit of a black box. (i.e. with the CLI and generic databases, every way to consume the results in a meaningful way seems to involve uploading to GitHub servers)

Overall, this makes sense though and the results reported on this branch are worth acting on. Thanks for proposing this. Tomorrow I'll evaluate the config/workflow and merge it in! 👍🏽

PR #1877, opened 2023-12-09
chore: Add dependabot config for action updates

XhmikosR:
I went with monthly so that it's not too noisy, but should still help keeping up actions up to date.


SethFalco:
Hey! This looks good, thanks for updating the action versions as well.

PR #1878, opened 2023-12-09
chore: Run CI on all branches to assist contributors

jdufresne:
Running CI on all branches allows contributors to run GitHub actions on their fork with minimal resistance. The contributor can inspect the results of their fork before opening a PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,12 +3,8 @@ env:
   FORCE_COLOR: 2
   NODE: 20
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
+  - push
+  - pull_request
 
 permissions:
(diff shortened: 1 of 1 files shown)


SethFalco:
I'm reluctant to enable `push` for all branches because of the environmental impact. It may add convenience, but it also means running the workflow for no reason most of the time. I'd rather deal with the inconvenience of running manually than promote wasting energy on redundant runs.

I actually wanted to remove the `push` action entirely in favor of linting/testing on release, but I don't have permissions to modify the repos secrets to migrate our release workflow to GitHub Actions.

The best thing that can be done is test locally, but if that doesn't work for you, you could just open a pull request to either repo to run tests, right?

jdufresne:
Makes sense. Thanks for this perspective.

PR #1879, opened 2023-12-09
fix(removeHiddenElems): handle defs better

### LINKED ISSUE #1851: Regression in 3.0.4: incorrectly removes reusable elements, but leaves references to it
**Describe the bug**
Combination of the following 3 plugins since version 3.0.4 incorrectly removes reusable element, resulting in broken icon:
```
'removeHiddenElems',
'convertPathData',
'reusePaths',
```

**To Reproduce**
Attempt to optimise the following icon using 3 plugins listed above:
```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 128 128">
	<path fill="#BD202E" d="M15.852 62.452Z"/>
	<path fill="#BD202E" d="M15.852 62.452Z" opacity=".35"/>
</svg>
```

**Expected behavior**
Expected icon that works: either empty (since paths are useless) or at least not broken.

**Actual behavior**
Actual result:
```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 128 128">
        <use href="#reuse-0"/>
        <use opacity=".35" href="#reuse-0"/>
    </svg>
```

Element with id `reuse-0` is missing.

 - SVGO Version: 3.0.4
 - NodeJs Version: 20.5.0
 - OS: macOS 14.1.1
 
**Additional context**
In version 3.0.3 same code results in:
```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 128 128">
        <defs>
                <path fill="#BD202E" d="M15.852 62.452Z" id="reuse-0"/>
        </defs>
        <use href="#reuse-0"/>
        <use opacity=".35" href="#reuse-0"/>
    </svg>
```
Plugin `removeHiddenElems` fails in 3.0.3, but at least icon is not broken.

KTibow:
Relocate logic to run in node.exit instead of checking in svg's exit:
  - remove broken uses
  Changed from checking each use to making a list of uses, to fix #1851
  - remove unused defs
  Fixed a typo so it actually removes them, and add some safeguards
  - remove empty defs
  Changed from checking each def to making a list of defs

### DIFF
--- a/plugins/removeHiddenElems.js
+++ b/plugins/removeHiddenElems.js
@@ -15,6 +15,7 @@ const {
 } = require('../lib/xast.js');
 const { collectStylesheet, computeStyle } = require('../lib/style.js');
 const { parsePathData } = require('../lib/path.js');
+const { hasScripts } = require('../lib/svgo/tools.js');
 
 const nonRendering = elemsGroups.nonRendering;
 
@@ -80,6 +81,11 @@ exports.fn = (root, params) => {
   const referencesById = new Map();
 
   /**
+   * If styles are present, we can't be sure if a definition is unused or not
+   */
+  let deoptimized = false;
+
+  /**
    * @param {XastChild} node
    * @param {XastParent} parentNode
    */
@@ -110,22 +116,6 @@ exports.fn = (root, params) => {
           return visitSkip;
         }
 
-        if (node.name == 'use') {
-          const reference = Object.keys(node.attributes).find(
-            (attr) => attr === 'href' || attr.endsWith('href')
-          );
-          const referenceValue = reference && node.attributes[reference];
-          const referenceId = referenceValue && referenceValue.slice(1);
-          if (referenceId) {
-            let refs = referencesById.get(referenceId);
-            if (!refs) {
-              refs = [];
-              referencesById.set(referenceId, refs);
-            }
-            refs.push({ node, parentNode });
-          }
-        }
-
         const computedStyle = computeStyle(stylesheet, node);
         // opacity="0"
         //
@@ -145,6 +135,32 @@ exports.fn = (root, params) => {
   return {
     element: {
       enter: (node, parentNode) => {
+        if (
+          (node.name === 'style' && node.children.length !== 0) ||
+          hasScripts(node)
+        ) {
+          deoptimized = true;
+          return;
+        }
+
+        // Store uses so that they can be removed if broken
+        // and used to determine if a definition is unused
+        if (node.name == 'use') {
+          const reference = Object.keys(node.attributes).find(
+            (attr) => attr === 'href' || attr.endsWith('href')
+          );
+          const referenceValue = reference && node.attributes[reference];
+          const referenceId = referenceValue && referenceValue.slice(1);
+          if (referenceId) {
+            let refs = referencesById.get(referenceId);
+            if (!refs) {
+              refs = [];
+              referencesById.set(referenceId, refs);
+            }
+            refs.push({ node, parentNode });
+          }
+        }
+
         // Removes hidden elements
         // https://www.w3schools.com/cssref/pr_class_visibility.asp
         const computedStyle = computeStyle(stylesheet, node);
@@ -393,17 +409,21 @@ exports.fn = (root, params) => {
         }
 
         // Remove definitions that are unused
-        for (const [
-          nonRenderedNode,
-          nonRenderedParent,
-        ] of nonRenderedNodes.entries()) {
-          const selector = referencesProps
-            .map((attr) => `[${attr}="url(#${nonRenderedNode.attributes.id})"]`)
-            .join(',');
+        if (!deoptimized) {
+          for (const [
+            nonRenderedNode,
+            nonRenderedParent,
+          ] of nonRenderedNodes.entries()) {
+            const id = nonRenderedNode.attributes.id;
+            const selector = referencesProps
+              .map((attr) => `[${attr}="url(#${id})"]`)
+              .concat(`[href="#${id}"]`, `[xlink\\:href="#${id}"]`)
+              .join(',');
 
-          const element = querySelector(root, selector);
-          if (element == null) {
-            detachNodeFromParent(nonRenderedNode, nonRenderedParent);
+            const element = querySelector(root, selector);
+            if (element == null) {
+              detachNodeFromParent(nonRenderedNode, nonRenderedParent);
+            }
           }
         }
       },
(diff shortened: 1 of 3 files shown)


KTibow:
Okay so turns out that the unused definitions logic was just always broken, so I'm having to fix the existing implementation

plugins/removeHiddenElems.js
```diff
@@ -123,6 +135,32 @@ exports.fn = (root, params) => {
...
       enter: (node, parentNode) => {
+        if (
+          (node.name === 'style' && node.children.length !== 0) ||
+          hasScripts(node)
+        ) {
+          deoptimized = true;
+          return;
+        }
+
+        // Store uses so that they can be removed if broken
```
SethFalco, line 146: Mind dropping the new comments?

Comments that just explain what the code is doing are redundant, this also applies to most of the other comments in the project.

The code is clear enough as it is. If it's really so complicated that it needs an explanation, it should be in its own function and get a JSDoc, or alternatively may need a refactor.

plugins/removeHiddenElems.js
```diff
@@ -356,37 +394,35 @@ exports.fn = (root, params) => {
...
+              detachNodeFromParent(node, parentNode);
+            }
           }
-
-          return;
         }
 
-        if (node.name === 'svg' && parentNode.type === 'root') {
+        // Remove definitions that are unused
+        if (!deoptimized) {
```
SethFalco, line 412: Could you provide an example/test where de-optimizing for styles makes sense?

(This section needs a refactor in general, we should be collecting the relevant data while visiting instead of CSS selecting at the end. I can handle this later, since it's out-of-scope for this PR, though.)

SethFalco (commented):
Hey! Thanks for submitting a PR.

I've left a few comments, but the proposed fix makes sense. There are just a few issues with the implementation.

Can you please also add a test that reproduces the issue you're fixing.

plugins/removeHiddenElems.js
```diff
@@ -123,6 +135,32 @@ exports.fn = (root, params) => {
...
+        ) {
+          deoptimized = true;
+          return;
+        }
+
+        // Store uses so that they can be removed if broken
+        // and used to determine if a definition is unused
+        if (node.name == 'use') {
+          const reference = Object.keys(node.attributes).find(
+            (attr) => attr === 'href' || attr.endsWith('href')
```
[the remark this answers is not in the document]
KTibow: typo :sweat_smile:

plugins/removeHiddenElems.js
KTibow: Something like this
```xml
<svg>
  <defs>
    <linearGradient id="a">
      ...
    </linearGradient>
  </defs>
  <rect width="128" height="128></rect>
  <style> rect { fill: url(#a); } </style>
</svg>
```

SethFalco: Make sense! I forgot definitions could be referenced from the style element.

I'd like if we could remove the exception on styles later, but this is fine for now. Later I can work on a PR to check the stylesheet for references to the ID as well.

KTibow:
Now I'm running into the problem where it has a different behavior in multipass than singlepass, so it fails no matter what.
Context: in this
```xml
    <defs>
        <linearGradient id="a">
        </linearGradient>
    </defs>
</svg>
```
singlepass converts it to this, because the empty defs remover runs before the final pass
```xml
    <defs />
</svg>
```
but multipass converts it to this
```xml
<svg xmlns="http://www.w3.org/2000/svg" />
```

What's the most reasonable solution to this? Is there an elegant way to run the empty defs remover after the final pass?

SethFalco:
> Is there an elegant way to run the empty defs remover after the final pass?

Perhaps the best approach would be to take from here:

https://github.com/svg/svgo/blob/3dc2f6fef7f67596eab64b0a98cfa68872e5609c/plugins/convertOneStopGradients.js#L36

So whenever we remove a node from a `<defs>`, store the effected `<defs`> node.
Then in `root` you can iterate these and check if they're empty there.

KTibow:
While there's more to address I'm off to lunch, I'll fix this soon

plugins/removeHiddenElems.js
```diff
@@ -74,6 +75,21 @@ exports.fn = (root, params) => {
    */
   const removedDefIds = new Set();
 
+  /**
+   * @type {Map<XastElement, XastParent>}
+   */
+  const allDefs = new Map();
+
+  /**
+   * @type {Map<string, Array<{node: XastElement, parentNode: XastParent }>>}
```
SethFalco, line 84:
```suggestion
   * @type {Map<string, Array<{node: XastElement, parentNode: XastParent}>>}
```

```suggestion
   * @type {Map<string, Array<{ node: XastElement, parentNode: XastParent }>>}
```

Very petty, and not sure if the linter cares, but could we do one of these? ^-^'

plugins/removeHiddenElems.js
```diff
@@ -350,46 +393,44 @@ exports.fn = (root, params) => {
...
-                attrValue.slice(attrValue.indexOf('#') + 1).trim()
-              )
-          );
-
-          if (referencesRemovedDef) {
-            detachNodeFromParent(node, parentNode);
+    },
+    root: {
+      exit: () => {
+        // Remove uses of deleted definitions
```
SethFalco, line 399:
```suggestion
```

Sorry for not being explicit, could we remove these comments too?
The code is all very clear already.

plugins/removeHiddenElems.js
```diff
@@ -350,46 +393,44 @@ exports.fn = (root, params) => {
...
+            for (const { node, parentNode } of refs) {
+              detachNodeFromParent(node, parentNode);
+            }
           }
-
-          return;
         }
 
-        if (node.name === 'svg' && parentNode.type === 'root') {
+        // Remove definitions that are unused
```
SethFalco, line 409:
```suggestion
```

plugins/removeHiddenElems.js
```diff
@@ -350,46 +393,44 @@ exports.fn = (root, params) => {
...
 
             const element = querySelector(root, selector);
             if (element == null) {
-              detachNodeFromParent(node, nonRenderedParent);
+              detachNodeFromParent(nonRenderedNode, nonRenderedParent);
             }
           }
         }
+
+        // Remove empty defs
```
SethFalco, line 428:
```suggestion
```

SethFalco (commented):
This looks good now! And thanks for resolving the other issues you found.
Just left a few _very petty_ notes, but happy to merge overall.

SethFalco:
Great! Thanks for resolving the issue, and improving the plugin in general.

Issue #1696, opened 2022-09-23
[cleanupIDs] `symbol` is incorrectly removed when its ID is only used by URL-encoded reference


ctrlcctrlv:
**Describe the bug**
 `symbol` is incorrectly removed when its ID is only used by URL-encoded reference.

Example files: https://gist.github.com/ctrlcctrlv/a985bd9fdf31ecc7e414ab29d49576ae

**To Reproduce**
```bash
svgo --config svgo.config.js -i working.svg -o working_o.svg
svgo --config svgo.config.js -i broken.svg -o broken_o.svg
```
Observe missing elements:
![image](https://user-images.githubusercontent.com/838783/191995306-78441cf7-fd5f-450b-9588-b9157cab7e4c.png)

**Expected behavior**
No elements to be missing:
![image](https://user-images.githubusercontent.com/838783/191995849-77ba0dbd-22f3-4c6e-a457-a180cbd5b391.png)

**Desktop (please complete the following information):**
 - SVGO Version 2.7.0
 - NodeJs Version v12.22.9
 - OS: Pop! OS 22.04

# Code issue

You cannot just do a basic string comparison on hyperlink references. This will never work how its implementor intended.

Direct `git blame` is to:

https://github.com/svg/svgo/blob/a9834efa1603198ae606765292215462ae193d0d/plugins/cleanupIDs.js#L16

By @TrySound.

However this bug dates back to original implementation of the plugin, making it ten years old. 40ab77680a9dc4fe6b848fc576995c86c6c34f5d by @deepsweet.

PR #1880, opened 2023-12-10
fix(cleanupIds): handle uri encoded references

SethFalco:
References to IDs in `href` or `url()` references are URIs, and so should be URI encoded/decoded.

We'll have to review and update all usage in SVGO for this, but for now, I've resolved the immediate issue found in Cleanup IDs.

## Related

* Fixes https://github.com/svg/svgo/issues/1696

### DIFF
--- a/plugins/cleanupIds.js
+++ b/plugins/cleanupIds.js
@@ -215,10 +215,11 @@ exports.fn = (_root, params) => {
               }
             }
             for (const id of ids) {
-              let refs = referencesById.get(id);
+              const decodedId = decodeURI(id);
+              let refs = referencesById.get(decodedId);
               if (refs == null) {
                 refs = [];
-                referencesById.set(id, refs);
+                referencesById.set(decodedId, refs);
               }
               refs.push({ element: node, name });
             }
@@ -261,7 +262,7 @@ exports.fn = (_root, params) => {
                 if (value.includes('#')) {
                   // replace id in href and url()
                   element.attributes[name] = value.replace(
-                    `#${id}`,
+                    `#${encodeURI(id)}`,
                     `#${currentIdString}`
                   );
                 } else {
--- a/test/plugins/cleanupIds.25.svg
+++ b/test/plugins/cleanupIds.25.svg
@@ -1,0 +1,21 @@
+Should handle non-ASCII IDs and resolve URI encoded references.
+
+See: https://github.com/svg/svgo/issues/1696
+
+===
+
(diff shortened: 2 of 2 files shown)


ctrlcctrlv:
Awesome, thanks! :heart_eyes:

simonhermann:
Hi @SethFalco 
we're using SVGO to optimize svgs only containing a `<symbol>`. Before 3.1, this worked as expected, with 3.1+, SVGO only returns an empty svg element. I guess it is related to this change.
Is this a bug or am I missing something?

our svgo.config.js
```
// @ts-check
/** @type {import('svgo').Config} */
module.exports = {
	plugins: [
		{
			name: 'preset-default',
			params: {
				overrides: {
					cleanupIds: false, // prevent removing unused symbols
					removeViewBox: false,
				},
			},
		},
		{
			name: 'sortAttrs',
			params: {
				xmlnsOrder: 'alphabetical',
			},
		},
	],
};
```

SethFalco:
@simonhermann

Hey! Sorry about that. Would you mind sharing an example with me, so I can review it in more detail?

In general, an SVG with only symbols should have always become empty after running through SVGO's default preset, as SVG sprites aren't a use-case the default preset is designed to handle.

I'm not dismissing your issue ofc, we'll get this addressed! Sorry for effectively releasing a breaking change without warning. We'll improve our documentation and handling of SVG sprites for the next minor release.

I'm not sure if it's caused by this PR in your case, but the problem you've raised was ultimately reported in another issue. The author closed it, but it's on my radar, and I'll publish a solution this weekend:

* https://github.com/svg/svgo/issues/1962

The default behavior will remain the same, but we'll add a preset specific to SVG sprites and add documentation for how to use that preset.

Meanwhile, I recommend you give that issue a read. You're also welcome to chime in if you have any specific thoughts on what should be in or out of an SVG sprite preset. No pressure though!

Let me know if you have any problems.

simonhermann:
Hi @SethFalco, thanks for your quick reply!
Nothing to apologize! From your link I see now `removeHiddenElems: false` makes the difference, I just assumed, setting `cleanupIds: false` was enough, since it was, before 3.1.

Maybe you could enhance the [docs on removeHiddenElems](https://svgo.dev/docs/plugins/remove-hidden-elems/) to include an example with `<symbol>`? 
I tried [searching the docs](https://svgo.dev/search?q=symbol), but did not find what I need to do to perserve unused symbols, but of course now I realise they are obviously hidden elements. 

Thank you for your support and the awesome work!

PR #1883, opened 2023-12-10
Tidy up eslint config and usage

jdufresne:
- Include *.mjs and *.cjs files in the eslint configuration. These files exist in the test suite.

- Remove eslint-disable comments that aren't silencing a error.


SethFalco:
Hey! Thanks for cleaning this up for us!

PR #1886, opened 2023-12-15
chore: remove dangling commas

SethFalco:
I was reviewing the changes that were applied after the related pull requests, and want to propose changing one setting.

We had `trailingComma` enabled before (default), but _in my opinion_ it just looks messy, especially in the README and documentation. ^-^'

Would anyone have any qualms with us disabling it?

## Related

Follow up to the following PRs:

* https://github.com/svg/svgo/pull/1870
* https://github.com/svg/svgo/pull/1883
* https://github.com/svg/svgo/pull/1885


SethFalco:
CC @jdufresne @KTibow as you've participated in the other lint PRs. Welcome to chime in if you're interested.

jdufresne:
FWIW, I prefer the dangling comma. When adding or removing the last line of a multi-line list, with a dangling comma the diff will  show exactly one line changed, but without the dangling comma, it forces two lines to change: the new line and then an adjustment for the last comma. The diff shows an unrelated line change to adjust the syntax. This is probably the main reason they are enabled by default in these tools.

Additionally, with opinionated style tools such as prettier I tend to prefer just accepting the the defaults even if they aren't a perfect match to my style. It requires less configuration and less discussion about less important details. The end result better matches the wider JavaScript community which more often accepts prettier defaults. So my proposal would be to remove `.prettierrc` entirely rather then adding to it.

All that said, the decision isn't important to me, so as you are the maintainer I obviously defer the decision to you, just providing an opinion.

SethFalco:
I'll concede, thanks to both of you for chiming in!

PR #1764, opened 2023-04-25
Improve mergePaths performance on large files

mozzie:
Removed unneeded calls from mergePaths plugin.
1) Instead of calling detachNodeFromParent, nodes are tagged and removed after each node has been looped through. 
2) Eliminated unneeded js2path and path2js conversion by keeping the current path in memory for the duration of the loop.

PR #1892, opened 2023-12-20
feat(convertPathData): convert c to q

### LINKED ISSUE #1891: Convert c to q
Sometimes `c` can be converted to `q`. This would be more compact.

This is done by drawing two lines, one from the start to the start handle, and one from the end to the end handle. Then intersect them to find the quadratic handle.

**Describe alternatives you've considered**
Implement this somewhere that isn't SVGO? not sure

**Additional context**
Read: https://yuri.is/splines/static/QuadraticApproximation.pdf
This is not always possible and I'm not sure how to check when it is possible yet

KTibow:
Some cubic beziers are quadratic beziers in disguise! They may have been degree elevated or just happen to be approximatable. Whatever the case is, we can find a smaller, faster representation as a quadratic bezier.
See https://stackoverflow.com/questions/2009160/how-do-i-convert-the-2-control-points-of-a-cubic-curve-to-the-single-control-poi for a better explanation.

## Results
Closes #1891
Compared to current SVGO, removes 6452 bytes on Isometric Madness w/o multipass and 6500 with.
This would probably help *some* icon sets a lot, haven't tested the effects yet though.


plugins/convertPathData.js
```diff
@@ -665,6 +668,44 @@ function filters(
...
+          // @ts-ignore
+          1.5 * (item.base[0] + data[2]) - 0.5 * (item.base[0] + data[4]);
+        if (Math.abs(x1 - x2) < error * 2) {
+          const y1 =
+            // @ts-ignore
+            1.5 * (item.base[1] + data[1]) - 0.5 * item.base[1];
+          const y2 =
+            // @ts-ignore
+            1.5 * (item.base[1] + data[3]) - 0.5 * (item.base[1] + data[5]);
+          if (Math.abs(y1 - y2) < error * 2) {
```
KTibow, line 687: The error can be twice as big because a handle always has half the effect on a curve, due to how curves are linear interpolations of interpolations.

SethFalco:
> while I know that having this many PRs might stress the maintainers out

Open pull requests are not stressful.

> it's annoying when they don't get merged for me too

… so long as contributors are patient and understand that it takes time to review them, or that they may be other blockers preventing the merge.

---

1. Maintainers are busy, they can't look at pull requests every week. If you pay attention to the date, it's also nearing Christmas, where most people are taking time off. Be more considerate.
2. I have actually been working on SVGO _despite the above_, however as I said to you in #1867, the regression testing needs a revamp.

I'm currently working on SVGO/testing performance and adding oxygen-icons to our test suite, which we're currently breaking over 200 icons with the default preset. We need this sorted out before potentially breaking more vectors.

I will make more progress this weekend and hopefully open a PR to track the status of this. Once that's up, SVGO will need to prioritize the relevant bugs, and improve performance so that it's CI-friendly. Then we can be more confident in merging other fixes and SVG optimizations.

We've discussed this before, and I ask kindly that you don't attempt to pressure maintainers or vent in pull requests. While triaging tickets and pull requests is important, as a maintainer, my priority is the health and stability of the project, not just reviewing pull requests quickly.

KTibow:
You're right, I wasn't being very considerate. Sorry about that.

This break is a time where it's easier for me to contribute, but it sounds like others might be taking other breaks. And while I like it when things move fast, it's unrealistic for me to try to force others to move at the same speed. Moving forward, I won't bother you with messages like that one again, but I'll keep the PRs coming if they don't bother you.

SethFalco:
> Sorry about that.

Thank you, I appreciate the response!

> but I'll keep the PRs coming if they don't bother you.

Sounds good! And just to be clear, your pull requests are very valuable, as we've seen each one continued to make the vectors smaller. SVGO just also has a reputation for being one of the more buggy optimizers, which I'd like to resolve.

> *Also if there are any weird cases of icons breaking, I'd be happy to investigate them.

That would be appreciated. Ideally, this weekend I'll have a PR open that adds oxygen-icons to our CI pipeline with a list of failing tests and the time it takes to run for me and in CI.

The goal will be to try to fix all broken tests, which is hopefully mostly repeats of a small subset of issues rather than many unique bugs. Then we need to try to improve performance. Even in my environment, it takes 20+ minutes to run, and CI is definitely going to be longer.

Some thoughts I've had:

* We could improve use of threads in our test suite, or SVGO in general.
* We should profile SVGO to identify which plugins take longest to run.
* We might be able to store and retrieve the test suites (W3C and oxygen-icons) more efficiently. For example, we could make a repo that publishes a Docker image that has the SVGs we want already. The image would still have to be fetched and unpacked, but it might still be faster than downloading from W3C and KDE servers and extracting in the runner.
* We could split regression tests for W3C and oxygen icons into two separate actions so they can run concurrently.
* We could consider rewriting the test suite with a more performant stack like Rust. However, I'd like to treat this as a last resort since maintainers and contributors will be more comfortable with Node.js. I'm also unsure if spawning a process from a binary would really be much faster than using Node.js which shares the process.

We do have pull requests that improve performance already, for example, I tested #1764 locally a while back, which is a significant improvement. I just didn't merge yet, as I wanted to rebase and refactor it, particularly to store a list/map of items to remove rather than set `_deleteAfterwards`.

KTibow:
> Even in my environment, it takes 20+ minutes to run, and CI is definitely going to be longer.

Okay, I have some questions:
- This sounds like regression testing. Are you using a testing library? Are you still using Playwright?
- You're referring to https://github.com/KDE/oxygen-icons, right? Are you just testing it on all SVGs contained within?

SethFalco:
> This sounds like regression testing. Are you using a testing library? Are you still using Playwright?

Yeah, we already have regression testing setup with the W3C Test Suite, so it's just expanding that to include Oxygen Icons. It may be worth adding others later, but not yet.

Since I'm continuing with the setup we already have, it's still using Playwright. Happy to hear ideas, though!

The only _preference_ is that we keep the repo and contribution workflow to Node.js and ensure tests are easy to run across operating systems.

> You're referring to https://github.com/KDE/oxygen-icons, right? Are you just testing it on all SVGs contained within?

Correct! And yeah, all SVGs in the `scalable/` directory.

---

I have regression tests working already. I was still evaluating if I'm happy with it, but I'll push and share my proposal already.

I've done nothing to optimize SVGO or our test runner itself yet, just evaluating how to fetch the test suites. My proposal is that we create `svg/svgo-test-suite`, a repo that hosts a GitHub Action that creates our test suite from multiple sources, and we fetch that in the existing test runner we have.

The benefits:

* We can fetch, remove files, and preprocess them upfront and reuse that archive across all pipelines.
  * We can omit unnecessary files so the SVGO pipeline never fetches them.
  * SVGZ files will already be converted to SVG.
  * We can repack the SVGs in the archive format best fit for our use case. (Currently still using `.tar.gz`, but I'm exploring what's the best archive format we can use that's easy to decompress cross-platform, preferably in Node.js.)

GitHub Action: https://github.com/SethFalco/svgo-test-suite/blob/main/.github/workflows/test.yml  
Resulting archive so far: https://sethfalco.github.io/svgo-test-suite/svgo-test-suite.tar.gz (377 MB)  
A test run: https://github.com/SethFalco/svgo/pull/1 (25 minutes)

KTibow:
I took a look at one of the fails. I think it was caused by a doubly-compressed SVG file, not SVGO.

SethFalco:
> I took a look at one of the fails. It might've been a doubly-compressed file? Not sure. I would suggest adding a sanity check, but SVGO should error out if fed invalid input, so I don't think that's needed.

Can you elaborate on what you mean here? I don't follow.

I've just checked two mismatching files now, and both show bugs in SVGO.

## `oxygen-icons-5.113.0/scalable/actions/small/22x22/mixer-digital.svg`

Before:

Screen Shot 2023-12-22 at 09 41 14

After:

Screen Shot 2023-12-22 at 09 40 41

MRE:

```xml
<svg xmlns="http://www.w3.org/2000/svg" width="128" height="128" viewBox="0 0 144 144">
  <linearGradient id="a" x1="40.7267" y1="93.091404" x2="81.4534" y2="93.091404" gradientUnits="userSpaceOnUse" gradientTransform="matrix(0.81951526,0,0,1.0312499,-119.11723,-164.7268)">
    <stop style="stop-color:#000000" offset="0"/>
    <stop offset="0.71665221" style="stop-color:#bdbdbd"/>
    <stop style="stop-color:#000000" offset="1"/>
  </linearGradient>
  <rect style="fill:url(#a);fill-opacity:1;stroke:none" width="32.726814" height="71.998985" x="-85.091736" y="-104.7258" transform="matrix(0,-1,-1,0,0,0)" />
</svg>
```

```js
module.exports = {
  plugins: [
    "convertTransform",
  ]
}
```

## `oxygen-icons-5.113.0/scalable/actions/small/22x22/mixer-digital.svg`

Before:

Screen Shot 2023-12-22 at 10 03 59

After:

Screen Shot 2023-12-22 at 10 04 20

MRE:

```xml
<svg xmlns="http://www.w3.org/2000/svg" width="32px" height="32px">
  <g transform="matrix(-0.35122607,0,0,-0.35775914,48.305763,353.01858)">
    <rect ry="1.9313376" rx="1.8960693" y="-91.979965" x="-969.97821" height="17.083014" width="16.771061" style="fill:#4d2600;fill-opacity:1" transform="matrix(0,-1,-1,0,0,0)" />
  </g>
</svg>
```

```js
module.exports = {
  plugins: [
    "convertTransform",
  ]
}
```

KTibow:
`svgs/oxygen-icons-5.113.0/scalable/mimetypes/small/64x64/image-x-adobe-dng.svg`
Input
image-x-adobe-dng
Output
out
I can't tell a difference. Processed with default optimize settings on upstream.

KTibow:
I looked at the examples you provided. While I don't understand matrixes enough to tell you why it's happening, I think it's an error related to `scaleBefore`. Frederic Wang's website puts the `scale` for `matrix(0 -1 -1 0 0 0)` inside/after, like `rotate(-90) scale(1,-1)`, but we set scaleBefore to false and put the scale outside/first.
\* The fact that "before" is vague makes things more confusing. Is "before" defined as coming first in the string, or as applying first to the object as it's innermost?

SethFalco:
>  I can't tell a difference. Processed with default optimize settings on upstream.

I think that's legit, it's barely perceivable.
Here is a diff from the regression testing (we should document that so it's clearer, sorry!):

image

There are two differences:

* The outline of the bottom-right image is slightly more solid in the original. (Not detected as a change, though.)
* The white area above the bottom-right image changes from `#f5f4f4` to `#f5f5f5`.

It fails with a combination of `cleanupIds` and `collapseGroups`.

```js
module.exports = {
  plugins: [
    "cleanupIds",
    "collapseGroups",
  ]
}
```

> * The fact that "before" is vague makes things more confusing. Is "before" defined as coming first in the string, or as applying first to the object as it's innermost?

Unfortunately, I'm not well-versed enough here either. Honestly, most of my time reviewing issues here is spending time reading the spec, documentation, or math than actual code. ^-^'

SethFalco (commented):
I think the math can be simplified, which is mentioned in the Stack Overflow post as well, but otherwise this looks great. Nice idea!

See any issues with the suggestion?

SethFalco (approved):
You can ignore the failing tests, you probably noticed, but we have an intermittent issue atm. Will look into that later, nothing to do with this PR.

Thanks for the contribution!

PR #1897, opened 2023-12-22
chore: speedier regression testing

KTibow:
Goes from 76 seconds to 61 seconds in my tests.
Two main changes:
- instead of chunking the array, use workers that subtract as needed
Bonus: this means that they go in alphabetical order
- serve SVGs, optimized or not, on demand instead of warming them up
This means no optimizing skipped tests
Admittedly won't be as useful once the PR that moves skipping tests into the extractor gets merged, but hey, it'll still decrease initial startup time

*I was going to make it even faster with `node-canvas`, but the w3 tests are a bit eccentric and fail to render properly there, and I don't know of any other convenient rendering libs

### DIFF
--- a/test/regression.js
+++ b/test/regression.js
@@ -125,10 +125,17 @@ const height = 720;
     // setup server
     const server = http.createServer(async (req, res) => {
       const name = req.url.slice(req.url.indexOf('/', 1));
-      const file = await fs.promises.readFile(
-        path.join(fixturesDir, name),
-        'utf-8',
-      );
+      let file;
+      try {
+        file = await fs.promises.readFile(
+          path.join(fixturesDir, name),
+          'utf-8',
+        );
+      } catch (error) {
+        res.statusCode = 404;
+        res.end();
+        return;
+      }
 
       if (req.url.startsWith('/original/')) {
         res.setHeader('Content-Type', 'image/svg+xml');
--- a/test/regression.js
+++ b/test/regression.js
@@ -3,6 +3,7 @@
 const fs = require('fs');
 const path = require('path');
 const http = require('http');
+const os = require('os');
 const { chromium } = require('playwright');
 const { PNG } = require('pngjs');
 const pixelmatch = require('pixelmatch');
@@ -12,6 +13,7 @@ const runTests = async ({ list }) => {
   let skipped = 0;
   let mismatched = 0;
   let passed = 0;
+  list.reverse();
   console.info('Start browser...');
   const processFile = async (page, name) => {
     if (
@@ -79,7 +81,7 @@ const runTests = async ({ list }) => {
   const worker = async () => {
     let item;
     const page = await context.newPage();
-    while ((item = list.shift())) {
+    while ((item = list.pop())) {
       await processFile(page, item);
     }
     await page.close();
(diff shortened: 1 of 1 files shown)


test/regression.js
```diff
@@ -124,54 +120,52 @@ const readdirRecursive = async (absolute, relative = '') => {
...
-        const name = req.url.slice('/original/'.length);
-        if (originalFiles.has(name)) {
-          res.setHeader('Content-Type', 'image/svg+xml');
-          res.end(originalFiles.get(name));
+    const server = http.createServer(async (req, res) => {
+      const name = req.url.startsWith('/original/')
+        ? req.url.slice('/original/'.length)
+        : req.url.startsWith('/optimized/')
+          ? req.url.slice('/optimized/'.length)
+          : undefined;
```
SethFalco, line 136:
```suggestion
      const url = req.url;
      const name = url.slice(url.indexOf("/", 1));
```

This works for both cases, and we don't have to worry about `undefined` since this is only used internally.

SethFalco (commented):
Overall, this looks great. Thank you!

I have a few ideas on how we can refactor this too, but I'll tackle them in a separate PR once this is merged.

test/regression.js
SethFalco: The logic you replaced it with isn't the same as my suggestion!

The filename can have path separators in it, so it's important to grab everything after the second separator, not just split.

The reason the tests pass now is because of the logic later that returns a 404 with empty content if we fail to read the file. (Which is bizarre in my opinion, so let's get rid of that.)

```suggestion
      const name = req.url.split('/')[2];
```

test/regression.js
```diff
@@ -124,57 +115,48 @@ const readdirRecursive = async (absolute, relative = '') => {
...
+          'utf-8',
+        );
       } catch (error) {
-        console.error(error);
-        console.error(`File: ${name}`);
-        failed += 1;
+        res.statusCode = 404;
+        res.end();
+        return;
       }
```
SethFalco, line 138: Could you replace all of this with just:

```js
const file = await fs.promises.readFile(
  path.join(fixturesDir, name),
  'utf-8',
);
```

If it fails to read the file, we should know about it. Not silently respond with a blank page. I'd expect this to crash the tests.

SethFalco:
I see… that would explain that logic. We had the fallback to empty response because some of the SVGs try to load other resources, and we want to ignore them.

Perhaps you can just put that back, now that the issue with the `name` variable is fixed. Sorry for the bad call there!

SethFalco (approved):

SethFalco:
I'm responding to this comment here, since it's more to do with our usage than the bug report itself:

https://github.com/microsoft/playwright/issues/28778#issuecomment-1868175086

I initially didn't try it deliberately, since inlining an SVG and loading it via an `` tag are different. While it would be nice to test both, figured it'd be best to stick to inline only for now.

However, I did just try it now and it does work… kind of. Failing tests went from 262 to 1,000+, and it's much slower. Possibly because of the URI encoding to make it work in a URL format, plus the browser has to load more data since the encoding makes the SVG fat.

I'll come back to that later, but for now, we can keep the HTTP server. 🤔

PR #1900, opened 2023-12-24
refactor: improve performance of stringifyPathData

SethFalco:
Refactors `#stringifyPathData` and the functions it calls, which now runs notably faster.
The scenarios I measured in microbenchmarks ran in half the time.

* When stringifying, return both the numeric and string representation so we can perform checks with math instead of string parsing.
* When stringifying path data, use a single loop instead of creating an intermediate array and looping through that after.

### Metrics

Time to run the full SVGO Test Suite proposed in https://github.com/SethFalco/svgo/pull/1.

| Environment | Time |
|---|---|
| GitHub Actions | 20.5 minutes |
| My personal laptop (4 cores, 8 threads) | 15.0 minutes |

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -243,15 +243,21 @@ const parsePathData = (string) => {
 exports.parsePathData = parsePathData;
 
 /**
- * @type {(number: number, precision?: number) => string}
+ * @type {(number: number, precision?: number) => {
+ *   roundedStr: string,
+ *   rounded: number
+ * }}
  */
-const stringifyNumber = (number, precision) => {
+const roundAndStringify = (number, precision) => {
   if (precision != null) {
     const ratio = 10 ** precision;
     number = Math.round(number * ratio) / ratio;
   }
-  // remove zero whole from decimal number
-  return removeLeadingZero(number);
+
+  return {
+    roundedStr: removeLeadingZero(number),
+    rounded: number,
+  };
 };
 
 /**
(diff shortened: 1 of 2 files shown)


lib/path.js
```diff
@@ -243,15 +243,21 @@ const parsePathData = (string) => {
...
-const stringifyNumber = (number, precision) => {
+const roundAndStringify = (number, precision) => {
   if (precision != null) {
     const ratio = 10 ** precision;
     number = Math.round(number * ratio) / ratio;
   }
-  // remove zero whole from decimal number
-  return removeLeadingZero(number);
+
+  return {
```
KTibow, line 257: Wonder if this makes the RAM usage increase

SethFalco: I'm not actually that experienced with memory profiling, but figured now's a good time to learn. I've taken a diff of two heap snapshots taken at the start and end of the `stringifyPathData` function.

I'm not sure what to think of `(object shape)` and `(compiled code)`, as these are managed by Node.js internally.

Excluding those:

* `(string)` - increase due to property names `rounded` and `roundedStr`.
* `Object`/`(array)`/`Array` - decrease since we scrapped the `combined` array.

Based on the objects that we can clearly trace to source code, we're using less memory. However, including `(object shape)` and `(compiled code)`, we're using more memory. I need to learn more about analyzing heap snapshots to know if further action should be taken on those.

## Before

image

## After

image

PR #1901, opened 2023-12-24
chore: Remove unused argument "input" in calls to optmize()

jdufresne:


SethFalco:
Thanks for spotting that! 👍🏽

By the way, I'll take a look at #1869 soon! Sorry for putting it off, I got caught up with trying to improve our regression testing so I've neglected other issues a bit. ^-^'

I'll aim for Friday (29 December) at the latest if I still don't have improved regression testing merged by then.

Issue #1429, opened 2021-03-15
SVGO v4


TrySound:
Just a list of possible changes

- replace convertPathData applyTransforms option with applyTransforms plugin
- throw errors instead of putting into result.error
- replace JSAPI class with set of utilities

strarsis:
Would it make sense to add the txml parser PR to the milestone, too?

vanillajonathan:
Maybe [TypeScript](https://www.typescriptlang.org/)?

TrySound:
Maybe. We already started adding jsdoc for ts.

TrySound:
Started covering plugins with tsdoc. Quite handy.
https://github.com/svg/svgo/pull/1527

septs:
Will this project eventually migrate to TypeScript only? or (most files)

TrySound:
Probably no, I'd like to avoid build step. And tsdoc looks like good enough.

vanillajonathan:
With tsdoc/esdoc/jsdoc you might want to generate HTML documentation which you can host on GitHub pages.

oscarotero:
Typescript or, at least, ES modules would be great. I created a fork to provide support for Deno some time ago (https://github.com/lumeland/svgo) but it's hard to maintain and update every new version. If the SVGO core used ESM it would be much more easy to adapt for Deno.

TrySound:
Thank you for this! Not sure about timing. Will try to start working on v3 before the end of the year.

TrySound:
Btw @oscarotero svgo provides esm module for browsers
https://unpkg.com/browse/svgo@2.7.0/dist/svgo.browser.js

The entry point is this module https://github.com/svg/svgo/blob/master/lib/svgo.js

Instead of https://github.com/svg/svgo/blob/master/lib/svgo-node.js used for node.

Can this help with deno?

oscarotero:
Yes, It's the same entry point that I've used in the Deno fork https://github.com/lumeland/svgo/blob/deno/mod.js
But the browser dist stores the SVG in `globalThis` (or window), and to use it in Deno it must be imported as module.

TrySound:
SVGs are not stored in globalThis. Just some dependencies rely on globalThis to cache polyfills.

oscarotero:
Sorry, you're right. I didn't know SVGO exports ES modules. Just tested it and seems to work fine. Thanks!

I'd love to have some official Deno support (to take advantage of types for configuration and stuff) but, for now, the browser dist just works.

[...]

PR #1905, opened 2023-12-26
chore: Convert project to us ECMAScript modules

jdufresne:
ECMAScript modules are the official standard format to package JavaScript code for reuse. Modules are defined using a variety of import and export statements. Node fully supports ECMAScript modules as they are currently specified and provides interoperability between them and its original module format, CommonJS.

More and more Node packages are switching from CommonJS to ESM, including SVGO dependencies such as csso.

Node documentation on ESM is available at:
https://nodejs.org/api/esm.html

### DIFF
--- a/lib/style.js
+++ b/lib/style.js
@@ -1,29 +1,25 @@
-'use strict';
-
-/**
- * @typedef {import('css-tree').Rule} CsstreeRule
- * @typedef {import('./types').Specificity} Specificity
- * @typedef {import('./types').Stylesheet} Stylesheet
- * @typedef {import('./types').StylesheetRule} StylesheetRule
- * @typedef {import('./types').StylesheetDeclaration} StylesheetDeclaration
- * @typedef {import('./types').ComputedStyles} ComputedStyles
- * @typedef {import('./types').XastRoot} XastRoot
- * @typedef {import('./types').XastElement} XastElement
- * @typedef {import('./types').XastParent} XastParent
- * @typedef {import('./types').XastChild} XastChild
- */
-
-const csstree = require('css-tree');
-const csswhat = require('css-what');
-const {
-  syntax: { specificity },
-} = require('csso');
-const { visit, matches } = require('./xast.js');
-const {
+import * as csstree from 'css-tree';
+import * as csswhat from 'css-what';
+import { syntax } from 'csso';
+import { visit, matches } from './xast.js';
+import {
   attrsGroups,
   inheritableAttrs,
   presentationNonInheritableGroupAttrs,
-} = require('../plugins/_collections.js');
+} from '../plugins/_collections.js';
+
+/**
+ * @typedef {import('css-tree').Rule} CsstreeRule
+ * @typedef {import('./types.js').Specificity} Specificity
+ * @typedef {import('./types.js').Stylesheet} Stylesheet
+ * @typedef {import('./types.js').StylesheetRule} StylesheetRule
+ * @typedef {import('./types.js').StylesheetDeclaration} StylesheetDeclaration
+ * @typedef {import('./types.js').ComputedStyles} ComputedStyles
+ * @typedef {import('./types.js').XastRoot} XastRoot
+ * @typedef {import('./types.js').XastElement} XastElement
+ * @typedef {import('./types.js').XastParent} XastParent
+ * @typedef {import('./types.js').XastChild} XastChild
+ */
 
 // @ts-ignore not defined in @types/csstree
 const csstreeWalkSkip = csstree.walk.skip;
@@ -60,7 +56,7 @@ const parseRule = (ruleNode, dynamic) => {
         }
       });
       rules.push({
-        specificity: specificity(node),
+        specificity: syntax.specificity(node),
         dynamic: hasPseudoClasses || dynamic,
         // compute specificity from original node to consider pseudo classes
         selector: csstree.generate(newNode),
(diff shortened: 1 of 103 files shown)


SethFalco:
Related: https://github.com/svg/svgo/issues/1429

Thanks for dedicating the time to perform the migration. This is something we knew had to be done eventually and was even one of the goals for SVGO v3, but was never prioritized. I'd also been putting it off.

I'll actually put the performance/test work on-hold to review these incoming pull requests. We're starting to get a lot of pull requests piling up now, plus it'll be great to work with modules instead.

I would like to do one more release with CommonJS with the improvements we've been making recently, and some optimizations for existing plugins [Kendell R](https://github.com/KTibow) has been working on.

We'll get this reviewed and merged after that, and prep for SVGO v4. 🎉

Meanwhile, I'd be happy to assist you with rebasing as we merge other pull requests. 👍🏽

jdufresne:
Sounds great. Thanks!

lib/style.js
```diff
@@ -61,7 +58,8 @@ const parseRule = (ruleNode, dynamic) => {
         }
       });
       rules.push({
-        specificity: specificity(node),
+        // @ts-ignore internal api
+        specificity: syntax.specificity(node),
```
jdufresne, line 62: The ts-ignore will become unnecessary if #1910 is merged. Same in `plugins/inlineStyles.js`.

SethFalco:
I've made the following changes:

* Rebased with `main`.
* Removed `use strict` from all files, as that is implicit in modules.
* Updated our Rollup config to also package a CJS bundle to preserve CommonJS compatibility.

> The entire contents of [JavaScript modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) are automatically in strict mode, with no statement needed to initiate it.
> 
> — https://developer.mozilla.org/docs/Web/JavaScript/Reference/Strict_mode#strict_mode_for_modules

As I've seen many package maintainers drop CJS support when they move to ESM, I was under the impression that maintaining both would be a challenge. After looking into it, it's actually not a problem at all. It just comes with the downside of increasing the package size. 🤔

For now, I'd like to support both CJS and ESM. With this in mind, it's probably better to save SVGO v4 for another time, as this isn't the breaking change I was expecting it to be. I'd like to continue resolving bugs in the current version.

Later this week I'll skim over the changes again, verify that it's working properly in both ESM, CJS, CLI, and browser, and potential add tests specific to the CJS module similar to what we have with the browser bundle. Then I'll be glad to merge this!

Please let me know if you think there are any issues with my changes.

package.json
```diff
@@ -1,9 +1,10 @@
 {
   "packageManager": "yarn@2.4.3",
   "name": "svgo",
-  "version": "3.2.0",
+  "version": "3.3.0",
   "description": "Nodejs-based tool for optimizing SVG vector graphics files",
   "license": "MIT",
+  "type": "module",
```
XhmikosR, line 7: Just to verify, will this work with `require` from CJS packages? Like `const svgo = require('svgo');`.

I was under the impression that this change breaks CJS packages and only works with modules, but it's been some time since I last read the docs :)

SethFalco: That change alone would break CJS packages, but I've updated our Rollup config to also create a CJS bundle, which we expose in our `package.json`.

See:

* https://github.com/svg/svgo/blob/main/rollup.config.js#L17
* https://github.com/svg/svgo/blob/main/package.json#L56
* https://github.com/svg/svgo/blob/main/test/svgo.cjs#L1

From my testing with `npm link`, this should work just fine with in CJS and `require`!

XhmikosR: Ah, nice, thanks for the explanation!

rollup.config.js
```diff
@@ -1,25 +1,44 @@
...
+    input: './lib/svgo.js',
+    output: {
+      file: './dist/svgo-node.cjs',
+      format: 'cjs',
+    },
+    external: Object.keys(PKG.dependencies),
+    onwarn(warning) {
+      throw Error(warning.toString());
+    },
+    plugins: [terser()],
```
XhmikosR, line 23: Shouldn't this use the same options as the esm config?

SethFalco: Thanks for calling that out, you're right, though I do have some thoughts here.

* We should keep `mangle: false`. ✅
* I strongly disagree with how we have `comment: false` presently, it should be `'some'` (default) to preserve comments regarding license, copyright, and attribution.
* Reviewing the docs, compress can hinder readability, but rather than turn it off entirely, we could just disable the rules that may impact have a negative impact and keep the smaller changes.

Later this week, I'll review the options for the terser plugin and see what we should do. I may revise `compress` and `format` to fulfil the motivation of the current settings (minify while retaining debug potential).

Related: https://github.com/svg/svgo/pull/1417

By any chance, do you personally utilize or benefit from the reduced minification of these bundles?

So far, thinking a config like this:

```js
/**
 * @typedef {import('@rollup/plugin-terser').Options} Options
 */

/** @type {Options} */
const terserOptions = {
  compress: {
    defaults: false,
    arrows: true,
    computed_props: true,
    conditionals: true,
    dead_code: true,
    drop_debugger: true,
    evaluate: true,
  },
  mangle: false,
  format: { 
    comments: 'some',
    keep_numbers: true,
    semicolons: false,
    shebang: false
  },
};
```

| File | `main` | Current Browser Options | Proposed Options |
|---|---|---|---|
| `svgo.browser.js` | 754.2 kB | 754.2 kB | 749.8 kB | 
| `svgo-node.cjs` | 123.7 kB | 184.5 kB | 182.8 kB |

XhmikosR: My point was that one of the two bundles didn't use the same options.

The options is another discussion :)

Thanks for looking into it!

PR #1909, opened 2023-12-27
chore: Use idiomatic Jest patterns to assert exceptions

jdufresne:
Avoid unnecessary assertions such as `expect(true).toBe(false)`. Instead, assert the exception using the `toThrow()` matcher. Where that is insufficient, use `expect.assertions()`. Unwrap exceptions in promises using `expect.rejects`.

These patterns are documented at:

- https://jestjs.io/docs/expect#tothrowerror
- https://jestjs.io/docs/tutorial-async#rejects
- https://jestjs.io/docs/expect#expectassertionsnumber

### DIFF
--- a/lib/svgo.test.js
+++ b/lib/svgo.test.js
@@ -267,13 +267,13 @@ test('plugin precision should override preset precision', () => {
 });
 
 test('provides informative error in result', () => {
+  expect.assertions(6);
   const svg = `<svg viewBox="0 0 120 120">
       <circle fill="#ff0000" cx=60.444444" cy="60" r="50"/>
     </svg>
   `;
   try {
     optimize(svg, { path: 'test.svg' });
-    expect(true).toEqual(false);
   } catch (error) {
     expect(error.name).toEqual('SvgoParserError');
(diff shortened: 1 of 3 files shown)


lib/svgo.test.js
```diff
@@ -285,13 +285,13 @@ test('provides informative error in result', () => {
...
   const svg = `<svg viewBox="0 0 120 120">
   <circle fill="#ff0000" cx=60.444444" cy="60" r="50"/>
 </svg>
 `;
   try {
     optimize(svg, { path: 'test.svg' });
-    expect(true).toEqual(false);
   } catch (error) {
     expect(error.toString())
       .toEqual(`SvgoParserError: test.svg:2:29: Unquoted attribute value
```
KTibow, line 297: why not convert this one to using `toThrow` too?

jdufresne: The `.toThrow()`  matcher matches against the `.message` attribute only, not to `toString()`. Here `toString()` is being is used to assert the source and line numbers in addition to the message.

PR #1910, opened 2023-12-27
chore: Update @types/csso and correct types

jdufresne:
Follow upstream commit:
https://github.com/DefinitelyTyped/DefinitelyTyped/commit/394c175b3bee0890febfb41b2321b1e11e8b1bdd

Can remove the ts-ignore when importing specificity.

The specificity function return a 3-tuple, so update the type alias in lib/types.d.ts.

### DIFF
--- a/lib/style.js
+++ b/lib/style.js
@@ -16,7 +16,6 @@
 const csstree = require('css-tree');
 const csswhat = require('css-what');
 const {
-  // @ts-ignore internal api
   syntax: { specificity },
 } = require('csso');
 const { visit, matches } = require('./xast.js');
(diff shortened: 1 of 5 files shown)


SethFalco:
Thanks for the fix! I was looking into making a similar pull request a few weeks ago, but shelved it as DefinitelyTyped doesn't accept types for internal APIs, which the `specificity` function isn't documented. 🤔

> We do not support exposing undocumented internal implementation details of libraries in `.d.ts` files.
> 
> — https://definitelytyped.github.io/guides/contributing.html

Might be that I was wrong and it's meant to be public, but in any case, happy to run with this now that it's defined in the types. 🎉

PR #1913, opened 2023-12-28
refactor(convertPaths): clean up plugin

SethFalco:
While reviewing https://github.com/svg/svgo/pull/1889 I'd made some changes locally which I just wanted to push.

This has no (or negligible) impact on performance, it's just to keep things cleaner.

* Make JSDocs with a single line, a single line.
* Declare variables individually in some places.
* Define type `any` on `prev`. We were literally disabling type checking with magic comments on _every usage_ anyway.
* Use `string.includes` instead of `~string.indexOf` as there is no significant difference while the former is more semantic.
* Stop using `apply` methods, we can just use the spread operator.
* Use `Math.sqrt` instead of `Math.hypot` as `Math.hypot` has overhead from improved precision, which we don't need here.
* Replace the regex decimal number check with a math implementation. It's just more logical, plus the original implementation would've failed on large integers or precision numbers anyway, i.e. `5e-324` (float) and `1.7976931348623157e+308` (integer)
* Rename some variables introduced in #1889 for clarity.

### DIFF
--- a/plugins/convertPathData.js
+++ b/plugins/convertPathData.js
@@ -15,25 +15,15 @@ exports.name = 'convertPathData';
 exports.description =
   'optimizes path data: writes in shorter form, applies transformations';
 
-/**
- * @type {(data: number[]) => number[]}
- */
+/** @type {(data: number[]) => number[]} */
 let roundData;
-/**
- * @type {number | false}
- */
+/** @type {number | false} */
 let precision;
-/**
- * @type {number}
- */
+/** @type {number} */
 let error;
-/**
- * @type {number}
- */
+/** @type {number} */
 let arcThreshold;
-/**
- * @type {number}
- */
+/** @type {number} */
 let arcTolerance;
 
 /**
@@ -393,16 +383,17 @@ function filters(
   params,
   { isSafeToUseZ, maybeHasStrokeAndLinecap, hasMarkerMid },
 ) {
-  var stringify = data2Path.bind(null, params),
-    relSubpoint = [0, 0],
-    pathBase = [0, 0],
-    prev = {};
+  const stringify = data2Path.bind(null, params);
+  const relSubpoint = [0, 0];
+  const pathBase = [0, 0];
+  /** @type {any} */
+  let prev = {};
   /** @type {Point | undefined} */
-  let qControlPoint;
+  let prevQControlPoint;
 
   path = path.filter(function (item, index, path) {
-    const qPoint = qControlPoint;
-    qControlPoint = undefined;
+    const qControlPoint = prevQControlPoint;
+    prevQControlPoint = undefined;
 
     let command = item.command;
     let data = item.args;
@@ -415,9 +406,8 @@ function filters(
       if (command === 's') {
         sdata = [0, 0].concat(data);
 
-        // @ts-ignore
-        var pdata = prev.args,
-          n = pdata.length;
+        const pdata = prev.args;
+        const n = pdata.length;
 
         // (-x, -y) of the prev tangent point relative to the current point
         sdata[0] = pdata[n - 2] - pdata[n - 4];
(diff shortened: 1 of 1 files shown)


KTibow:
> Use Math.sqrt instead of Math.hypot as Math.hypot has overhead from improved precision, which we don't need here.

Did the testing really show that? I would think a function that does it all in native code must be faster than going back and forth between JS and native code.

SethFalco:
> Did the testing really show that? I would think a function that does it all in native code must be faster than going back and forth between JS and native code.

Yes and no! There is a measurable difference in synthetic benchmarks, but in practice the difference isn't really significant and may be margin of error (< 1 second).

Here are some benchmarks I ran:

```js
const benchmark = require('benchmark');
const suite = new benchmark.Suite;

const data = [
  [3, 4],
  [10, 10],
  [134, 5235],
  [352523, 52352],
  [525.52352, 532625.5235235],
  [4245, 1352.2352],
];

suite
  .add('Math.sqrt', () => data.map(([dx, dy]) => Math.sqrt(dx ** 2 + dy ** 2)))
  .add('Math.hypot', () => data.map(([dx, dy]) => Math.hypot(dx, dy)))
  .add('Math.sq', () => data.map(([dx, dy]) => (dx ** 2 + dy ** 2) ** 0.5))
  .on('cycle', (event) => console.log(String(event.target)))
  .run({ 'async': true });

// Math.sqrt x 5,724,619 ops/sec ±1.87% (89 runs sampled)
// Math.hypot x 3,313,527 ops/sec ±0.74% (95 runs sampled)
// ** x 5,466,879 ops/sec ±0.96% (95 runs sampled)
```

When looking into why `Math.hypot` was running slower, I found the following:

> `Math.hypot` also avoids overflow/underflow problems if the magnitude of your numbers is very large. The largest number you can represent in JS is [`Number.MAX_VALUE`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_VALUE), which is around 10^308.

In other words, `Math.hypot` does a lot more than what `Math.pow` and `Math.sqrt` does, but we don't work at that scale or with that much precision here.

XhmikosR:
I'd still use `Math.hypot` unless it the performance impact is really affecting svgo. It's more explicit and descriptive like what you did with `string.includes` Otherwise we would still use bitwise operators too, but descriptive and cleaner code is preferable IMHO unless the performance impact shows.

BTW, I recall I had attempted to apply some fixes (similar to the ones you make) across the codebase a couple of years ago which I found with xo, but we never landed them.

SethFalco:
> Otherwise we would still use bitwise operators too

This is different, actually. `Math.sqrt` and `Math.hypot` have a real difference both in benchmarks and in theory, though I'll admit in SVGO it's negligible. The bitwise operators aren't actually faster because it's being used in a loosely resolved conditional.

I assumed the bitwise operator was used for performance, so I did actually check before replacing that with `String.includes`. While `~'cs'.indexOf(…)` is faster than `'cs'.includes(…)`, the loose conditional is slower and has a larger impact.

Though, an even faster approach would probably be to create a `Set` upfront and check if the value is in the set. I opted to just try that now:

```js
const benchmark = require('benchmark');
const suite = new benchmark.Suite;

const data = [
  'c',
  's',
  'a',
  'C',
  'S',
  'A'
];

const str = 'cs';
const set = new Set([
  'c',
  's'
]);

suite
  .add('Bitwise indexOf', () => data.map((d) => ~str.indexOf(d) ? 1 : -1))
  .add('String includes', () => data.map((d) => str.includes(d) ? 1 : -1))
  .add('Set has', () => data.map((d) => set.has(d) ? 1 : -1))
  .on('cycle', (event) => console.log(String(event.target)))
  .run({ 'async': true });

// Bitwise indexOf x 18,704,501 ops/sec ±1.54% (93 runs sampled)
// String includes x 19,032,155 ops/sec ±0.87% (94 runs sampled)
// Set has x 24,285,056 ops/sec ±2.36% (91 runs sampled)
```

> but descriptive and cleaner code is preferable IMHO unless the performance impact shows

I agree with this in general. The reason I opted to change it was just because I felt both were clean enough tbh.

However, given the difference is negligible, when I'm done with what I'm working on, I'll just double-check if it made a practical difference and if not, I'll change it back. 👍🏽

XhmikosR:
Don't get me wrong, I'm all for performance myself. I just don't think it's worth reducing readability/maintainability unless we hit a critical path and performance is noticeable. but I think we are all on the same boat here :)

PR #1917, opened 2023-12-31
Allow .txt extension on plugin test files

johnkenny54:
test/plugins/_index.test.js currently requires plugin test files to have a .svg extension, but the files are not valid SVG files. This has a couple of bad side effects:
- it confuses editors that are expecting valid SVG
- trying to view the test files on GitHub shows an "invalid image format" message rather than the file contents

This change allows the test files to have an optional .txt extension (e.g., filename.01.svg.txt).


jdufresne:
I think something like this is a good idea and would improve my editor experience.

Any reason not to apply to all existing test files? I think we should for the same reason, so they can be seen in the diff and opened in an editor.

johnkenny54:
I agree it should be applied to all files, but since there are 300+, I figured it might be a good idea to start by making it optional in case there are unintended side effects (or if there's a better extension that .txt).

jdufresne:
Just my opinion but I don't think it should be optional. If there are side effects to the change it would be good to understand it now when introducing the feature.

SethFalco:
I appreciate the problem, but am partial to the proposed solution. Happy to hear more feedback, though. The proposal is probably still better than what we're doing now, though.

> I think something like this is a good idea and would improve my editor experience.

This would degrade my default editor experience, though I can easily fix it with a setting override. The test files will lose syntax highlighting, which was working correctly before.

The solution to this would be to add the following in my VSC settings:

```json
{
  "files.associations": {
    "*.svg.txt": "xml"
  }
}
```

Other potential solutions might be:

* Split the test into multiple valid files, but that will result in so many files it'd be a mess.
* Define test cases in an XML file with the nodes `summary`, `input`, `output`, and `params`. However, this would make tests for whitespace kind of annoying. Could be worth it, though.
* The same as above, but in an SVG file with two nested SVGs. Has the same con, but does offer the benefit of being easy to open and render in a browser, which I do a lot to verify the test actually renders the same content.

Does anyone have thoughts on any of these? Otherwise, happy to run with this, but I agree we should rename all existing tests to `.svg.txt` in that case. 🤔 

> in case there are unintended side effects

> If there are side effects to the change

I haven't tested this, but I think the only side effect will be that the `prefixIds` tests will need to have the instances of the file name in the IDs updated as well. i.e. `prefixIds_03_svg` → `prefixIds_03_svg_txt`

KTibow:
> The solution to this would be to add the following in my VSC settings:

Put this in a `.vscode`

SethFalco:
> Put this in a `.vscode`

Nah, I'd rather not. I also dislike when other projects include a `.vscode`, unless it's a project that literally depends on Visual Studio Code like an VSC extension.

1. It clutters the repo imo when it includes editor-specific settings. Fine with ESLint, prettier, EditorConfig, etc. But if we do VSC, then next will be JetBrains, Eclipse, Theia, Sublime, etc.
2. From my experience, most projects that include a `.vscode` bother me more than help. For example, in this case, contributors will find dozens of problems being reported because of invalid XML in the top and bottom, which is partly what we're meant to be solving with this PR.
3. VSC makes it annoying to ignore or override workspace settings imported from a repo, especially when running git commands later, since deleting the file makes the repo dirty.
4. Contributor/maintainers should be responsible for configuring their editors to suit them, not the project, imo. I'm just one maintainer, I don't think it makes sense for me to include files in the repo for everyone for something like this.

In other words, linting code structure, linting, etc belongs in the repo, but editor **_preferences_** do not, imo.

For clarity, not a blocker. I was just noting how this affects a VSC user and what the solution would be if they wanted to preserve the XML syntax highlighting.

KTibow:
Huh. I haven't ran into issues with problems being reported, probably because I use a separate extension for SVGs. But it sounds like you wouldn't want to add suggested extensions here if you don't like editor-specific config.

johnkenny54:
I agree about not adding .vscode to the repository, but how about adding .vscode to .gitignore? That would make things easier for me - code formatting in SVGO is different than what I use in other projects, but if I add a workspace config for SVGO it creates a .vscode directory,

SethFalco:
> But it sounds like you wouldn't want to add suggested extensions here if you don't like editor-specific config.

Are you referring to [Workspace recommended extensions](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions)?

If so, then yes, actually. If contributors overwhelming express interest in it, then I'm happy to consider it. But in general, I don't like seeing projects include configs for specific tooling unless the project depends on that tool.

I can think of some exceptions. For example, when GNOME recommends building their projects with GNOME Builder, I think that's great for beginners to C++/Rust that don't really understand the build system. Or if a certain tool includes documentation for quirks with working in a specific editor, then I could also get behind that. But including settings so that the out-of-the-box experience is subjectively better with specific editors… eh, it feels like the start of one of two things:

* Coupling the project to that editor.
* People wanting us to include preferences/recommendations for other editors.

> but how about adding .vscode to .gitignore? 

I have no qualms with this! I usually include that in other projects anyway. (Mostly to stop others from committing their `.vscode` directories. ^-^')

johnkenny54:
> code formatting in SVGO is different than what I use in other projects

Main difference is that I don't use Prettier. If .vscode is in .gitignore, that's not a problem - I just enable it for my svgo folder.

johnkenny54:
Kind of a long thread, but it seems like the consensus is that the following would be an improvement:
- make .svg.txt extension mandatory
- rename all existing .svg files
- make any necessary changes to test code to handle any side effects.
- add .vscode to .gitignore

Is this accurate? If everyone is OK with this, I can go ahead and try to make these changes.

SethFalco:
> Is this accurate? If everyone is OK with this, I can go ahead and try to make these changes.

Can confirm that's fine with me. 👍🏽

johnkenny54:
All checks pass, but the diffs seem to include the ESM changes as well, so not sure if this will cause problems.

My previous uses of git have been mostly solo projects with simple workflow, so clearly I'm having some conceptual issues with rebasing and the current workflow.

SethFalco:
It looks like you did both a `merge` then `rebase` is all. You only need to do one or the other! As we don't merge to `main` in this branch, it seems `rebase` works out better.

All I did was this to remove the merge commit:

```sh
git reset --hard HEAD~1
git push johnkenny54 svgtest --force-with-lease
```

Meanwhile, you'd performed the rebase correctly. It was just that you already had a merge commit included in this branch before doing it.

test/plugins/_index.test.js
```diff
@@ -5,7 +5,7 @@ import { fileURLToPath } from 'url';
 import { optimize } from '../../lib/svgo.js';
 
 const regEOL = new RegExp(EOL, 'g');
-const regFilename = /^(.*)\.(\d+)\.svg$/;
+const regFilename = /^(.*)\.(\d+)\.svg\.txt?$/;
```
jdufresne, line 8:
```suggestion
const regFilename = /^(.*)\.(\d+)\.svg\.txt$/;
```

PR #1923, opened 2024-01-03
Fixed bug in removeUselessDefs

### LINKED ISSUE #1922: removeUselessDefs removes <symbol> which contains children with IDs
If the  `<symbol>` itself does not have an ID, removeUselessDefs removes `<symbol>` elements which contains children with IDs.

**To Reproduce**
Optimize the following with removeUselessDefs enabled:

```
    <rect fill="url(#a)" width="64" height="64"/>
    <symbol>
        <linearGradient id="a">
            <stop offset="5%" stop-color="gold" />
        </linearGradient>
    </symbol>
</svg>
``` 

**Expected behavior**
File is unchanged because linearGradient is referenced.

**Additional context**
This bug is part of the cause of test-regression failures on media-flash.svg.
- SVGO Version 3.2

johnkenny54:
This change applies the same logic used to process `<def>` elements to all non-rendering elements (i.e. preserve referenced children, and only delete the parent if no children remain).

This fix is part of the solution to removing the media-flash.svg mismatches in test-regression. This fix alone does not change the number of regression mismatches.

Closes #1922.

### DIFF
--- a/test/plugins/removeUselessDefs.04.svg
+++ b/test/plugins/removeUselessDefs.04.svg
@@ -1,0 +1,23 @@
+Don't remove used symbols.
+
+===
+
+<svg xmlns="http://www.w3.org/2000/svg">
+    <rect fill="url(#a)" width="64" height="64"/>
+    <symbol>
+        <linearGradient id="a">
+            <stop offset="5%" stop-color="gold" />
+        </linearGradient>
+    </symbol>
+</svg>
+
+@@@
+
+<svg xmlns="http://www.w3.org/2000/svg">
+    <rect fill="url(#a)" width="64" height="64"/>
+    <symbol>
+        <linearGradient id="a">
+            <stop offset="5%" stop-color="gold"/>
+        </linearGradient>
+    </symbol>
+</svg>
--- a/test/plugins/removeUselessDefs.04.svg
+++ b/test/plugins/removeUselessDefs.04.svg
@@ -1,4 +1,4 @@
-Don't remove used symbols.
+Don't remove non-rendering elements if children have IDs.
 
 ===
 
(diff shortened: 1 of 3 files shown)


test/plugins/removeUselessDefs.05.svg
```diff
@@ -0,0 +1,23 @@
+Don't remove used symbols.
```
SethFalco, line 1: Could you please explain this test?

It doesn't look like it would've been broken before, and it doesn't feature a `defs` node or other hidden element.

Just want to check was this added due to a known issue, or were you just playing it safe and checking more cases?

test/plugins/removeUselessDefs.04.svg
```diff
@@ -0,0 +1,23 @@
+Don't remove used symbols.
```
SethFalco, line 1:
```suggestion
Don't remove non-rendering elements if children have IDs.
```

Could we make the description this instead? It's just a bit clearer imo.

SethFalco (commented):
This looks great! Thanks for looking through and spotting the issue.

This also reveals a bug in Firefox, so I'll file a report that too. The new test case doesn't render in Firefox, but does render in Chromium and Eye of GNOME.

test/plugins/removeUselessDefs.05.svg
johnkenny54: Basically playing it safe. Since `<g>` is neither in the non-rendering set nor a `<def>`, just wanted to make sure nothing weird happened in that case.

SethFalco:
Bug report filed here:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1872907

PR #1925, opened 2024-01-03
Fixed bug in removeHiddenElems

### LINKED ISSUE #1924: removeHiddenElems removes <symbol> which contains children with IDs
If the `<symbol>` itself does not have an ID, removeHiddenElems removes `<symbol>`elements which contains children with IDs.

**To Reproduce**
Optimize the following with removeHiddenElems enabled:

```
    <rect fill="url(#a)" width="64" height="64"/>
    <symbol>
        <linearGradient id="a">
            <stop offset="5%" stop-color="gold" />
        </linearGradient>
    </symbol>
</svg>

``` 

**Expected behavior**
File is unchanged because linearGradient is referenced.

**Additional context**
This bug is part of the cause of test-regression failures on media-flash.svg.
- SVGO Version 3.2

johnkenny54:
This change removes non-rendering nodes only if they have no referenced children.

This fix is part of the solution to removing the media-flash.svg mismatches in test-regression. On its own, this fix resolves 8 mismatches in test-regression. In conjunction with PR #1923 this fix resolves 14 mismatches in test-regression.

**Further Discussion**
This plugin has some code that seems like it needs to be removed, but maybe I'm missing something. In addition to the usual return of the visitor functions, this plugin has an explicit call to visit(), with different code than what is in the return value. That causes the tree to be traversed twice, and the `<symbol>` element was being removed in both sets of code. I removed the code in the explicit call to visit(), but there is still code in the visit() call that removes elements with opacity=0. This seems weird because:
- traversing the tree twice seems inefficient
- removing the elements based on what is in a `<style>` element seems contradictory to what happens in the visitor function that is returned by the plugin (where optimization is disabled in the presence of a `<style>` element)

So the code I submitted fixes the referenced issue, but it seems like this plugin should also be changed by:
- removing the explicit call to visit()
- changing the expected behavior of removeHiddenElems.02.svg so that the source file is not changed

Is there a good reason for the code to have the explicit call to visit()?

Closes #1924

### DIFF
--- a/plugins/removeHiddenElems.js
+++ b/plugins/removeHiddenElems.js
@@ -94,6 +94,23 @@ exports.fn = (root, params) => {
   let deoptimized = false;
 
   /**
+   * Nodes can't be removed if they or any of their children have an id attribute that is referenced.
+   * @param {XastElement} node
+   * @returns boolean
+   */
+  function canRemoveNonRenderingNode(node) {
+    if (allReferences.has(node.attributes.id)) {
+      return false;
+    }
+    for (const child of node.children) {
+      if (child.type === 'element' && !canRemoveNonRenderingNode(child)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
    * @param {XastChild} node
    * @param {XastParent} parentNode
    */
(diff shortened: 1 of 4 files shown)


johnkenny54:
It looks like this change introduced 2 mismatches in the W3C test regressions, which I hadn't noticed locally since the overall number of regressions (including the oxygen SVGs) was reduced. I will look into this.

johnkenny54:
Regression tests fixed. Reverted some of the changes, including some of the logic in the explicit call to visit() which were necessary to accumulate list of nonRenderedNodes. Added a test case to catch the problem that was causing regression failure.

With these changes, this fix on its own would not have reduced the number of regression mismatches, but now that PR #1923 has been merged, this fix reduces mismatches by 6.

It looks to me like all the logic in the explicit call to visit could be moved to returned function - this would eliminate the extra tree traversal, and also make the plugin easier to follow. But unless it's causing problems as is, that should probably be a separate PR.

SethFalco (approved):
LGTM! Thanks for identifying the issue and submitting a patch.

SethFalco:
I rebased locally to change `.svg` to `.svg.txt` and didn't push before merging… 🤦🏽
I'll address that in a separate PR.

PR #1927, opened 2024-01-06
fix(convertPathData): fix q control point behavior when item is removed

### LINKED ISSUE #1926: v3.2.0 'Cannot read properties of undefined' error in reflectPoint
**Describe the bug**

In version 3.2.0 `svgo` fails on certain files with the following error when using the `convertPathData` plugin. The issue does not occur in v3.1.0 and earlier.

```
/tmp/node_modules/svgo/plugins/convertPathData.js:1160
  return [2 * base[0] - controlPoint[0], 2 * base[1] - controlPoint[1]];
                                    ^

TypeError: Cannot read properties of undefined (reading '0')
    at reflectPoint (/tmp/node_modules/svgo/plugins/convertPathData.js:1160:37)
    at /tmp/node_modules/svgo/plugins/convertPathData.js:802:43
    at Array.filter (<anonymous>)
    at filters (/tmp/node_modules/svgo/plugins/convertPathData.js:397:15)
    at Object.enter (/tmp/node_modules/svgo/plugins/convertPathData.js:183:20)
    at visit (/tmp/node_modules/svgo/lib/xast.js:52:30)
    at visit (/tmp/node_modules/svgo/lib/xast.js:68:9)
    at visit (/tmp/node_modules/svgo/lib/xast.js:68:9)
    at visit (/tmp/node_modules/svgo/lib/xast.js:68:9)
    at visit (/tmp/node_modules/svgo/lib/xast.js:68:9)

Node.js v20.6.1
```
An [example file](https://raw.githubusercontent.com/benjaminion/upgrading-ethereum-book/capella/src/images/charts/hysteresis.svg) that triggers the error - the file [validates with no issues](https://validator.w3.org/nu/?doc=https%3A%2F%2Fraw.githubusercontent.com%2Fbenjaminion%2Fupgrading-ethereum-book%2Fcapella%2Fsrc%2Fimages%2Fcharts%2Fhysteresis.svg). 

**To Reproduce**
Run the following with the above file as input:

```
const svgo = require('svgo')
const fs = require('fs')

const originalSvg = fs.readFileSync('hysteresis.svg', 'utf8')
const svg = svgo.optimize(
  originalSvg,
  {
    plugins: ['convertPathData']
  }
)
```

**Expected behavior**
No errors.

- SVGO Version: 3.2.0
- NodeJs Version 20.6.1
- OS: Ubuntu 22.04.3 LTS

KTibow:
fix #1926 by only resetting the control point if an item is included
realistically this should be fine as it'll still get reset on move commands, and likely if a command is dropped the control point stayed the same

there's another issue with how in the case `M0 0Q10 0 10 10Q10 10 10 10T20 20`, convertPathData should convert `T` to longhand and update the control point before dropping the useless command (changing the control point), but that's out of scope for this PR. (if my thinking is correct, i can't think of any cases where converting from q to t is affected by this behavior)

### DIFF
--- a/plugins/convertPathData.js
+++ b/plugins/convertPathData.js
@@ -394,7 +394,6 @@ function filters(
 
   path = path.filter(function (item, index, path) {
     const qControlPoint = prevQControlPoint;
-    prevQControlPoint = undefined;
 
     let command = item.command;
     let data = item.args;
@@ -895,6 +894,8 @@ function filters(
         // @ts-ignore
         prevQControlPoint = item.coords;
       }
+    } else {
+      prevQControlPoint = undefined;
     }
     prev = item;
     return true;
--- a/test/plugins/convertPathData.36.svg.txt
+++ b/test/plugins/convertPathData.36.svg.txt
@@ -1,0 +1,15 @@
+Should process/optimize q correctly even if useless commands in between.
+
+See: https://github.com/svg/svgo/issues/1926
+
+===
+
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1 20">
+  <path d="M-6.3 9.9q.7-4.5.2-5-.5-.5-1.5-.5l0 0q-.4 0-2 .3"/>
+</svg>
+
+@@@
+
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1 20">
+    <path d="M-6.3 9.9q.7-4.5.2-5t-1.5-.5q-.4 0-2 .3"/>
+</svg>


SethFalco (approved):
Thanks for taking a look at this. The fix makes sense to me.

I'd appreciate it if you'd include a test case in future PRs. I've added one for you here, though. 👍🏽

C0rn3j:
Is there a release being planned soon?  
I've broken my teeth on this and was a bit baffled to find it was fixed three months ago.

Thanks to the both of you for fixing it.

SethFalco:
Hey! I was planning to make a release by the end of the month.

Maintenance was stalled last month as I got ill, then had some client work that took priority. I expect to be back on SVGO next week, though. 👍🏽

vltansky:
Looks like this commit reverted in `3.3.2`. Is it intentioned?
Got the same error

SethFalco:
Yeah, if you read the release notes, due to breaking changes we deprecated and reverted the last release.

If you aren't impacted by the breaking changes in v3.3.1, you're welcome to continue using that until v4 comes out. Meanwhile, I'd recommend waiting for v4 (or at least the v4 release candidates) which will include these bug fixes.

C0rn3j:
That's unfortunate, as latest stable will bring back old bugs, would it perhaps be a good idea to keep the bug fixes and reverting just the module changes?

SethFalco:
> as latest stable will bring back old bugs

If you were using v3.3.0 or v3.3.1 without issue, you're welcome to continue using it!

> keep the bug fixes and reverting just the module changes?

I would've done that, but the switch to ESM was done before any of the bug fixes, so I couldn't just cherry-pick that change away. With v4 being planned anyway, it seemed more worthwhile to focus on that rather than backport.

I was hoping to have a v4 release candidate out this week, but I was held up. I'd expect SVGO v4.0.0-rc-1 to be out next week, though. That will include the bug fixes, though you'll also need to peek at the release notes to see breaking changes.

(Mostly impacting imports, and a few changes to default plugins like `removeViewBox`, `removeTitle`, and also renaming `removeScriptElement` to `removeScripts` etc.)

Meanwhile, if it's urgent, I'd recommend either disabling the optimization, or using the patch feature of Yarn/pnpm (or `patch-package` if using npm) to apply an adapter version of this PR to your repository until the next release is ready.

vltansky:
Ah got it, thanks.
CleanShot 2024-06-04 at 09 36 03@2x
Guess this label confused me

SethFalco:
Just a heads-up, the release candidate for SVGO v4 is now available. Feel free to use/try it, and if you encounter any other problems, report an issue!

https://github.com/svg/svgo/releases/tag/v4.0.0-rc.0

jlarmstrongiv:
Is the second case:

> there's another issue with how in the case M0 0Q10 0 10 10Q10 10 10 10T20 20, convertPathData should convert T to longhand and update the control point before dropping the useless command (changing the control point), but that's out of scope for this PR. (if my thinking is correct, i can't think of any cases where converting from q to t is affected by this behavior)

covered in another PR?

PR #1930, opened 2024-01-06
fix: collapseGroups does not move attributes atomically

### LINKED ISSUE #1928: collapseGroups does not move attributes atomically
One thing the collapseGroups plugin does is move all attributes from a `<g>` element to its child if it only has a single child. It appears that the code intends this to be an "all or none" move (i.e., if any attributes can't be moved, then no attributes are moved).

As currently implemented, the code will move attributes partially and unpredictably - it moves attributes to the child sequentially until it encounters an attribute that it can't move, then it stops moving attributes. So it is possible that only some attributes are moved, and which ones are moved depends on the order of attributes, which may be changed by other plugins.

This bug is one of the reasons for regression failure of \scalable\devices\media-flash.svg.

SVGO 3.2

johnkenny54:
In the test case file, before this change, the transform attribute on the group was moved to the child, but the style containing the filter was not moved. This was changing the rendering.

regression.js before fix:

- Mismatched: 214
- Passed: 5308
- Total reduction 764239791 bytes

After fix:

- Mismatched: 201
- Passed: 5321
- Total reduction 764238752 bytes

This removes 13 regression mismatches. Compression of regression files is reduced by 1039 bytes.

Closes #1928

### DIFF
--- a/plugins/collapseGroups.js
+++ b/plugins/collapseGroups.js
@@ -9,33 +9,6 @@ export const name = 'collapseGroups';
 export const description = 'collapses useless groups';
 
 /**
- * @param {XastElement} node
- * @param {XastElement} firstChild
- * @returns boolean
- */
-function canMoveAttributesToSingleChild(node, firstChild) {
-  for (const [name, value] of Object.entries(node.attributes)) {
-    // avoid copying to not conflict with animated attribute
-    if (hasAnimatedAttr(firstChild, name)) {
-      return false;
-    }
-    if (firstChild.attributes[name] == null) {
-      continue;
-    } else if (name === 'transform') {
-      continue;
-    } else if (firstChild.attributes[name] === 'inherit') {
-      continue;
-    } else if (
-      inheritableAttrs.has(name) === false &&
-      firstChild.attributes[name] !== value
-    ) {
-      return false;
-    }
-  }
-  return true;
-}
-
-/**
  * @type {(node: XastNode, name: string) => boolean}
  */
 const hasAnimatedAttr = (node, name) => {
@@ -108,20 +81,30 @@ export const fn = () => {
                 node.attributes.transform == null &&
                 firstChild.attributes.transform == null))
           ) {
-            if (!canMoveAttributesToSingleChild(node, firstChild)) {
-              return;
-            }
+            const newChildElemAttrs = { ...firstChild.attributes };
+
             for (const [name, value] of Object.entries(node.attributes)) {
-              if (firstChild.attributes[name] == null) {
-                firstChild.attributes[name] = value;
-              } else if (name === 'transform') {
-                firstChild.attributes[name] =
-                  value + ' ' + firstChild.attributes[name];
-              } else if (firstChild.attributes[name] === 'inherit') {
-                firstChild.attributes[name] = value;
+              // avoid copying to not conflict with animated attribute
+              if (hasAnimatedAttr(firstChild, name)) {
+                return;
               }
-              delete node.attributes[name];
+
+              if (newChildElemAttrs[name] == null) {
+                newChildElemAttrs[name] = value;
+              } else if (name === 'transform') {
+                newChildElemAttrs[name] = value + ' ' + newChildElemAttrs[name];
+              } else if (newChildElemAttrs[name] === 'inherit') {
+                newChildElemAttrs[name] = value;
+              } else if (
+                !inheritableAttrs.has(name) &&
+                newChildElemAttrs[name] !== value
+              ) {
+                return;
+              }
             }
+
+            node.attributes = {};
+            firstChild.attributes = newChildElemAttrs;
           }
         }
 
--- a/plugins/collapseGroups.js
+++ b/plugins/collapseGroups.js
@@ -1,11 +1,39 @@
 import { inheritableAttrs, elemsGroups } from './_collections.js';
 
 /**
+ * @typedef {import('../lib/types.js').XastElement} XastElement
  * @typedef {import('../lib/types.js').XastNode} XastNode
  */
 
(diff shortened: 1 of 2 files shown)


test/plugins/collapseGroups.17.svg.txt
```diff
@@ -0,0 +1,43 @@
...
+        <g style="filter:url(#b)">
+            <linearGradient id="a" x1="3.97" x2="3.97" y1="2.425" y2="7.55" gradientUnits="userSpaceOnUse">
+                <stop offset="1" style="stop-color:#fff"/>
+            </linearGradient>
+        </g>
+    </symbol>
+    <g transform="matrix(0 -1.2285 -1.2285 0 112.337 268.327)" style="filter:url(#b)">
+        <path d="M.84.348H7.1V5.8H3.344z" style="fill:url(#c)"/>
+    </g>
+</svg>
```
KTibow, line 22: I somewhat-strongly dislike this test case. It doesn't render in your browser, and it has too much cruft. Use something like this as the input instead.
```suggestion
<svg xmlns="http://www.w3.org/2000/svg" width="60" height="60">
  <defs>
    <filter id="a">
      <feGaussianBlur stdDeviation="1"/>
    </filter>
  </defs>
  <g transform="scale(2)" style="filter:url(#a)">
    <rect width="20" height="20" fill="blue" style=""/>
  </g>
</svg>
```

johnkenny54: I'd like to keep it as is. The test file is a distillation of the Oxygen icon that was failing. It took me a really long time to isolate the problem, and I know for sure that the original file I submitted was failing before the change.

KTibow: Then I hope you will consider at least making it render in a browser and ideally distill it more

johnkenny54: Simplified the test case. New test case fails on main branch, passes here.

KTibow: The "incorrect" behavior renders the same as the "correct" behavior, but for the purposes of demonstrating the required behavior this test works fine.

johnkenny54: Yeah, I think it's pretty hard to build an example where the bug really makes a difference. That's one of the reasons it was so hard to track down. And also why the regression tests with "real" SVGs are important.

SethFalco:
```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 88 88">
  <filter id="a">
    <feGaussianBlur stdDeviation="1"/>
  </filter>
  <g transform="matrix(0.6875,0,0,0.6875,20.34375,66.34375)" style="filter:url(#a)">
    <path d="M 33.346591,-83.471591 L -10.744318,-36.471591 L -10.49989,-32.5 L 34.409091,-77.375 L 78.951759,-33.318072 L 80.5625,-36.471591 L 35.471591,-83.471591 L 33.346591,-83.471591 z" style="fill-opacity:1"/>
  </g>
</svg>
```

Here is an example that reproduces the problem and clearly demonstrates the change in rendering caused by the plugin.

It looks like what you describe to be the issue and solution is correct and applies to this example. Could you use this for the test case though, please!

johnkenny54: I changed the test file.

I did pretty much what you did to isolate this, but this one was difficult to isolate because the file didn't fail if you only ran collapseGroups - you also had to run some other combination of plugins that would reorder the attributes in a way that would cause collapseGroups to trigger the error, and deleting parts of the original file sometimes changed things in such a way that the error didn't show up, so there were a lot of red herrings.

Also, see PR #1932 - this is a start on changing regression.js to make testing/analysis easier. I've been using this version to analyze problems and test changes. Still more to be done obviously, but it's a start.

plugins/collapseGroups.js
```diff
@@ -80,23 +108,17 @@ export const fn = () => {
                 node.attributes.transform == null &&
                 firstChild.attributes.transform == null))
           ) {
+            if (!canMoveAttributesToSingleChild(node, firstChild)) {
+              return;
+            }
             for (const [name, value] of Object.entries(node.attributes)) {
```
SethFalco, line 114: Instead of potentially iterating twice, I think it'd be better to shallow copy the attributes and assign them if we complete the loop without returning.

This achieves the same thing while being fewer lines of code.

```js
const newChildElemAttrs = { ...firstChild.attributes };

for (const [name, value] of Object.entries(node.attributes)) {
  // avoid copying to not conflict with animated attribute
  if (hasAnimatedAttr(firstChild, name)) {
    return;
  }

  if (newChildElemAttrs[name] == null) {
    newChildElemAttrs[name] = value;
  } else if (name === 'transform') {
    newChildElemAttrs[name] = value + ' ' + newChildElemAttrs[name];
  } else if (newChildElemAttrs[name] === 'inherit') {
    newChildElemAttrs[name] = value;
  } else if (!inheritableAttrs.has(name) && newChildElemAttrs[name] !== value) {
    return;
  }
}

node.attributes = {};
firstChild.attributes = newChildElemAttrs;
```

johnkenny54: Made the change and re-ran regression.js locally, same results as above.

The reason my regression numbers are different is that I'm running on Windows, and there seem to be a small number of (consistent) false positives. Every one I've looked at so far has been something in a `<text>` element which looks like a minor antialiasing issue. Hopefully I'll figure out a solution to this at some point.

SethFalco: Thanks for the explanation. It helps (and sucks) to know different operating systems, or possibly different hardware(?), have an impact on our tests.

PR #1932, opened 2024-01-06
Add statistics to regression.js

johnkenny54:
I've been using regression.js a lot locally to test the Oxygen icons and W3C files, both to isolate problems and to verify that my changes aren't having any unintended side effects.

These changes should have no impact on the regressions being run as part of CI.

Changes to regression.js:

- Record statistics during test run (input file length, optimized file length).
- Write a timestamped .tsv file to tmp directory at end of run (for each file, show pass/fail, orig length, optimized length, reduction in size).
- Show total reduction in file size on console at end of tests.
- Added some configuration constants. Eventually I would like to make these command line options. I've been using this file to track down the regression errors with the Oxygen icons and it's very helpful to be able to run the regressions on a subset of files and/or with alternate configurations.

Added tmp to .gitignore

### DIFF
--- a/test/regression.js
+++ b/test/regression.js
@@ -4,6 +4,7 @@
  */
 
 import fs from 'node:fs/promises';
+import fsSync from 'node:fs';
 import http from 'http';
 import os from 'os';
 import path from 'path';
@@ -18,11 +19,22 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const width = 960;
 const height = 720;
 
+const FILE_PATTERN = new RegExp('media-flash');
+// const FILE_PATTERN = new RegExp('.*');
+let configFileName;
+const DEFAULT_CONFIG = {
+  floatPrecision: 4,
+};
+const CONFIG = configFileName ? readConfigFile(configFileName) : DEFAULT_CONFIG;
+
+const stats = {};
+
 /** @type {PageScreenshotOptions} */
 const screenshotOptions = {
   omitBackground: true,
   clip: { x: 0, y: 0, width, height },
   animations: 'disabled',
+  timeout: 80000,
 };
 
 /**
@@ -38,9 +50,15 @@ const runTests = async (list) => {
    * @param {string} name
    */
   const processFile = async (page, name) => {
-    await page.goto(`http://localhost:5000/original/${name}`);
+    const fileStats = {};
+    stats[name.replaceAll('\\', '/')] = fileStats;
+    await page.goto(`http://localhost:5000/original/${name}`, {
+      timeout: 80000,
+    });
     const originalBuffer = await page.screenshot(screenshotOptions);
-    await page.goto(`http://localhost:5000/optimized/${name}`);
+    await page.goto(`http://localhost:5000/optimized/${name}`, {
+      timeout: 80000,
+    });
     const optimizedBufferPromise = page.screenshot(screenshotOptions);
 
     const writeDiffs = process.env.NO_DIFF == null;
@@ -58,9 +76,11 @@ const runTests = async (list) => {
     if (matched <= 4) {
       console.info(`${name} is passed`);
       passed++;
+      fileStats.result = 'pass';
     } else {
       mismatched++;
       console.error(`${name} is mismatched`);
+      fileStats.result = 'mismatch';
       if (diff) {
         const file = path.join(
           __dirname,
@@ -76,6 +96,9 @@ const runTests = async (list) => {
     let item;
     const page = await context.newPage();
     while ((item = list.pop())) {
+      if (!FILE_PATTERN.test(item)) {
+        continue;
+      }
       await processFile(page, item);
     }
     await page.close();
@@ -95,6 +118,12 @@ const runTests = async (list) => {
   return mismatched === 0;
 };
 
+function readConfigFile(fileName) {
+  const data = fsSync.readFileSync(fileName);
+  const json = JSON.parse(data);
+  return json;
+}
+
 (async () => {
   try {
     const start = process.hrtime.bigint();
@@ -102,6 +131,7 @@ const runTests = async (list) => {
     const filesPromise = fs.readdir(fixturesDir, { recursive: true });
     const server = http.createServer(async (req, res) => {
       const name = req.url.slice(req.url.indexOf('/', 1));
+      const statsName = name.substring(1);
       let file;
       try {
         file = await fs.readFile(path.join(fixturesDir, name), 'utf-8');
@@ -112,14 +142,15 @@ const runTests = async (list) => {
       }
 
       if (req.url.startsWith('/original/')) {
+        stats[statsName].lengthOrig = file.length;
         res.setHeader('Content-Type', 'image/svg+xml');
         res.end(file);
         return;
       }
       if (req.url.startsWith('/optimized/')) {
-        const optimized = optimize(file, {
-          floatPrecision: 4,
-        });
+        const optimized = optimize(file, CONFIG);
+        stats[statsName].lengthOpt = optimized.data.length;
+
         res.setHeader('Content-Type', 'image/svg+xml');
         res.end(optimized.data);
         return;
@@ -134,6 +165,28 @@ const runTests = async (list) => {
     server.close();
     const end = process.hrtime.bigint();
     const diff = (end - start) / BigInt(1e6);
+
+    // Write statistics.
+    const statArray = [
+      ['Name', 'Result', 'Orig Len', 'Opt Len', 'Reduction'].join('\t'),
+    ];
+    let totalReduction = 0;
+    for (const name of Object.keys(stats).sort()) {
+      const fileStats = stats[name];
+      const orig = fileStats.lengthOrig;
+      const opt = fileStats.lengthOpt;
+      const reduction = orig - opt;
+      totalReduction += reduction;
+      statArray.push([name, fileStats.result, orig, opt, reduction].join('\t'));
+    }
+    const statsFileName = `tmp/regression-stats-${new Date()
+      .toISOString()
+      .replaceAll(':', '')
+      .substring(0, 17)}.tsv`;
+    await fs.mkdir(path.dirname(statsFileName), { recursive: true });
+    await fs.writeFile(statsFileName, statArray.join('\n'));
+
+    console.info(`Total reduction ${totalReduction} bytes`);
     if (passed) {
       console.info(`Regression tests successfully completed in ${diff}ms`);
     } else {
--- a/test/regression.js
+++ b/test/regression.js
@@ -19,8 +19,7 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const width = 960;
 const height = 720;
 
-const FILE_PATTERN = new RegExp('media-flash');
-// const FILE_PATTERN = new RegExp('.*');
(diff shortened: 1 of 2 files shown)


test/regression.js
```diff
@@ -18,11 +19,21 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
...
+const CONFIG = configFileName ? readConfigFile(configFileName) : DEFAULT_CONFIG;
+
+const stats = {};
+
 /** @type {PageScreenshotOptions} */
 const screenshotOptions = {
   omitBackground: true,
   clip: { x: 0, y: 0, width, height },
   animations: 'disabled',
+  timeout: 80000,
```
SethFalco, line 36: Can you explain why you're adding the timeouts?
Given the nature of these tests, I don't think they'd ever be needed to it seems to just be adding noise. 🤔

test/regression.js
```diff
@@ -18,11 +19,21 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const width = 960;
 const height = 720;
 
+const FILE_PATTERN = new RegExp('.*');
```
SethFalco, line 22: It'd be much simpler to just catch an arg and handle this when collecting files.

```diff
    try {
      const start = process.hrtime.bigint();
      const fixturesDir = path.join(__dirname, 'regression-fixtures');
      const filesPromise = fs.readdir(fixturesDir, { recursive: true });
+     const patternArg = process.argv[2];
+     const regex = (patternArg) ? new RegExp(patternArg) : null;
…
-     const list = (await filesPromise).filter((name) => name.endsWith('.svg'));
+     const list = (await filesPromise)
+       .filter((name) => name.endsWith('.svg'))
+       .filter((name) => regex === null || regex.test(name));
```

Usage:

* `node test/regression color-prop`
* `yarn run test-regression color-prop`

test/regression.js
```diff
@@ -18,11 +19,21 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
...
+let configFileName;
+const DEFAULT_CONFIG = {
+  floatPrecision: 4,
+};
+const CONFIG = configFileName ? readConfigFile(configFileName) : DEFAULT_CONFIG;
```
SethFalco, line 27: I'd much prefer if we didn't use config files or create output files ourselves.

For input, we can read from CLI as you allude to in your PR description, or an environment variable.

For output, we should keep it to STDOUT, and the contributor can choose to pipe or redirect it to whatever tool/file they want if they want to process it or present it differently.

test/regression.js
```diff
@@ -112,14 +141,15 @@ const runTests = async (list) => {
...
         res.setHeader('Content-Type', 'image/svg+xml');
         res.end(file);
         return;
       }
       if (req.url.startsWith('/optimized/')) {
-        const optimized = optimize(file, {
-          floatPrecision: 4,
-        });
+        const optimized = optimize(file, CONFIG);
+        stats[statsName].lengthOpt = optimized.data.length;
```
SethFalco, line 151: Is this actually useful? I shouldn't store stats for the sake of storing stats.

I'd imagine it's only helpful to store the overall bytes reduced, or the average percentage reduced across files. There's no need to store the difference between all files, though.

In the case that someone is focused on a specific file, they can pass the argument to limit files tested.

test/regression.js
```diff
@@ -134,6 +164,28 @@ const runTests = async (list) => {
     server.close();
     const end = process.hrtime.bigint();
     const diff = (end - start) / BigInt(1e6);
+
+    // Write statistics.
+    const statArray = [
+      ['Name', 'Result', 'Orig Len', 'Opt Len', 'Reduction'].join('\t'),
```
SethFalco, line 170: I don't really see the benefit of presenting the data in a TSV file. It's just adding a lot of noise here. It would be nice to improve the output of the pipeline, and if you want this, you can pipe that and process it however you want when you run it.

It's not something I'd be keen on us maintaining as part of this script, especially as this is tied to your workflow and not for general contributors/maintainers.

SethFalco (commented):
Thanks for peeking at this. I agree the output of our regression tests could be better, but I disagree with how some of this was approached.

In particular, I'd prefer if we kept regression.js as a simple CLI tool that just reads args and writes to STDOUT, and we can work on improving that experience rather than maintaining multiple output formats.

test/regression.js
johnkenny54: > I'd imagine it's only helpful to store the overall bytes reduced, or the average percentage reduced across files. There's no need to store the difference between all files, though.

Yes there is. If you know that there is an overall difference in 2 runs, the only way to tell which files contribute is to diff the two .tsv files. See real world example below.

> Can you explain why you're adding the timeouts?

Running locally in the background they were occasionally timing out with the default (30000). Kind of a bummer when they fail 20+ minutes into a run.

> It'd be much simpler to just catch an arg and handle this when collecting files.

There are other options that I have been using as well, so making command line arguments positional doesn't seem like a good idea. For now I was just collecting them up in one place to see what options were useful.

> For output, we should keep it to STDOUT, and the contributor can choose to pipe or redirect it to whatever tool/file they want if they want to process it or present it differently.

I don't see how that would work with the .tsv. Basically the .tsv output is creating a log file. Not optional in the version I submitted, but easy enough to make it optional if we want.

> I don't really see the benefit of presenting the data in a TSV file.

They're easy to create, easy to diff, and can be opened in a spreadsheet for further analysis.

Here's an example of how I've been using this:
- I had a one line change that theoretically improved performance using multipass without changing behavior.
- Ran regression.js before/after the change (note that this required being able to change the config in regression.js).
    - total run time reduced by 7%
    - number of mismatches identical
    - **Total reduction changed from 786597195 to 786598159 bytes**
- Diffed the .tsv files to see what was changed. None of the files seemed to have changed pass/fail status, but quite a few had different output sizes.
- Drilled down into one of the files. It was a large file, and the only change was a small change in one of the path segments.
- At that point I said "clearly there's a change in behavior, not sure if it's good, bad, or indifferent, but I need to table this until I have time to understand it better".

So bottom line I guess is that I have found this very useful, but if no one else does, seems like the best solution is for me to maintain my copy locally.

SethFalco: Our difference in opinion here probably comes from how we use the terminal. I'm not against improving the output. I'm only against creating specific files, in a specific location, in a specific format.

I don't want us to maintain multiple output formats here, but rather just to maintain nice output to STDOUT and have contributors utilize the versatility that the terminal offers to do whatever they want with it.

> If you know that there is an overall difference in 2 runs, the only way to tell which files contribute is to diff the two .tsv files.

This is already possible with the current setup. See my usage at the bottom.

> Running locally in the background they were occasionally timing out with the default (30000).

The part that confused me was actually that the default timeout for Playwright is `0` which means no timeout. Now I'm more confused. ^-^'

Sorry, when or if this occurs again, would you mind sharing the error message or just confirming that it was Playwright throwing the error?

> Maximum operation time in milliseconds. Defaults to `0` - no timeout.
> 
> — https://playwright.dev/docs/api/class-page#page-drag-and-drop-option-timeout

> There are other options that I have been using as well, so making command line arguments positional doesn't seem like a good idea.

Valid, perhaps we should just wrap this in `commander` then. We already depend on it anyway, so it doesn't hurt to add it to our test workflow.

> I don't see how that would work with the .tsv. Basically the .tsv output is creating a log file. Not optional in the version I submitted, but easy enough to make it optional if we want.

We can have cleaner output to STDOUT in a structured format (TSV or otherwise) then you can do add `> path/to/file` or `2> path/to.file` to the end of the command to output it wherever you want. On Linux and macOS, most shells support this, on Windows PowerShell does too.

> | **Operator** | **Description** | Syntax |
> |---|---|---|
> | `>` | Send specified stream to a file. | `n>` |
> | `>>` | **Append** specified stream to a file. | `n>>` |
> | `>&1` | _Redirects_ the specified stream to the **Success** stream. | `n>&1` |
> 
> — https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_redirection?view=powershell-7.4#powershell-redirection-operators

Note: `1` refers to STDOUT, while `2` refers to STDERR.

> They're easy to create, easy to diff, and can be opened in a spreadsheet for further analysis.

That's the benefits of logging the results, not the benefit to your solution. I already do this quite easily, I'll show how at the end of this comment.

---

I'll explain how I use the script, which may clear things up. I've intentionally broken one of the plugins to produce a difference before and after.

```sh
# switch to the main branch
git switch main

# run tests and redirect the mismatches to old.txt, creating it if it doesn't exist
node test/regression 2> old.txt

# switch to the branch i want to test
git switch other-branch

# run tests and redirect the mismatches to new.txt, creating it if it doesn't exist
node test/regression 2> new.txt

# sort results so it's easier so it's easier on the eye
# this can also be done in VSC with the "Sort Lines Ascending" command
sort before.txt -o before.txt && sort after.txt -o after.txt
```

Now you can use whatever diffing tool you want, or process the files however suits you.

For example, I can do `diff before.txt after.txt` to see the differences in my terminal.

```diff
$ diff before.txt after.txt 
1c1
< Regression tests failed in 60812ms
---
> Regression tests failed in 62640ms
7a8
> svgs/oxygen-icons-5.113.0/scalable/apps/kalarm.svg is mismatched
17a19
> svgs/oxygen-icons-5.113.0/scalable/apps/preferences-desktop-default-applications.svg is mismatched
```

I could use the Visual Studio Code GUI to see the differences:



In both cases, I can see that the change introduced two new mismatches and didn't resolve any.

johnkenny54: The error was clearly Playwright, you should be able to see it yourself if you set the timeout to 1. No idea where the 30000 was coming from, but the error message and stack trace were clear. It does look like you can set a default timeout on the context, I'll try that, it would be cleaner to do it one place.

As for the other stuff, I agree it would be nice if this was more elegant, but I'm not likely to put effort into that anytime soon, there are plenty of bugs and performance issues to deal with. You can close this PR, though it would be nice to have tmp or something like it in .gitignore so I could stash files without having to remember to keep them out of commits.

SethFalco: > The error was clearly Playwright, you should be able to see it yourself if you set the timeout to 1.

I know I can trigger it myself if I set a low timeout, but just don't understand why it'd happen when it's unset. 🤔

This is something that needs to be reported upstream if `0` (the default) isn't behaving as no timeout. Or alternatively, if the default isn't actually `0` when it's documented as such.

> but I'm not likely to put effort into that anytime soon

No problem at all. I appreciate that you put forward your ideas regardless, and we'll certainly reference parts of it when we come back to this.

> though it would be nice to have tmp or something like it in .gitignore

There's no need for us to have this. You can define a global git ignore somewhere and configure git to use it.

```sh
# create a file called ".gitignore" in your home directory, i.e. C:\Users\UwU\.gitignore
# (over)write "tmp/" to the .gitignore file
echo tmp/ > "%USERPROFILE%\.gitignore"

# configure git to globally apply your gitignore file
git config --global core.excludesFile "%USERPROFILE%\.gitignore"
```

Now across all projects, you can comfortably make a sandbox `tmp` directory that is ignored.

johnkenny54: Thanks for the tip on .gitignore, that makes a lot of things easier.

As for Playwright, I removed the timeout code and haven't had any problems, which had me befuddled until I remembered that you had updated the Playwright version recently. So it seems likely the problems I was having were with the old version, and the timeout settings are no longer necessary.

Issue #1941, opened 2024-01-22
prefixIds fails to prefix when classes and comments are on the same line


john-neptune:
**Describe the bug**

This works:
<style>
.cls-1 {}

.cls-2 {}
</style>

This fails:
<style>
.cls-1 {} 
.cls-2 {}
</style>

.cls-1 would be prefixed, but .cls-2 and subsquent classes would not.

**To Reproduce**

Create SVG with:
<style>
.cls-1 {} 
.cls-2 {}
</style>

Run the prefixIds plugin on the SVG.

**Expected behavior**
All classes should be prefixed.

**Desktop (please complete the following information):**

- SVGO Version: 3.2.0
- NodeJs Version: 20.9.0
- OS: Sonoma 14.2.1

**Additional context**
I have a fix and will PR shortly.

SethFalco:
Just a heads-up! Markdown usually supports safe HTML, so your issue is a lot easier to read when you put code in code blocks! The comments were actually hidden before.

Thanks for reporting the issue and submitting a patch!

john-neptune:
Yep, for sure.  I just missed that.  Thanks for the assist!

PR #1942, opened 2024-01-22
Fix/prefix ids sameline comments

john-neptune:
https://github.com/svg/svgo/issues/1941

Fix prefixing classes on same line as a comment

Before, this worked:
```
<style>
.cls-1 {}
<!-- foo -->
.cls-2 {}
</style>
```

but this failed:
```
<style>
.cls-1 {} <!-- foo -->
.cls-2 {}
</style>
```

.cls-1 would be prefixed, but .cls-2 and subsquent classes would not.

The root cause is an empty return that was missed during a refactor to
switch from forEach() with a lambda to a for () loop.

Add a new test to cover this case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ coverage
 .DS_Store
 .vscode
 *.log
+.*.sw?
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,11 @@ coverage
 .DS_Store
 .vscode
 *.log
-.*.sw?
+
+# https://github.com/github/gitignore/blob/main/Global/Vim.gitignore
+[._]*.s[a-v][a-z]
+!*.svg  # comment out if you don't need vector files
+[._]*.sw[a-p]
+[._]s[a-rt-v][a-z]
+[._]ss[a-gi-z]
+[._]sw[a-p]
--- a/plugins/prefixIds.js
+++ b/plugins/prefixIds.js
@@ -183,7 +183,6 @@ export const fn = (_root, params, info) => {
             });
 
             child.value = csstree.generate(cssAst);
-            return;
           }
         }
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -15,3 +15,4 @@ coverage
 .DS_Store
 .vscode
 *.log
+.*.sw?
```
KTibow, line 18: what kind of file is this? is this exclusion a change that you can't just keep on your local machine?

john-neptune: vim is a pretty common editor.  When vim opens a file like foo.txt, it creates a temporary file named .foo.txt.swp which holds editor state.  They shouldn't be committed.  Seems to fit with excluding .DS_Store, .vscode, etc.

If you don't want to keep this part, that's okay.

SethFalco (approved):
LGTM! Thanks for reporting the issue and submitting a patch. This will be in the next release of SVGO.

Also, sorry for breaking your implementation during the rebase of https://github.com/svg/svgo/pull/1736. ^-^'

.gitignore
SethFalco: Before merging, I'm cool with excluding Vi/Vim/Neovim swap files, but instead of this rule, perhaps we should take the rules from [github/gitignore](https://github.com/github/gitignore/blob/main/Global/Vim.gitignore):

```suggestion

# https://github.com/github/gitignore/blob/main/Global/Vim.gitignore
[._]*.s[a-v][a-z]
!*.svg  # comment out if you don't need vector files
[._]*.sw[a-p]
[._]s[a-rt-v][a-z]
[._]ss[a-gi-z]
[._]sw[a-p]
```

Would you mind doing that as part of this PR?

For future reference, you can take the approach from the comment below too. The feature was actually intended for editor-specific files:

* https://github.com/svg/svgo/pull/1932#discussion_r1445458975

john-neptune: I accepted the suggested change.

Yes, I am aware of the .gitignore hierarchy.  My preference is to favor making a project self-contained for really common bits to make it harder for people, especially occasional committers, to accidentally commit incorrect stuff.  But also, I am a rare committer on this project and yield 100% to the maintainers on this point.

No worries on the rebasing.  Not even completely sure it was the rebasing that broke it.  But we have a fix and a test now!  Thanks for the quick review!

PR #1943, opened 2024-01-25
chore: add optional schema prop to plugins

SethFalco:
Following conversations with maintainers of third-party clients using SVGO, something that has come up is that SVGO does not expose its default behavior or accepted parameters in a way other clients can act on.

This PR introduces a new optional property for plugins called `schema`, which is a JSON Schema that describes the parameters a plugin accepts and any constraints it may have.

This is intended to imitate [Option Schemas](https://eslint.org/docs/latest/extend/custom-rules#options-schemas) in ESLint. Custom Rules in ESLint can expose a `schema` property, which is a JSON Schema that describes the options the rule supports. Exactly what this PR is adding.

This has numerous uses:

* Third-party clients can now check what parameters are available or validate against them.
* Parameter documentation will live closer to the parameters.
  * svg/svgo.dev will have to move the exported `schema` property to the plugins respective documentation page.
* We can improve the help pages in CLI by describing the parameters in more detail, including constraints.

This is just a minimal migration/implementation of these types, but there is more to do to clean things up, improve constraints, etc. 

To notable things we should also do (doesn't have to be in the JSON Schema):

* Provide human-readable names for plugins.
* Expose which plugins are enabled by default. (`preset-default` doesn't do this atm)

## Related

* https://github.com/jakearchibald/svgomg/issues/431#issuecomment-1874980317
* https://github.com/sonnyp/OhMySVG/pull/10#discussion_r1452783046

## CC

@jakearchibald


KTibow:
I would rather schema be generated from TS or TS be generated from schemas than have both in the same codebase.

SethFalco:
@KTibow See: https://github.com/jakearchibald/svgomg/issues/431#issuecomment-1834054041

KTibow:
Right, well there are libraries that let you get a typedef rather easily from the JSON schemas introduced in this PR.
> ## FromSchema
> 
> The FromSchema method lets you infer TS types directly from JSON schemas:
> 
> ```ts
> import { FromSchema } from "json-schema-to-ts";
> 
> const dogSchema = {
>   type: "object",
>   properties: {
>     name: { type: "string" },
>     age: { type: "integer" },
>     hobbies: { type: "array", items: { type: "string" } },
>     favoriteFood: { enum: ["pizza", "taco", "fries"] },
>   },
>   required: ["name", "age"],
> } as const;
> 
> type Dog = FromSchema<typeof dogSchema>;
> // => Will infer the same type as above
> ```

SethFalco:
Interesting, that has a really cool interface. Though, that would also mean shipping a dependency on `json-schema-to-ts` which is a bit eh.

In general, I would prefer a static solution if we were going to do something like that. Regardless, I've given this one a brief look, and it looks like it only works in TypeScript sources, so the particular library isn't going to work for us anyway

I can get the library to work easily in TypeScript sources, but in JavaScript, nope. I can literally take the same JavaScript file, and tsc will be fine if the plugin ends with `.ts` but will complain if it ends with `.js` even though I have the same and valid code in both cases.

plugins/cleanupListOfValues.js
```diff
@@ -1,8 +1,45 @@
...
+export const schema = {
+  type: 'object',
+  properties: {
+    floatPrecision: {
+      title: 'Remove',
+      description:
+        'Number of decimal places to round to, using conventional rounding rules.',
+      type: 'number',
+      default: 3,
+    },
```
jakearchibald, line 20: Since I'd want to represent this as a slider, what maximum value would you be happy with here?

SethFalco: I'm 99% sure we have a limit of 20 somewhere for `floatPrecision`. However, I wanted some time to peek around in case it varies between plugins.

For now, I'll assume it'll be `minimum: 0, maximum: 20, multipleOf: 1`. But it's always possible that in this case or others the conclusion is that there shouldn't be a maximum.

jakearchibald:
I'll work on the schemas. I guess there'll be some work to make the docs pull from these schemas too?

Right now, `title` is only in the mdx files. I'm going to move that into the plugin module too, since I need that text in SVGOMG.

jakearchibald:
Is there a particular reason `applyTransforms.js` doesn't have an exported `name`?

SethFalco:
@jakearchibald @KTibow 

It was not forgotten. The plan was to eventually make it a standalone plugin, and the logic was migrated out of `convertPathData` until it's considered ready to be used as such.

I don't know what was pending before it's considered ready to be a standalone plugin, though. However, it is not part of the public API.

jakearchibald:
Cool. I'll use the absence of the `name` to detect plugins that aren't public.

jakearchibald:
Ohh nice! Is the order of plugins important? `preset-default` seems to have them in a specific order, whereas `builtin` has them in alphabetical order.

SethFalco:
> Ohh nice! Is the order of plugins important? `preset-default` seems to have them in a specific order, whereas `builtin` has them in alphabetical order.

Shoot… you're right. I was actually thinking about doing the following as part of the requirement:

> Expose which plugins are enabled by default. (preset-default doesn't do this atm)

We can update the `createPrefix` utility in SVGO to add a property which we'll export, containing an array of plugins in the preset in the order they execute in. The `createPreset` function also receives said array, so it's just a matter of returning it.

For example, the following, though I'm happy to discuss the interface. We could just use a `preset` property which is immediately assigned to the plugin array rather than separate properties.

```diff
  export const createPreset = ({ name, plugins }) => {
    return {
      name,
+     preset: true,
+     plugins
      fn: (ast, params, info) => {
        const { floatPrecision, overrides } = params;
        const globalOverrides = {};
        if (floatPrecision != null) {
          globalOverrides.floatPrecision = floatPrecision;
        }
        if (overrides) {
          const pluginNames = plugins.map(({ name }) => name);
          for (const pluginName of Object.keys(overrides)) {
            if (!pluginNames.includes(pluginName)) {
              console.warn(
                `You are trying to configure ${pluginName} which is not part of ${name}.\n` +
                  `Try to put it before or after, for example\n\n` +
                  `plugins: [\n` +
                  `  {\n` +
                  `    name: '${name}',\n` +
                  `  },\n` +
                  `  '${pluginName}'\n` +
                  `]\n`,
              );
            }
          }
        }
        invokePlugins(ast, info, plugins, overrides, globalOverrides);
      },
    };
  };
```

jakearchibald:
Currently I'm exporting the plugins from preset-default: https://github.com/jakearchibald/svgo/blob/chore/schema/plugins/preset-default.js#L39-L76

Then over in SVGOMG:

```js
import { plugins as defaultPlugins } from 'svgo/plugins/preset-default';
import presetDefault from 'svgo/plugins/preset-default';
import { builtin } from 'svgo/lib/builtin';

const builtinSet = new Set(builtin);
builtinSet.delete(presetDefault);

const allPlugins = new Set([
  // Get the initial order from preset-default
  ...defaultPlugins,
  // Add the rest in whatever order
  ...builtinSet,
]);
```

I'm happy to keep doing this, or the preset/plugins thing.

SethFalco:
Before continuing, just want to say thank you for taking this on!

---

> Currently I'm exporting the plugins from preset-default: https://github.com/jakearchibald/svgo/blob/chore/schema/plugins/preset-default.js#L39-L76

I would prefer if we took the approach I proposed (which would achieve a similar interface) as it's easier to maintain on our side in the case that we make more presets, which is an idea that's been thrown around a few times.

With the way you did it, while it does mean the list is defined once which is great, it's simpler if presets didn't have to know to export another property imo.

Then you'd do:

```js
import presetDefault from 'svgo/plugins/preset-default';
import { builtin } from 'svgo/lib/builtin';

const builtinSet = new Set(builtin);
builtinSet.delete(presetDefault);

const allPlugins = new Set([
  // Get the initial order from preset-default
  ...presetDefault.plugins,
  // Add the rest in whatever order
  ...builtinSet,
]);
```

jakearchibald:
No problem, I'll do that.

jakearchibald:
I'm going to add a `title` export from each plugin, which will be the same as the title used in the mdx pages.

For `description`, how do you feel about:

```js
export const description = `Adds attributes to the outer most [`<svg>`](https://developer.mozilla.org/docs/Web/SVG/Element/svg) element in the document. This is not an optimization and will increase the size of SVG documents.`;
```

As in, supporting markdown here. I don't mind either way.

plugins/addAttributesToSVGElement.js
```diff
@@ -1,6 +1,29 @@
...
 
+/** @type {JSONSchema} */
+export const schema = {
+  type: 'object',
+  required: ['attributes'],
+  properties: {
+    attributes: {
+      title: 'Attributes',
+      description:
+        'Attributes to add to the `<svg>` element. If key/value pairs are passed, the attributes are added with the paired value. If an array is passed, attributes are added with no key associated with them.',
```
jakearchibald, line 16: A lot of this description only applies to the JS API, and wouldn't apply to the UI. I'm not sure what to do about that, because the docs need the JS details.

SethFalco:
@jakearchibald 

> I'm going to add a title export from each plugin, which will be the same as the title used in the mdx pages.

I'm not into the idea tbh. I'm happy for SVGO to make itself easier for frontend applications to wrap around, but I don't want to expose that much of a public interface that we'd be exporting the same property multiple times for different clients.

I'd like for our plugins to keep JS friendly plugin IDs, then you could either map them on your end, or use a function to create a UI friendly name from the plugin ID.

It's not the perfect solution, but I think it's better than introducing things that don't serve SVGO's needs into our public API. A frontend application's interface is very flexible, but as a library once we introduce new properties, we're pretty much stuck with it for a while, and exporting a `name` and `title` separately isn't something I want stuck in SVGO for now.

I was also hesitant to add UI-friendly names for individual parameters when I started working on this, but I opted to run with it since the parameter is meant to be a JSON Schema and that's part of the spec.

> For description, how do you feel about:

In general, I'm fine with this. However, we should probably save that for when SVGO is prepared to do it's v4 release. (Drastically changes the structure, length, and markup of the property, which is part of our public interface.)

> A lot of this description only applies to the JS API, and wouldn't apply to the UI. I'm not sure what to do about that, because the docs need the JS details, but they don't make sense in SVGOMG.

You'd probably want to override specific strings on your side for when something we do doesn't make sense for you.

That way, you can use our docs when it's available, but you can override it when your adaptation/wrapper doesn't match our interface 1-2-1.

We definitely can't handle this on our side, as SVGOMG is just one example of a third-party client, and we can't maintain docs here that are specific to your client when other clients might represent this differently.

Probably something like your `t()` (function most i18n libraries export) that lets you swap out, normalize, or localize strings that SVGO give you.

jakearchibald:
> > I'm going to add a title export from each plugin, which will be the same as the title used in the mdx pages.
> 
> I'm not into the idea tbh. I'm happy for SVGO to make itself easier for frontend applications to wrap around, but I don't want to expose that much of a public interface that we'd be exporting the same property multiple times for different clients

My intent was that the docs would use this exported name too, rather than duplicating it.

SethFalco:
> My intent was that the docs would use this exported name too, rather than duplicating it.

When I said duplicating, I meant the `name` vs `title`, two properties that mean the same thing but aimed at different interfaces.

I don't want to clutter the JavaScript interface with these details. That should be managed outside of it, imo. For the docs, we'll either use the exported names, or map them to human-readable names in svgo.dev, but not pollute the JavaScript interface with properties exclusively for documentation/clients. That's better managed externally.

Issue #1946, opened 2024-01-28
Error: EMFILE: too many open files


codermapuche:
**Describe the bug**
When run svgo recursive into a foler tree with huge of vectors files get EMFILE: too many open files

**To Reproduce**
Have a vectors folder with huge ammount of svg files and run:
svgo -f vectors --multipass --pretty -r 
Get:
Error: EMFILE: too many open files, open 'cloudy-day.svg'

**Expected behavior**
Run over all files

**Desktop (please complete the following information):**

- SVGO Version 3.2.0
- NodeJs Version v16.13.0
- OS: Windows 10, 64bits

XhmikosR:
My 2 cents: I've hit this issue in the past with other tools and it seems that using graceful-fs was a quick way to fix the issue.

SethFalco:
I'll close this as a duplicate of #2094.

While this was reported first, the newer ticket is more detailed.

PR #1947, opened 2024-01-28
Fix Bug #1946

codermapuche:
Optimize a large number of vectors sequentially instead of parallel to avoid the simultaneously opened file limit

### DIFF
--- a/lib/svgo/coa.js
+++ b/lib/svgo/coa.js
@@ -303,20 +303,27 @@ function optimizeFolder(config, dir, output) {
 function processDirectory(config, dir, files, output) {
   // take only *.svg files, recursively if necessary
   var svgFilesDescriptions = getFilesDescriptions(config, dir, files, output);
-
-  return svgFilesDescriptions.length
-    ? Promise.all(
-        svgFilesDescriptions.map((fileDescription) =>
-          optimizeFile(
-            config,
-            fileDescription.inputPath,
-            fileDescription.outputPath,
-          ),
-        ),
-      )
-    : Promise.reject(
-        new Error(`No SVG files have been found in '${dir}' directory.`),
-      );
+  
+  if ( svgFilesDescriptions.length == 0 ) {
+    return Promise.reject(new Error('No SVG files have been found in "' + dir + '" directory.'));
+  }
+  
+  var semaphore = Promise.resolve([ ]);
+  
+  svgFilesDescriptions.reduce(function(semaphore, fileDescription) {
+    return semaphore.then(function(results) {
+      return optimizeFile(
+        config,
+        fileDescription.inputPath,
+        fileDescription.outputPath,
+      ).then(function(result) {
+        results.push(result);
+        return results;        
+      });      
+    });
+  }, semaphore);
+  
+  return semaphore;
 }
 
 /**


KTibow:
I don't like how this code is formatted. Just make it asynchronous if you want to use Promises, it makes it easier to read.

codermapuche:
@KTibow Feel free to format it however you prefer, I just tried using the tool, it broke and I fixed it, I hope this code helps you improve it faster

SethFalco:
> I don't like how this code is formatted. Just make it asynchronous if you want to use Promises, it makes it easier to read.

Agreed, this is an odd way to approach it in JavaScript. I'm guessing it's just habits from other languages like C++.

> Feel free to format it however you prefer, I just tried using the tool, it broke and I fixed it

If you're not prepared to address feedback constructively, it's better to open an issue instead of a pull request. Not to mention, there are literally failing tests.

---

While this does technically resolve the problem, it's also problematic to drop the parallelization.

* We should continue handling multiple files at once, either batching according to cores count, using an arbitrary number, or using a library like [`graceful-fs`](https://www.npmjs.com/package/graceful-fs) which handles it automatically via backoff.
* If we do any solution other than use graceful-fs, we should still check for the file limit error and provide a helpful error message.

In any case, happy to address formatting later, but could you please resolve linting and tests? Feel free to read the [CONTRIBUTING guide](https://github.com/svg/svgo/blob/main/CONTRIBUTING.md) for more context.

lib/svgo/coa.js
```diff
@@ -300,23 +301,32 @@ function optimizeFolder(config, dir, output) {
...
-            fileDescription.outputPath,
-          ),
-        ),
-      )
-    : Promise.reject(
-        new Error(`No SVG files have been found in '${dir}' directory.`),
-      );
+  
+  if ( svgFilesDescriptions.length == 0 ) {
+    throw 'No SVG files have been found in "' + dir + '" directory.';
```
KTibow, line 309: This should still be an `Error`

lib/svgo/coa.js
```diff
@@ -300,23 +301,32 @@ function optimizeFolder(config, dir, output) {
...
+  }
+  
+  let results = [ ],
```
KTibow, line 312: Run Prettier to fix some of the failing tests, it will fix the formatting on a couple lines including this one

lib/svgo/coa.js
```diff
@@ -300,23 +301,32 @@ function optimizeFolder(config, dir, output) {
...
+      cpus = os.cpus().length;
+
+  while ( svgFilesDescriptions.length > 0 ) {
+    let _svgFilesDescriptions = svgFilesDescriptions.splice(0, cpus);
+    
+    _svgFilesDescriptions = _svgFilesDescriptions.map(function(fileDescription) {
```
KTibow, line 318: I'm not sure if this line errors or not, I would expect it to since with TS once a variable has a type you don't change it. Usually here you would create a new variable to store these results.

lib/svgo/coa.js
```diff
@@ -300,23 +301,32 @@ function optimizeFolder(config, dir, output) {
...
-    : Promise.reject(
-        new Error(`No SVG files have been found in '${dir}' directory.`),
-      );
+  
+  if ( svgFilesDescriptions.length == 0 ) {
+    throw 'No SVG files have been found in "' + dir + '" directory.';
+  }
+  
+  let results = [ ],
+      cpus = os.cpus().length;
```
KTibow, line 313: It's worth noting that since SVGO doesn't run in parallel it might not make sense to scale this by CPUs

lib/svgo/coa.js
```diff
@@ -300,23 +301,32 @@ function optimizeFolder(config, dir, output) {
...
+    
+    _svgFilesDescriptions = _svgFilesDescriptions.map(function(fileDescription) {
+      return optimizeFile(
+        config,
+        fileDescription.inputPath,
+        fileDescription.outputPath,
+      );   
+    });
+    
+    results = results.concat(await Promise.all(_svgFilesDescriptions));
```
KTibow, line 326: This is slower than the existing version because each "chunk" of files takes as long as the longest file. I would suggest changing this to a worker-based method, let me know if you want me to explain

lib/svgo/coa.js
codermapuche: Eventloop run in single thread in fact. 
The use of cpu count is a number >= 1 and <= ulimit.
The better number is ulimit, but is only available under linux platform, i dont found a cross-platform way to determine max files open limit.

lib/svgo/coa.js
codermapuche: Yes, this is slower. This PR in fact run a throttle mechanism over chunks, this is the goal. Workers dont fix this, under linux the kernel has a limit over how many files can be open at same time, this limit is shared by all process of user, and another limit is shared for all process. The only way to prevent EMFILE error runing svgo over a huge number of svg is go slower and do the job in chunks.
The performance impact is minimun, runing over folder with 80k+ of svg works well.

lib/svgo/coa.js
codermapuche: Maybe you are right, i dont use TS, i do this changes over the npm downloaded code

lib/svgo/coa.js
KTibow: You're misunderstanding what I said. Right now, it goes like "10 files open - > 6 files open - > 1 file open" if one takes longer than the rest, leading to it taking longer to start the next set of files. If you instead create 10 workers, each one will continously have 1 file open.

PR #1952, opened 2024-02-04
Check for filter in style as well as attribute.

johnkenny54:
collapseGroups was not detecting a filter when it was in a style rather than an attribute.

This change resolves one regression mismatch (svgs/oxygen-icons-5.113.0/scalable/mimetypes/small/64x64/image-x-adobe-dng.svg) and reduces pixel mismatches by 67.

This type of problem seems like it is probably widespread (including in the same conditional in this plugin, where it is looking for a clip-path). I didn't see any tools to query the style attribute in a standard way. The solution in this PR is not going to scale well - it seems like it would be desirable to have a standardized, efficient way to query/change the style attribute - maybe it's there and I'm missing it?

### DIFF
--- a/plugins/collapseGroups.js
+++ b/plugins/collapseGroups.js
@@ -71,7 +71,7 @@ export const fn = (root) => {
           node.children.length === 1
         ) {
           const firstChild = node.children[0];
-          let nodeHasFilter = node.attributes.filter ? true : false;
+          let nodeHasFilter = !!node.attributes.filter;
           if (!nodeHasFilter && node.attributes.style) {
             if (computeStyle(stylesheet, node)['filter']) {
               nodeHasFilter = true;
--- a/plugins/collapseGroups.js
+++ b/plugins/collapseGroups.js
@@ -1,3 +1,4 @@
+import { parseStyleDeclarations } from '../lib/style.js';
 import { inheritableAttrs, elemsGroups } from './_collections.js';
 
 /**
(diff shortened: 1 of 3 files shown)


KTibow:
Why do this instead of getComputedStyle?

johnkenny54:
> Why do this instead of getComputedStyle?

Are you talking about `computeStyle()`? This seems like it has a huge overhead.

SethFalco:
> Are you talking about computeStyle()? This seems like it has a huge overhead.

I'd recommend using `computeStyle` too. It will also account for the stylesheet.

plugins/collapseGroups.js
```diff
@@ -68,11 +71,17 @@ export const fn = () => {
           node.children.length === 1
         ) {
           const firstChild = node.children[0];
+          let nodeHasFilter = node.attributes.filter ? true : false;
```
XhmikosR, line 74: This could probably be simplified to `let nodeHasFilter = node.attributes.filter`

SethFalco: If we were to change this, I'd prefer this so it's explicitly a boolean:

```suggestion
          let nodeHasFilter = !!node.attributes.filter;
```

Naming a variable after a boolean when it's actual a truthy/falsey makes the code confusing imo.

johnkenny54: TypeScript doesn't like `let nodeHasFilter = node.attributes.filter`.

KTibow: Also `Boolean(expr)` is always an option

SethFalco (commented):
Left a minor comment, but looks good overall!

plugins/collapseGroups.js
```diff
@@ -68,11 +71,17 @@ export const fn = () => {
           node.children.length === 1
         ) {
           const firstChild = node.children[0];
+          let nodeHasFilter = !!node.attributes.filter;
+          if (!nodeHasFilter && node.attributes.style) {
+            if (computeStyle(stylesheet, node)['filter']) {
+              nodeHasFilter = true;
+            }
+          }
```
SethFalco, line 79:
```suggestion
          const nodeHasFilter = !!(
            node.attributes.filter || computeStyle(stylesheet, node).filter
          );
```

Now that you're using `computeStyle`, and considering that it can have a `filter` applied without being in the `style` attribute, you can drop that check for if the `style` attribute is defined.

I think it's simplest to stick a `const` and do both checks in the same line. It should mean the exact same thing as what you have now.

PR #1958, opened 2024-02-09
chore: Update the README docs to suggest svgo.config.mjs

anthonyryan1:
When running svgo with a custom config named svgo.config.js on node 20.11.0. It fails to load the config file with the following error:

```
./svgo.config.js:1
export default {
^^^^^^

SyntaxError: Unexpected token 'export'
    at internalCompileFunction (node:internal/vm:77:18)
    at wrapSafe (node:internal/modules/cjs/loader:1288:20)
    at Module._compile (node:internal/modules/cjs/loader:1340:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at cjsLoader (node:internal/modules/esm/translators:356:17)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:305:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)
```

Renaming it to `.mjs` makes node process it properly

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,9 @@ svgo --help
 
 SVGO has a plugin architecture. You can read more about all plugins in [Plugins | SVGO Documentation](https://svgo.dev/docs/plugins/), and the default plugins in [Preset Default | SVGO Documentation](https://svgo.dev/docs/preset-default/).
 
-SVGO reads the configuration from `svgo.config.js` or the `--config path/to/config.js` command-line option. Some other parameters can be configured though command-line options too.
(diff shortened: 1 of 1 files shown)


SethFalco (approved):
I've tested locally, and this works as intended in the following scenarios:

* Not in a project.
* In a Common JS project.
* In an ESM project.

In future, I'd like for us to document both the CJS and ESM config, but I'm happy with this for now as it's clearer across project (or non-project) setups.

Thanks for the contribution!

PR #1959, opened 2024-02-11
fix: <a> element not parsed as text

johnkenny54:
`<a>` elements should be parsed the same as `<tspan>`, etc. This change adds `<a>` to the textContent collection. It also adds a test file for the parser, updates the expected results for several test cases, and changes the removeScriptElement plugin to remove empty text nodes contained by removed `<a>` elements.

To see the problem visually, optimize the following file with default SVGO settings and view both input and output files:

```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
    <text x="10" y="20">this is a test</text>
    <text x="10" y="35" xml:space="preserve">
        <a href="https://svgwg.org/svg2-draft/linking.html#Links">
this is a test   
        </a>
    </text>
</svg>
```

Before the fix, the second line is shifted from the correct position after being optimized.

### DIFF
--- a/lib/parser.test.js
+++ b/lib/parser.test.js
@@ -1,0 +1,32 @@
+import { parseSvg } from './parser.js';
+import { stringifySvg } from './stringifier.js';
+
+const input = `<svg xmlns="http://www.w3.org/2000/svg">
+<text x="10" y="35" xml:space="preserve">
+    <a href="x">
+this is a test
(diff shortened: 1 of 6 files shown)


SethFalco (approved):
This makes sense, thanks for spotting and addressing this issue.

PR #1961, opened 2024-02-12
fix: stringifyPathData needs space before scientific notation.

johnkenny54:
stringifyPathData was not adding a space before numbers which used scientific notation.

This fix does not change the results of regression tests using the current value of floatPrecision: 4 (125 failures, 3794 pixel differences), but when run at floatPrecision:7, failures are reduced from 164 to 93, and pixel mismatches reduced from 32,058 to 3,233.

### DIFF
--- a/lib/path.js
+++ b/lib/path.js
@@ -285,9 +285,7 @@ const stringifyArgs = (command, args, precision, disableSpaceAfterFlags) => {
       result += roundedStr;
     } else if (
       !Number.isInteger(previous) &&
-      rounded != 0 &&
-      rounded < 1 &&
-      rounded > -1
+      !'0123456789'.includes(roundedStr[0])
     ) {
       // remove space before decimal with zero whole
       // only when previous number is also decimal
--- a/lib/path.js
+++ b/lib/path.js
@@ -283,10 +283,7 @@ const stringifyArgs = (command, args, precision, disableSpaceAfterFlags) => {
     } else if (i === 0 || rounded < 0) {
       // avoid space before first and negative numbers
       result += roundedStr;
-    } else if (
-      !Number.isInteger(previous) &&
-      !isDigit(roundedStr[0])
-    ) {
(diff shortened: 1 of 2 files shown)


lib/path.js
```diff
@@ -285,9 +285,7 @@ const stringifyArgs = (command, args, precision, disableSpaceAfterFlags) => {
       result += roundedStr;
     } else if (
       !Number.isInteger(previous) &&
-      rounded != 0 &&
-      rounded < 1 &&
-      rounded > -1
+      !'0123456789'.includes(roundedStr[0])
```
SethFalco, line 288: This file had a `isDigit` function defined. Could you use that instead?

```suggestion
      !isDigit(roundedStr[0])
```

This should be the same thing, but runs notably faster.

<details>

<summary> Benchmark: </summary>

```js
const Benchmark = require('benchmark');
const suite = new Benchmark.Suite;

const data = [
  '0',
  '1',
  '-',
  '.',
]

const isDigit = (c) => {
  const codePoint = c.codePointAt(0);
  if (codePoint == null) {
    return false;
  }
  return 48 <= codePoint && codePoint <= 57;
};

suite.add('#includes', () => {
  data.map((d) => '0123456789'.includes(d));
})
.add('between chars', () => {
  data.map((d) => d >= '0' && d <= '9');
})
.add('between chars v2', () => {
  data.map(isDigit);
})
.on('cycle', (event) => {
  console.log(String(event.target));
})
.on('complete', function() {
  console.log('Fastest is ' + this.filter('fastest').map('name'));
})
.run({ 'async': true });
```

```
#includes x 27,552,035 ops/sec ±1.06% (92 runs sampled)
between chars x 40,564,628 ops/sec ±1.23% (94 runs sampled)
between chars v2 x 57,363,760 ops/sec ±0.55% (94 runs sampled)
Fastest is between chars v2
```

</details>

SethFalco (commented):
Makes sense, but I've made a suggestion. Happy to merge once that's addressed.

PR #1963, opened 2024-02-13
Optionally convert rects with rounded corners to paths

baarde:
Currently, rectangles with rounded corners are excluded from the `convertShapeToPath` optimization.

This PR adds the conversion of rectangles with rounded corners to paths when the `convertArcs` option is enabled.

Though the converted `<path>` is quite longer than the source `<rect>`, the conversion is still useful to people who wish to merge all the shapes into a single path.

### DIFF
--- a/docs/03-plugins/convert-shape-to-path.mdx
+++ b/docs/03-plugins/convert-shape-to-path.mdx
@@ -5,7 +5,7 @@ svgo:
   defaultPlugin: true
   parameters:
     convertArcs:
-      description: If to convert <code>&lt;circle&gt;</code> and <code>&lt;ellipse&gt;</code> elements to paths.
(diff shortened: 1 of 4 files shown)


baarde:
### Special cases
* `rx` is absent → `ry` is used instead.
*  `ry` is absent → `rx` is used instead.
* `rx` or `ry` is equal to zero → The `<rect>` is converted without rounded corners.
* `rx` or `ry` is negative or not a number → The `<rect>` is not converted.
* The computed value of `rx` is greater than half the `width` → `rx` is clamped to half the `width`.
* The computed value of `ry` is greater than half the `height` → `ry` is clamped to half the `height`.

GreLI:
AFAIK, I wrote that plugin and excluded it just for the fact it makes paths longer than rects. Nothing to do with optimization.

dr-itz:
> AFAIK, I wrote that plugin and excluded it just for the fact it makes paths longer than rects. Nothing to do with optimization.

while the  `path` version is longer than the `rect`, it would be great to have this as an option (not as default). Because tools like fontello.com cannot handle anything other than plain `path`...

Issue #1951, opened 2024-02-03
Remove trailing M


KTibow:
**Is your feature request related to a problem? Please describe.**
Some paths have trailing move commands. These can be removed to optimize the path.

**Describe the solution you'd like**
Remove trailing move commands. (Consider excluding gradient fills)

**Describe alternatives you've considered**
None

**Additional context**
None

PR #1964, opened 2024-02-15
fix: mergePaths should not merge paths with styles that depend on bounding box

### LINKED ISSUE #1267: Combining multiple shapes into a <path> breaks <linearGradient>

johnkenny54:
This PR fixes the problem raised in issue #1267. The general issue is that merging paths into a single element changes the bounding box, so any style that depends on the bounding box of the element may change behavior. This PR disables merging for any paths that have:

- A fill, filter, or stroke that references a URL.
- A clip-path or mask.

It's possible that there are other scenarios where merging should be disabled, but those above are the only ones that I was able to construct a test case for so far.

With the current regression settings, there is no change in pixel mismatches in the regressions. With the Oxygen files, only one file seems to be affected - scalable/status/weather-few-clouds-night.svg. The clouds were rendered incorrectly in the optimized version before the fix.

Closes #1267

### DIFF
--- a/plugins/mergePaths.js
+++ b/plugins/mergePaths.js
@@ -16,29 +16,14 @@ export const description = 'merges multiple paths in one if possible';
 /**
  * @param {XastElement} element
  * @param {ComputedStyles} computedStyle
+ * @param {string} attName
  */
-function elementHasGradientFill(element, computedStyle) {
-  /**
-   * @param {string} str
-   */
-  function hasGradientFill(str) {
-    return Boolean(str) && str.includes('url(');
+function elementHasGradient(element, computedStyle, attName) {
+  const style = computedStyle[attName];
+  if (!style || style.type !== 'static') {
+    return false;
   }
-
-  /**
-   * @param {StaticStyle|DynamicStyle|undefined} style
-   */
-  function hasGradientFillStyle(style) {
-    if (!style || style.type !== 'static') {
-      return false;
-    }
-    return hasGradientFill(style.value);
-  }
-
-  return (
-    hasGradientFill(element.attributes.fill) ||
-    hasGradientFillStyle(computedStyle['fill'])
-  );
+  return style.value && style.value.includes('url(');
 }
 
 /**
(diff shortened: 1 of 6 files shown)


plugins/mergePaths.js
```diff
@@ -10,6 +13,19 @@ import { path2js, js2path, intersects } from './_path.js';
...
+ * @param {XastElement} element
+ * @param {ComputedStyles} computedStyle
+ * @param {string} attName
+ */
+function elementHasURL(element, computedStyle, attName) {
+  const style = computedStyle[attName];
+  if (!style || style.type !== 'static') {
+    return false;
+  }
+  return style.value && style.value.includes('url(');
```
SethFalco, line 26: Perhaps this should be refactored to reuse the `includeUrlReference` utility?

Just helps to keep how we determine if something has a URL reference consistent. I'd also propose using `Url` instead of `URL` for consistency as well, for example when compared to our existing `includesUrlReference` function.

```suggestion
 * @param {ComputedStyles} computedStyle
 * @param {string} attName
 * @returns {boolean}
 */
function elementHasUrl(computedStyle, attName) {
  const style = computedStyle[attName];

  if (style?.type === 'static') {
    return includesUrlReference(style.value);
  }

  return false;
```

(This would also require changing other references to `elementHasURL`.)

SethFalco (commented):
I've left one minor comment, but overall this looks good. Thanks for spotting the issue and submitting a patch.

SethFalco (approved):

KTibow:
I'm a bit late here but this "is the path bounding box sensitive?" logic would be useful for #1951, and I'm going to move it to a function to reduce duplication. Which file would that function best fit in?

SethFalco:
@KTibow Nothing shouts out to me, so I'd default to `lib/svgo/tools.js`.

PR #1976, opened 2024-04-08
fix(makeProgram): specify .mjs and .cjs config files are allowed

wojtekmaj:
Looking at the source code:

https://github.com/svg/svgo/blob/2442f742398f6c40bcf83b5c34e4a71a023adae1/lib/svgo-node.js#L39-L50

it looks to me like not only .js, but also .mjs, and .cjs extensions are allowed in config files. This PR aligns the help text with the actual behavior.

### DIFF
--- a/lib/svgo/coa.js
+++ b/lib/svgo/coa.js
@@ -49,7 +49,7 @@ export default function makeProgram(program) {
     )
     .option(
       '--config <CONFIG>',
-      'Custom config file. Only .js, .mjs, .cjs is supported',
+      'Custom config file, only .js, .mjs, and .cjs is supported',
     )
     .option(
       '--datauri <FORMAT>',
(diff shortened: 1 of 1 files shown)


lib/svgo/coa.js
```diff
@@ -47,7 +47,10 @@ export default function makeProgram(program) {
       '-p, --precision <INTEGER>',
       'Set number of digits in the fractional part, overrides plugins params',
     )
-    .option('--config <CONFIG>', 'Custom config file, only .js is supported')
+    .option(
+      '--config <CONFIG>',
+      'Custom config file. Only .js, .mjs, .cjs is supported',
```
SethFalco, line 52:
```suggestion
      'Custom config file, only .js, .mjs, and .cjs is supported',
```

Just to keep things consistent with our other help strings and docs.

wojtekmaj: Ah! Also a fan of the Oxford comma. Not sure how I missed that :D

SethFalco (approved):
LGTM! Thanks again! 👍🏽

PR #1977, opened 2024-04-08
feat: add support for `--disable` and `--enable` flags back

### LINKED ISSUE #1356: error: unknown option
I used an old version of SVGO like e.g. [comand](https://github.com/JoKalliauer/cleanupSVG/blob/master/o4compression.sh#L39)

```svgo -i input.svg -o output.svg --pretty --indent=1 -p 3  --enable=removeHiddenElems --disable=removeUnknownsAndDefaults```

and I got `
error: unknown option '--enable=removeHiddenElems'
`

However it seems svgo is not backwards compatible.

I have read #1329 and [#1337/files](https://github.com/svg/svgo/pull/1337/files), and do not understand what to do.

I have tried to copy svgo.config.js from https://github.com/svg/svgo/issues/1329#issuecomment-781585374 into the current folder and got
```bash
Error: Cannot find module 'svgo'
Require stack:
- /home/jkalliau/Documents/GitHub/cleanupSVG/svgo.config.js
- /home/jkalliau/prgm/node-v15.4.0-linux-x64/lib/node_modules/svgo/lib/svgo-node.js
- /home/jkalliau/prgm/node-v15.4.0-linux-x64/lib/node_modules/svgo/lib/svgo/coa.js
- /home/jkalliau/prgm/node-v15.4.0-linux-x64/lib/node_modules/svgo/bin/svgo
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:925:15)
    at Function.Module._load (node:internal/modules/cjs/loader:769:27)
    at Module.require (node:internal/modules/cjs/loader:997:19)
    at require (node:internal/modules/cjs/helpers:92:18)
    at Object.<anonymous> (/home/jkalliau/Documents/GitHub/cleanupSVG/svgo.config.js:3:34)
    at Module._compile (node:internal/modules/cjs/loader:1108:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1137:10)
    at Module.load (node:internal/modules/cjs/loader:973:32)
    at Function.Module._load (node:internal/modules/cjs/loader:813:14)
    at Module.require (node:internal/modules/cjs/loader:997:19)
```

### LINKED ISSUE #1622: CLI option to disable plugins is no longer available
**Describe the bug**
CLI option to disable plugins is no longer available. It makes at least one (hah, probably many) developer sad.

**To Reproduce**
1. Go to console
2. Type svgo --disable=removeViewBox some.svg
3. Press enter
4. See error - unknown option

**Expected behavior**
Simple working option to remove some plugin by choice. And not removing functionality with an update.

**Screenshots**
svgo --disable=removeViewBox draft.svg
error: unknown option '--disable=removeViewBox'

 - SVGO Version [2.8.0]
 - NodeJs Version [12.22.7]
 - OS: [windows 10]
 
**Additional context**
So sad.

wojtekmaj:
Closes #1356
Closes #1622

In https://github.com/svg/svgo/issues/1622#issuecomment-971440346 we read:

> I consider adding it back at some point but it should be able to work with presets somehow. This is still need to be figured out.

I believe that by using globalOverrides, we can make it very easily, but I must admit, this was the first time I've worked with this code, so I might have missed something.

I did test the code though and it disabled the plugins I wanted when using CLI, so at the very least we're getting somewhere.

I've also increased JSDoc coverage here, mainly for my own sanity when working with these arrays of plugins. Hope you don't mind that scope creep :)

### DIFF
--- a/lib/svgo.js
+++ b/lib/svgo.js
@@ -4,11 +4,22 @@ import { builtin } from './builtin.js';
 import { invokePlugins } from './svgo/plugins.js';
 import { encodeSVGDatauri } from './svgo/tools.js';
 
+/**
+ * @template T {any}
+ * @typedef {import('./types.js').PluginConfig<T>} PluginConfig<T>
+ * @typedef {import('./types.js').PresetConfig} PresetConfig
+ */
+
 const pluginsMap = {};
 for (const plugin of builtin) {
   pluginsMap[plugin.name] = plugin;
 }
 
+/**
+ * @template T {any}
+ * @param {string | PresetConfig| PluginConfig<T>} plugin plugin name or plugin config
+ * @returns {PresetConfig| PluginConfig<T> | null} plugin config or null if plugin was not found
+ */
 const resolvePluginConfig = (plugin) => {
   if (typeof plugin === 'string') {
     // resolve builtin plugin specified as string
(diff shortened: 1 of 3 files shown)


SethFalco:
Thank you for opening the pull request!

I don't see us accepting usage like this in the short-term.

Since the CLI options were removed, the way SVGO is configured has become more like a pipeline. Things like the order of plugins matter, and then approaching things like plugin parameters, or arrays, etc. Disabling plugins could work, but for enabling plugins, it doesn't really make sense in my opinion.

I'd be happy to see us bring back a way to enable/disable plugins from the CLI without the use of a `svgo.config.js` file, but the most likely way this would be accepted is as an in-line JSON version of the SVGO config object either in the `--config` argument, or in a separated `--config-json`/`--config-str`/`--config-inline` argument.

So:

```sh
# use one plugin
svgo file.svg --config-str '{"plugins": ["removeHiddenElems"]}'

# use default plugins plus enable one more plugin to run after preset-default
svgo file.svg --config-str '{"plugins": ["preset-default", "removeViewBox"]}'

# run preset-default but exclude one of the plugins
svgo file.svg --config-str '{"plugins": [{"name": "preset-default", "params": {"overrides": {"removeHiddenElems": false}}}]}'
```

I'm leaving https://github.com/svg/svgo/issues/1356 open, but will consider a resolution in future.

PR #1982, opened 2024-04-17
fix(cleanupIds): handle no-uri-encoded references

### LINKED ISSUE #1981: [cleanupIDs] symbol is incorrectly removed when its ID is *NOT* used by URL-encoded reference
**Describe the bug**

Last year's PR #1880 fixed the issue with URL encoding of IDs in `href` or `url()`. However, in practical use, we found that due to differences in SVG export tools or improper use, the SVG files delivered by designers to developers often have references to graphics, gradients, etc., in `url()` that are not necessarily URL-encoded.

Here is one of my actual examples (with business desensitization):

```svg
  <defs>
    <linearGradient id="渐变_1" x1="0%" y1="0%" x2="100%" y2="0%">
      <stop offset="10%" stop-color="#c71700" />
      <stop offset="60%" stop-color="orange" />
      <stop offset="100%" stop-color="#5a2100" />
    </linearGradient>
  </defs>
  <rect x="30" y="30" height="150" width="370" fill="url(#渐变_1)" />
</svg>
```

After this SVG was processed by svgo, it became like this:
```svg
  <path fill="url(#渐变_1)" d="M30 30h370v150H30z"/>
</svg>
```

I traced the code and found that during the search for reference IDs, only URL-encoded IDs were used, which led to the unencoded IDs not being matched. As a result, in subsequent processing, they were considered as unreferenced, useless nodes and were deleted.

https://github.com/svg/svgo/blob/8d6385bd9ab49d1d300a10268930238baa5eb269/plugins/cleanupIds.js#L236

I will submit a PR shortly to fix this issue.

- SVGO Version 3.2.0
- NodeJs Version 20.12.0
- OS: macOS

liuweifeng:
fixed #1981 .

### DIFF
--- a/plugins/cleanupIds.js
+++ b/plugins/cleanupIds.js
@@ -233,7 +233,7 @@ export const fn = (_root, params) => {
                 if (value.includes('#')) {
                   // replace id in href and url()
                   element.attributes[name] = value.replace(
-                    `#${encodeURI(id)}`,
+                    new RegExp(`#(${id}|${encodeURI(id)})`),
                     `#${currentIdString}`,
                   );
                 } else {
--- a/plugins/cleanupIds.js
+++ b/plugins/cleanupIds.js
@@ -232,10 +232,9 @@ export const fn = (_root, params) => {
                 const value = element.attributes[name];
                 if (value.includes('#')) {
                   // replace id in href and url()
-                  element.attributes[name] = value.replace(
-                    new RegExp(`#(${id}|${encodeURI(id)})`),
-                    `#${currentIdString}`,
-                  );
+                  element.attributes[name] = value
+                    .replace(`#${encodeURI(id)}`, `#${currentIdString}`)
+                    .replace(`#${id}`, `#${currentIdString}`);
                 } else {
                   // replace id in begin attribute
                   element.attributes[name] = value.replace(
--- a/test/plugins/cleanupIds.26.svg.txt
+++ b/test/plugins/cleanupIds.26.svg.txt
@@ -1,6 +1,6 @@
-Should handle non-ASCII IDs and resolve URI encoded references.
+Should handle non-ASCII IDs and resolve not correctly URI encoded references.
 
-See: https://github.com/svg/svgo/issues/1696
+See: https://github.com/svg/svgo/issues/1981
 
 ===
 
--- a/test/plugins/cleanupIds.26.svg.txt
+++ b/test/plugins/cleanupIds.26.svg.txt
@@ -7,9 +7,7 @@ See: https://github.com/svg/svgo/issues/1981
 <svg xmlns="http://www.w3.org/2000/svg">
     <defs>
         <linearGradient id="渐变_1" x1="0%" y1="0%" x2="100%" y2="0%">
-			      <stop offset="10%" stop-color="#c71700" />
-            <stop offset="60%" stop-color="orange" />
-            <stop offset="100%" stop-color="#5a2100" />
+            <stop stop-color="#5a2100" />
         </linearGradient>
     </defs>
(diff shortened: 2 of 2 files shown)


plugins/cleanupIds.js
```diff
@@ -233,7 +233,7 @@ export const fn = (_root, params) => {
                 if (value.includes('#')) {
                   // replace id in href and url()
                   element.attributes[name] = value.replace(
-                    `#${encodeURI(id)}`,
+                    new RegExp(`#(${id}|${encodeURI(id)})`),
                     `#${currentIdString}`,
                   );
```
SethFalco, line 238: Something like this would be better:

```js
                  element.attributes[name] = value
                    .replace(`#${encodeURI(id)}`, `#${currentIdString}`)
                    .replace(`#${id}`, `#${currentIdString}`);
```

The proposed version opens us up to a potential ReDoS, as a user provides the SVG which we'd derive both the regex and value. A malicious user could upload a vulnerable SVG to a server that uses SVGO to abuse this.

Let's keep things simple and just chain two replaces. 👍🏽

test/plugins/cleanupIds.26.svg.txt
```diff
@@ -0,0 +1,29 @@
...
+    <rect x="30" y="30" height="150" width="370" fill="url(#渐变_1)" />
+</svg>
+
+@@@
+
+<svg xmlns="http://www.w3.org/2000/svg">
+    <defs>
+        <linearGradient id="a" x1="0%" y1="0%" x2="100%" y2="0%">
+            <stop offset="10%" stop-color="#c71700"/>
+            <stop offset="60%" stop-color="orange"/>
```
SethFalco, line 24:
```suggestion
            <stop offset="100%" stop-color="#5a2100" />
        </linearGradient>
    </defs>
    <rect x="30" y="30" height="150" width="370" fill="url(#渐变_1)" />
</svg>

@@@

<svg xmlns="http://www.w3.org/2000/svg">
    <defs>
        <linearGradient id="a" x1="0%" y1="0%" x2="100%" y2="0%">
            <stop offset="10%" stop-color="#c71700"/>
```

Just to keep the tests minimal, can we omit the two lines with offset 60%? We prefer for the tests to be as small as possible, so they just focus on what they're intended to test.

SethFalco (commented):
I've left a comment as I believe there is a vulnerability in this version.

Overall, the issue and solution makes sense though. Thanks for submitting a patch!

plugins/cleanupIds.js
liuweifeng: Thank you for your review. It was indeed an oversight on my part, and I have already submitted the fix to the current PR.

SethFalco: Thanks for addressing this.

Just leaving more context here, as I received an email asking about this vulnerability, and as it was never merged, I see no issue with demonstrating how to abuse it publically for anyone interested.

**pwned.svg**
```svg
  <defs>
    <linearGradient id="(#+)+$" x1="0%" y1="0%" x2="100%" y2="0%">
      <stop offset="10%" stop-color="#c71700" />
    </linearGradient>
  </defs>
  <rect x="30" y="30" height="150" width="370" fill="############################X url('#(#+)+$')" />
</svg>
```

Then one way or another, a computer attempts to optimize this SVG. In CLI this would be with:

```sh
svgo pwned.svg -o out.svg
```

It would identify an element with an ID of `(#+)+$`, an evil regex, then match it to the value of `############################X url('#(#+)+$')` further down, with each additional `#` doubling the time it takes to process the SVG.

The above took 15 seconds on my computer (16 threads), so with just 7 more `#` symbols would take it to an hour to process.

SethFalco (approved):

SethFalco:
Thanks for spotting the issue and submitting a patch! This will be released by the end of the month.

PR #1984, opened 2024-05-08
Fixed V3.3.0 Common JS missing loadConfig and optimize exports

nuintun:
esm
```ts
import { loadConfig, optimize } from 'svgo';
```

cjs
```ts
const { loadConfig, optimize } = require('svgo');
```

browser
```ts
import { optimize } from 'svgo/browser';
```

### DIFF
--- a/lib/svgo-node.d.ts
+++ b/lib/svgo-node.d.ts
@@ -1,0 +1,17 @@
+import { Config, optimize } from './svgo';
+
+export { optimize };
+
+/**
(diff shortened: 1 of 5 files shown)


SethFalco (approved):
Thanks for investigating this as soon as you spotted the issue, and sorry for letting that through!

I'll be sure to check if there's any other public interfaces that we need to cover with tests too.

I've added a commit to your PR to at least add one test case, just to verify that `loadConfig` is defined. Once CI passes, happy to merge and I'll make a release. 👍🏽

PR #1985, opened 2024-05-08
chore: add package-lock.json to `.gitignore`

XhmikosR:


SethFalco:
Can you explain these changes, please? We use yarn, not npm, so I'm not sure why `package-lock.json` would need disabling.

XhmikosR:
Yeah, the thing is that which these changes even if one does npm install,
they'll get what they need to without the extra lock file.

I don't use yarn a lot and I hit this myself.

It doesn't hurt anything, it's just more explicit even for people who
accidentally use npm.

SethFalco:
In my opinion, the con is that it adds clutter to the repo. My thoughts are in a similar vain to why we don't include editor specific configs like `.vscode` in the repository which has come up before.

If we include configs in the repo for tooling that isn't a part of our stack, the files tend to accumulate. This repo does not use npm, so it doesn't make sense to include an `.npmrc`.

I'm use to using npm too and even prefer it, but rather than making changes in the repo, I've just ingrained a habit to check for lock files before contributing to a project. I believe this is a better solution than having project that use Yarn configure npm.

This is especially relavent considering for me at least, running `npm install` makes undesirable changes to the `yarn.lock` file, other commands like when publishing also behave differently. Only Yarn is supported for contributions, and while others are welcome to use whatever tools they'd like, and I don't want us to have files that suggest otherwise.

XhmikosR:
That's OK, but let's add it to .gitignore at least.

PR #1987, opened 2024-05-08
chore: update devDependencies

XhmikosR:
Requires #1986 and a rebase

PR #1986, opened 2024-05-08
chore: remove del devDependency

XhmikosR:

PR #1988, opened 2024-05-08
chore: add cross-env and rimraf so that tests work everywhere

XhmikosR:
Without this tests fail on Windows cmd (CI works because by default it's using bash)

Requires #1986 and #1987.


SethFalco (approved):
LGTM! Thanks for the contribution!

PR #1989, opened 2024-05-08
chore: update yarn to v3.8.2

XhmikosR:
Requires #1988 and a rebase


SethFalco:
LGTM! 👍🏽 

Just so others are aware, as this involves large obuscated files. I did verify that running the following commands locally produces the same result:

```sh
yarn set version 3.8.2
yarn install
yarn run fix
```

PR #1994, opened 2024-05-08
Fix "purly" typo

kevinebaugh:
Noticed here: https://svgo.dev/docs/plugins/remove-title/#:~:text=The%20SVG%20is%20purly%20aesthetic


SethFalco (approved):
Thanks for spotting the issue and submitting the fix!

PR #1997, opened 2024-05-09
fix: fixed Config definition for node

nuintun:
Re export `Config` definition for node module.

image

### DIFF
--- a/lib/svgo-node.d.ts
+++ b/lib/svgo-node.d.ts
@@ -1,6 +1,6 @@
-import { VERSION, Config, optimize, builtinPlugins } from './svgo';
+import { Config } from './svgo';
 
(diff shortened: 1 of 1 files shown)


XhmikosR:
Just thinking out loud here, can't you just export from directly instead of importing and then exporting?

nuintun:
@XhmikosR 

```ts
import { Config, optimize } from './svgo';

export { Config, optimize };
```

are equal to

```ts
export { Config, optimize } from './svgo';
```

Directly export is just a syntax sugar for reexport variables that are not used in the current code.

SethFalco:
Hey, thanks for submitting the PR! I've amended the commit to export everything instead, which matches the exports in previous versions.

PR #1999, opened 2024-05-09
fix: export package.json and dist/svgo.browser.js again

### LINKED ISSUE #1998: `package.json`/ `dist/svgo.browser.js` are no longer exported
Using v3.3.1:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in C:\Users\xmr\Desktop\svgomg\node_modules\svgo\package.json
```

Adding this in the exports fixes the issue but this means that if people where requiring other things that the ones exported right now, they will be broken:

```
"./package.json": "./package.json"
```

There must be a way to export everything again so that it's not a breaking change.

XhmikosR:
Fixes #1998

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,10 @@
       "require": "./dist/svgo-node.cjs",
       "types": "./lib/svgo-node.d.ts"
     },
+    "./dist/svgo.browser.js": {
+      "import": "./dist/svgo.browser.js",
+      "types": "./lib/svgo.d.ts"
+    },
     "./browser": {
       "import": "./dist/svgo.browser.js",
       "types": "./lib/svgo.d.ts"
--- a/package.json
+++ b/package.json
@@ -62,7 +62,8 @@
     "./browser": {
       "import": "./dist/svgo.browser.js",
       "types": "./lib/svgo.d.ts"
-    }
+    },
+    "./package.json": "./package.json"
   },
   "typesVersions": {
     "*": {


package.json
```diff
@@ -59,10 +59,15 @@
       "require": "./dist/svgo-node.cjs",
       "types": "./lib/svgo-node.d.ts"
     },
+    "./dist/svgo.browser.js": {
```
XhmikosR, line 62: If I'm not mistaken, that this will still fail if one requires `dist/svgo.browser`. We might need to tweak it.

SethFalco:
Hmm, we could do something like this:

```json
  "exports": {
    ".": {
      "import": "./lib/svgo-node.js",
      "require": "./dist/svgo-node.cjs",
      "types": "./lib/svgo-node.d.ts"
    },
    "./browser": {
      "import": "./dist/svgo.browser.js",
      "types": "./lib/svgo.d.ts"
    },
    "./dist/svgo.browser": {
      "import": "./dist/svgo.browser.js",
      "types": "./lib/svgo.d.ts"
    },
    "./bin/*": "./bin/*",
    "./dist/*": "./dist/*",
    "./lib/*": "./lib/*",
    "./plugins/*": "./plugins/*",
    "./package.json": "./package.json"
  },
```

We could also do:

```json
  "exports": {
    ".": {
      "import": "./lib/svgo-node.js",
      "require": "./dist/svgo-node.cjs",
      "types": "./lib/svgo-node.d.ts"
    },
    "./browser": {
      "import": "./dist/svgo.browser.js",
      "types": "./lib/svgo.d.ts"
    },
    "./dist/svgo.browser": {
      "import": "./dist/svgo.browser.js",
      "types": "./lib/svgo.d.ts"
    },
    "./*": "./*"
  },
```

However, this will only make the imports available with the `.js`/`.json` suffix, excluding `./dist/svgo.browser` which has an explicit alias defined. I think it's the best we could do for now without explicitly adding aliases for each and every file in the repo.

Alternatively, it could be worth deprecated v3.3.0 and v3.3.1, and making a new release v3.3.2 which reverts the changes. Then we just make a v4.0.0 release, with less concern over breaking changes.

XhmikosR:
Agreed, @SethFalco the latter is the best choice.

But while at it, we should look into any other major changes slated for v4 like dropping any Node.js version support

SethFalco:
Now that we're committed to do a v4 release instead, I'm thinking we could close this, and make the change in imports one of the documented breaking changes.

That way, there's no need to define the aliases and duplicate type versions.

If there are no qualms with that, feel free to close the issue. Otherwise, hit me with any feedback!

XhmikosR:
Sounds good!

Let's document the changes for the release notes

XhmikosR:
Actually, I still think we need package.json there, don't we @SethFalco ?

SethFalco:
> Actually, I still think we need package.json there, don't we @SethFalco ?

As far as I'm concerned, `package.json` isn't considered part of our public interface, so normally I'd say it doesn't belong there.

I noticed in the other issue that it looked like you were importing it, though. Mind sharing your use case and we'll see if it makes sense, or maybe SVGO could do something better to export what is it you need?

For example, if you want the SVGO version at runtime, then this is something we could export ourselves.

XhmikosR:
We use it in svgomg for the version only during build time:

https://github.com/jakearchibald/svgomg/blob/d70f2619fbf6f662d04b6bccd9485558c06b52ba/gulpfile.js#L5

SethFalco:
Understood, instead of export `package.json` then, how about we export a version like how some other libraries do:

```js
import { version } from "react";
import _ from "lodash";
import axios from "axios";

console.log(version, _.VERSION, axios.VERSION);
// 18.3.1 4.17.21 1.6.8
```

I'll write up an issue later, should be simple to do.

---

Thanks for being so active with pull requests and issues, by the way. I'll be reviewing them over the next few days as I find time between other tasks.

PR #2001, opened 2024-05-09
chore: update ESLint to v9 and switch to flat config

XhmikosR:
~~Requires #1987 and probably a rebase after that is merged~~


SethFalco:
I added a commit that makes the following changes:

* Drops the comment of `global ignores` as the section is fairly self-explanatory and the JS Docs explain what the property does.
* Dropped some of the implicit values, namely treating `.js` and `.mjs` as ESM and `.cjs` as CJS. This is just to simplify the config, since the default values are already intuitive.

Meanwhile, LGTM! Thanks for helping with some of the chores!

PR #2002, opened 2024-05-09
chore!: drop Node.js < 16 support

XhmikosR:
@SethFalco not sure if you want to drop support for Node.js 16 too, it doesn't seem needed AFAICT.

BTW Node.js 14 wasn't tested on CI on main.


SethFalco:
> not sure if you want to drop support for Node.js 16 too

I was thining about that recently, and not sure yet myself. Node.js v16 is out of support, but their are still lots of projects/companies using it, including some of my clients (freelance), so I'm leaning towards keeping it, at least until we'd actually benefit doing something that requires v18.

Note: Our test code already requires Node.js v18 I believe, but atm the released build shouldn't.

XhmikosR:
We can always cut a new major version if needed. Right now we don't seem to use any features from newer Node.js so I think we can keep Node.js 16 support.

PR #2004, opened 2024-05-09
chore!: Update commander to v11

XhmikosR:
Commander v11 requires Node.js 16 so this should probably be merged after #2002 and the other PRs.

Please note that this bumps many major versions. I checked commander's release notes, but this could definitely use some more eyes for any breaking changes that might affect svgo.

Ping me when needed for a rebase.


SethFalco:
I've rebased for you, migrated away from the method that was deprecated in commander v8, and tacked on a JS Doc for typings. 

This looks good to me though, most of the breaking changes between major versions didn't affect us at all from what I can tell. Pretty confident this will be fine after looking through the release notes, testing locally, running tests (which includes our command-line interface), etc.

We'll see if anything comes up from users that test the release candidate, though.

PR #2012, opened 2024-05-20
[docs] fix transposition of preset-default

gnowland:
`preset-default` was transposed to `default-preset` in the docs. This PR fixes it.


SethFalco (approved):
Thanks for noticing! I keep making that mistake. 😓

PR #2015, opened 2024-05-26
fix(convertColors): accept commaless rgb

KTibow:
Commaless `rgb(0 0 0)` is now allowed by browsers, and using commas is considered [legacy](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb#legacy_syntax_comma-separated_values). This allows the convertColors plugin to parse and convert these colors, even making SVGs more compatible.


SethFalco (approved):
LGTM! Thanks for submitting a patch.

PR #2016, opened 2024-05-26
feat: export version constant

SethFalco:
Adds a `VERSION` export, similar to React, Axios, and Lodash.

This will enable third-party libraries to check what version of SVGO is in use during runtime, and is intended to supersede the need to import our `package.json` file, which we don't want to make part of our public API.

## Chores

Updates our ESLint config from `2021` to `'latest'`. I wanted to keep it at 2022 or later as we can use top-level awaits.

## Usage

```js
import { VERSION } from 'svgo';

console.log(VERSION);
// 4.0.0
```

```js
const svgo = require('svgo');

console.log(svgo.VERSION);
// 4.0.0
```

## Related

* From discussion in https://github.com/svg/svgo/pull/1999#issuecomment-2102802055

### DIFF
--- a/lib/version.js
+++ b/lib/version.js
@@ -1,0 +1,2 @@
+/** Version of SVGO. */
+export const VERSION = '4.0.0';
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,7 +47,7 @@ jobs:
       - run: yarn install
       - run: yarn playwright install --with-deps chromium
       - run: yarn test
-      - run: yarn test-bundles
+      - run: yarn test:bundles
   regression:
     name: Test regressions
     runs-on: ubuntu-latest
@@ -65,4 +65,4 @@ jobs:
           cache: yarn
       - run: yarn install
       - run: yarn playwright install --with-deps chromium
-      - run: yarn test-regression
+      - run: yarn test:regression
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -20,7 +20,7 @@ export default [
   },
   {
     languageOptions: {
-      ecmaVersion: 2021,
+      ecmaVersion: 'latest',
       globals: {
         ...globals.nodeBuiltin,
       },
--- a/lib/svgo-node.js
+++ b/lib/svgo-node.js
@@ -2,7 +2,7 @@ import os from 'os';
 import fs from 'fs';
 import { pathToFileURL } from 'url';
 import path from 'path';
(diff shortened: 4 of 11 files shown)


lib/version.js
```diff
@@ -0,0 +1,2 @@
+/** Version of SVGO. */
+export const VERSION = '4.0.0';
```
XhmikosR, line 2: Just wondering why don't you just require package.json? It seems simpler...

Also, mind you that `require`s come with a penalty when in CJS; modules is a different case since they are inherently a lot faster when loading.

SethFalco: 1. ESM doesn't have stable support for importing JSON files yet.
2. When using `require`, Rollup doesn't bundle/inline JSON files, which breaks the browser build. (Tries to import `package.json` rather than inlining the JSON at build time, though my initial concern was actually that it would bundle the whole `package.json` instead of just the `version` which I also don't want.)
3. Don't want to take an approach with reading it from `fs` as this would break the browser version where `fs` isn't available.

Most other projects that export a version appear to take a similar approach, which is what I referenced when I did this solution:

* Axios, which uses a [gulpfile](https://github.com/axios/axios/blob/0e4f9fa29077ebee4499facea6be1492b42e8a26/gulpfile.js#L81).
* Moment.js, which uses a [grunt task](https://github.com/moment/moment/blob/18aba135ab927ffe7f868ee09276979bed6993a6/tasks/bump_version.js#L10).
* Handelbars, which also uses a [grunt task](https://github.com/handlebars-lang/handlebars.js/blob/25c696b8891e65f7f7e8405ecdf0a6b17808d237/tasks/version.js#L25).

I've only seen 2 projects so far take your approach with `require`, but neither is applicable in our case:
* Mongoose, but it only exports the version in the [Node.js exports](https://github.com/Automattic/mongoose/blob/master/index.js#L30), and isn't available in the [browser bundle](https://github.com/Automattic/mongoose/blob/master/lib/browser.js).
* Winston, which [requires](https://github.com/winstonjs/winston/blob/debf4fad6a4ee4121904781dc83f8723be604fbe/lib/winston.js#L17), but [lacks browser support](https://github.com/winstonjs/winston/issues/2196) so supporting both isn't a concern for them.

Feature parity between Node.js and browser is important for SVGO, and this seemed like the best approach for that.

If you have a better idea, you're welcome to pitch it though! I was waiting before merging, specifically to see if others would propose a better approach.

PR #2026, opened 2024-06-08
chore(ci): add Node.js 22

XhmikosR:
The matrix is getting a little big but better test everything. We could probably just test Node.js 16 on Ubuntu to save some time, or exclude some things. Let me know if you have any suggestions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,6 +31,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
+          - 22
           - 20
           - 18
           - 16


SethFalco:
> We could probably just test Node.js 16 on Ubuntu to save some time

Regarding time specifically, it's not an immediate concern since they run in parallel, and we have unlimited built minutes. Though, I do still value the idea to reduce running unnecessary tasks. (Wasted energy.)

When I added Windows to CI in https://github.com/svg/svgo/pull/1853, I was thinking it might be best to only run the Windows task on the latest LTS version of Node.js (so 20 atm). I opted against it to play it safe, but thinking more about it, I don't really see the value.

In practice, I don't really expect to encounter issues that differ in both Node.js version and operating system, though it probably can happen. We definitely want to test each Node.js version, and we want to test each operating system for platform specific behavior, but I don't think we need every combination of the two. 🤔

Rather than drop Windows on Node.js v16, it might be worth only testing Windows on Node.js v20. (Node.js v22 isn't officially LTS until October 2024.)

SethFalco:
Regarding the speed, that's because I recently merged the PR that uses [svg/svgo-test-suite](https://github.com/svg/svgo-test-suite).

TL;DR: We no longer only test for regressions on 60~ or so W3C test files, but also another 4,000+ icons from Oxygen Icons (KDE icon set). I was hoping to improve performance more before merging, but I decided to get it in already to improve QA before making the SVGO v4 release.

Reference: 

* https://github.com/svg/svgo/pull/1918
* https://github.com/KDE/oxygen-icons

It's a nice icon set to test against, since it has a lot of more complex shapes. But unfortunately, takes the regression test task from 1-2 minutes to 18-20 minutes, but seems worth it to me.

Some examples (these are the PNG versions, though):





---

Sure though, we can leave it. Later, I may still evaluate disabling some Windows builds, but we'll figure it out later.

PR #2032, opened 2024-06-11
feat: convertColors plugin ignore currentColor in 'mask' elements

### LINKED ISSUE #2031: Mask colors are changed when using convertColors and currentColor: true
**Describe the bug**
when using the `convertColors` plugin with the parameter `currentColor` set to `true`, if the svg contains `mask` elements the color attributes (fill/stroke) are changed, this behaivour is unwanted in my case because it breaks the masks depending on the current color.

**To Reproduce**

test.svg:
```
<svg viewBox="-10 -10 120 120">
  <svg width="100" height="100" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
    <mask id="path-1-inside">
      <path fill-rule="evenodd" clip-rule="evenodd" fill="white"
        d="M9.14286 2H4.85714C4.38376 2.00001 4.00001 2.30381 4 2.67857V20.3214C4.00001 20.6962 4.38376 21 4.85714 21H9.14286C9.61624 21 9.99999 20.6962 10 20.3214V2.67857C9.99999 2.30381 9.61624 2.00001 9.14286 2Z" />
    </mask>
    <path
      d="M9.14286 2L9.1429 0H9.14286V2ZM4.85714 2V0H4.8571L4.85714 2ZM4 2.67857L2 2.6785V2.67857H4ZM4 20.3214H2V20.3215L4 20.3214ZM4.85714 21L4.8571 23H4.85714V21ZM9.14286 21V23H9.1429L9.14286 21ZM10 20.3214L12 20.3215V20.3214H10ZM10 2.67857H12V2.6785L10 2.67857ZM9.14286 0H4.85714V4H9.14286V0ZM4.8571 0C3.73997 2.65837e-05 2.00007 0.78891 2 2.6785L6 2.67865C5.99998 3.21624 5.71952 3.5871 5.49246 3.76685C5.26878 3.94393 5.03752 4 4.85719 4L4.8571 0ZM2 2.67857V20.3214H6V2.67857H2ZM2 20.3215C2.00007 22.2111 3.73997 23 4.8571 23L4.85719 19C5.03752 19 5.26878 19.0561 5.49246 19.2332C5.71952 19.4129 5.99998 19.7838 6 20.3214L2 20.3215ZM4.85714 23H9.14286V19H4.85714V23ZM9.1429 23C10.26 23 11.9999 22.2111 12 20.3215L8 20.3214C8.00002 19.7838 8.28048 19.4129 8.50754 19.2332C8.73122 19.0561 8.96248 19 9.14281 19L9.1429 23ZM12 20.3214V2.67857H8V20.3214H12ZM12 2.6785C11.9999 0.78891 10.26 2.65837e-05 9.1429 0L9.14281 4C8.96248 4 8.73122 3.94393 8.50754 3.76685C8.28048 3.5871 8.00002 3.21624 8 2.67865L12 2.6785Z"
      fill="red" mask="url(#path-1-inside)" />
  </svg>
</svg>
```

config.js:
```
module.exports = {
  plugins: [
    {
      name: 'convertColors',
      params: {
        currentColor: true
      }
    }
  ]
}
```

run:
```
npx svgo test.svg -o out.svg --config config.js
```

**Expected behavior**
in out.svg, `path` inside `mask` shouldn't have `fill` set to `currentColor`, but should keep `white`

- SVGO Version: 3.3.2
- NodeJs Version: 20.5.1
- OS: Ubuntu

salvan13:
`convertColors` plugin ignore `currentColor` in `mask` elements since it makes them not working properly

fixes #2031

### DIFF
--- a/plugins/convertColors.js
+++ b/plugins/convertColors.js
@@ -44,10 +44,9 @@ const convertRgbToHex = ([r, g, b]) => {
 };
 
 /**
- * @type {(node: XastNode) => boolean}
+ * @type {(node: XastElement) => boolean}
  */
-const isMask = (node) =>
-  node && node.type === 'element' && node.name === 'mask';
+const isMask = (node) => node.name === 'mask'
 
 /**
  * Convert different colors formats in element attributes to hex.
--- a/plugins/convertColors.js
+++ b/plugins/convertColors.js
@@ -1,5 +1,10 @@
 import { colorsNames, colorsProps, colorsShortNames } from './_collections.js';
 import { includesUrlReference } from '../lib/svgo/tools.js';
+import { visitSkip } from '../lib/xast.js';
+
+/**
+ * @typedef {import('../lib/types.js').XastNode} XastNode
+ */
 
 export const name = 'convertColors';
 export const description =
@@ -40,6 +45,12 @@ const convertRgbToHex = ([r, g, b]) => {
 };
 
 /**
+ * @type {(ignoredNodes: string[], node: XastNode) => boolean}
+ */
+const isIgnored = (ignoredElements, node) =>
+  node && node.type === 'element' && ignoredElements.includes(node.name);
+
+/**
  * Convert different colors formats in element attributes to hex.
  *
  * @see https://www.w3.org/TR/SVG11/types.html#DataTypeColor
@@ -71,11 +82,15 @@ export const fn = (_root, params) => {
     convertCase = 'lower',
     shorthex = true,
     shortname = true,
+    ignoreElements = [],
   } = params;
 
   return {
     element: {
       enter: (node) => {
+        if (currentColor && isIgnored(ignoreElements, node)) {
+          return visitSkip;
+        }
         for (const [name, value] of Object.entries(node.attributes)) {
           if (colorsProps.has(name)) {
             let val = value;
--- a/plugins/convertColors.js
+++ b/plugins/convertColors.js
@@ -48,7 +48,7 @@ const convertRgbToHex = ([r, g, b]) => {
  * @type {(ignoredNodes: string[], node: XastNode) => boolean}
  */
 const isIgnored = (ignoredElements, node) =>
-  node && node.type === 'element' && ignoredElements.includes(node.name);
+  node && node.type === 'element' && ignoredElements && ignoredElements.includes(node.name);
 
 /**
  * Convert different colors formats in element attributes to hex.
@@ -88,7 +88,7 @@ export const fn = (_root, params) => {
   return {
     element: {
       enter: (node) => {
-        if (currentColor && isIgnored(ignoreElements, node)) {
+        if (isIgnored(ignoreElements, node)) {
           return visitSkip;
         }
         for (const [name, value] of Object.entries(node.attributes)) {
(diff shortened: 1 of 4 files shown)


SethFalco:
Rather than adding the param, imo the plugin should just internally ignore masks for the `convertColor` option.

This was a use-case that was overlooked when the option was added, but I don't see a use-case to convert a mask to `currentColor` at all. (If that comes up in future, we can amend the `convertColor` option or introduce a new one to include masks.)

An issue with this solution is that it calls `visitSkip`, which stops all optimization. This would also stop it from applying any of the other options too, like converting `white` to `#fff`, which is valid for masks and shaves 1 byte.

A better way to approach it would be to have a counter at the top that increments when an open `mask` tag is hit, and decrements when a close `mask` tag is hit, then only apply `convertColor` option when that counter is 0. (We're not inside a mask.)

How does that sound to you? If you have any feedback or want guidance, feel free to ask!

salvan13:
Hello! Thanks for your reply, it sounds good to me!

I also think that there is no use case for `mask` to convert colors to `currentColor`, I'll try to implement it as you suggested without a config and with a counter to not apply `currentColor` to `mask`, I'll update the PR asap!

salvan13:
Hello @SethFalco I updated the PR, now it seems working great :smiley: 
the mask colors are optimized but not changed to `currentColor`.
Thanks for your review!

plugins/convertColors.js
```diff
@@ -39,6 +43,12 @@ const convertRgbToHex = ([r, g, b]) => {
   return '#' + hexNumber.toString(16).slice(1).toUpperCase();
 };
 
+/**
+ * @type {(node: XastNode) => boolean}
+ */
+const isMask = (node) =>
+  node && node.type === 'element' && node.name === 'mask';
```
SethFalco, line 50:
```suggestion
 * @type {(node: XastElement) => boolean}
 */
const isMask = (node) => node.name === 'mask'
```

(Also requires changing the `@typedef` at the top if this is used!)

In both places that you call `#isMask`, you actually have a `XastElement`, a more concrete type of `XastNode` that allows you to safely make some assumptions.

In both cases, you can safely check `node.name` without any defensive checks.

(Though it might be better to just inline the condition now that it's so short.)

SethFalco (commented):
Looks good to me! I've raised a minor issues with the types/defensive checks, but the logic is sound.

plugins/convertColors.js
```diff
@@ -1,6 +1,10 @@
 import { colorsNames, colorsProps, colorsShortNames } from './_collections.js';
 import { includesUrlReference } from '../lib/svgo/tools.js';
 
+/**
+ * @typedef {import('../lib/types.js').XastNode} XastNode
```
SethFalco, line 5:
```suggestion
 * @typedef {import('../lib/types.js').XastElement} XastElement
```

salvan13: I have inlined the condition :)

SethFalco: Ahh, scrap this! Didn't realize you were still working on it.

salvan13: it's fine :) let me know if it's ok now!

SethFalco:
Awesome, thank you very much for the contribution!

This will be releases in v4.0.0-rc.1, and eventually v4.0.0.

Issue #1968, opened 2024-03-05
cleanupAttrs in 'preset-default' overrides is not respected


sriramrudraraju:
**Describe the bug**
 `cleanupAttrs` to ignore spacing from svgs

The below config isnt working

```javascript
export default {
  plugins: [
    {
      name: 'preset-default',
      params: {
        overrides: {
          // disable a default plugin
          removeViewBox: false,
          cleanupAttrs: false, // added this to base config

          // customize the params of a default plugin
          inlineStyles: {
            onlyMatchedOnce: false,
          },
         // tried the below config too
         //  also tried switching cleanupAttrs: false and this
         cleanupAttrs: {
            spaces: false
          }
        },
      },
    },
  ],
};
```

**To Reproduce**

I tried the  below code in the svgo playground as well

```javascript
const svg = `
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" viewBox=" 0 0      150 100 " width="150">
  <!-- Created with love! -->
  <defs>
    <ellipse cx="50" cy="50.0" rx="50.00" ry="auto" fill="black" id="circle"/>
  </defs>
  <g>
    <use href="#circle" transform="skewX(16)"/>
    <rect id="useless" width="0" height="0" fill="#ff0000"/>
  </g>
</svg>
`;

const svgoConfig = {
  js2svg: { indent: 2, pretty: true },
  plugins: [
    {
      name: 'preset-default',
      params: {
        overrides: {
          // disable a default plugin
          removeViewBox: false,

          // customize the params of a default plugin
          inlineStyles: {
            onlyMatchedOnce: false,
          },
          cleanupAttrs: {
            spaces: false
          }
        },
      },
    }
  ]
}

render(<SvgoPreview svg={svg} svgoConfig={svgoConfig}/>);
```

```javascript
// rendered output
<svg xmlns="http://www.w3.org/2000/svg" width="150" viewBox="0 0 150 100">
  <defs>
    <circle id="a" cx="50" cy="50" r="50" fill="#000"/>
  </defs>
  <use href="#a" transform="skewX(16)"/>
</svg>
```

```javascript
// desired output (spacing in viewbox)
<svg xmlns="http://www.w3.org/2000/svg" width="150" viewBox=" 0 0      150 100 ">
  <defs>
    <circle id="a" cx="50" cy="50" r="50" fill="#000"/>
  </defs>
  <use href="#a" transform="skewX(16)"/>
</svg>
[...]

PR #2036, opened 2024-06-14
fix: don't insert 0 at start or end of attribute if whitespace

SethFalco:
### Docs

Updates the documentation to clarify explicitly that the plugin impacts whitespace. Internally, it's because we parse all numbers in the attribute into an array, then rewrite it back.

### Bug Fix

I also fix a bug that I discovered while testing the issue reported. When an attribute like `"  0   0  150 100  "` comes up, SVGO inserts a `0` to the start and end of the attribute, completely changing the meaning.

This was because when we do the split, the array we get back starts with an empty string and ends with an empty string, which were then converted to `0`, resulting in `"0 0 0 150 100 0"` instead of `"0 0 150 100"`.

```js
seth@seth-pc-tux:~$ node
Welcome to Node.js v20.11.0.
Type ".help" for more information.
> Number("")
0
```

Now we filter them out before parsing.

### Performance

Just applying a recommendation from SonarLint, use `RegExp#exec` instead of `String#match`. 

### Related

* Closes https://github.com/svg/svgo/issues/1968

### DIFF
--- a/plugins/cleanupNumericValues.js
+++ b/plugins/cleanupNumericValues.js
@@ -39,6 +39,7 @@ export const fn = (_root, params) => {
         if (node.attributes.viewBox != null) {
           const nums = node.attributes.viewBox.split(/\s,?\s*|,\s*/g);
           node.attributes.viewBox = nums
+            .filter((value) => value.length != 0)
             .map((value) => {
               const num = Number(value);
               return Number.isNaN(num)
@@ -54,7 +55,7 @@ export const fn = (_root, params) => {
             continue;
           }
 
-          const match = value.match(regNumericValues);
+          const match = regNumericValues.exec(value);
 
           // if attribute value matches regNumericValues
           if (match) {
--- a/docs/03-plugins/cleanupNumericValues.mdx
+++ b/docs/03-plugins/cleanupNumericValues.mdx
@@ -18,4 +18,4 @@ svgo:
       default: true
 ---
 
-Rounds numeric values, and removes the unit when it's `px` as this is the default.
+Rounds numeric values, removes the unit when it's `px` as this is the default, and removes redundant spaces around and between numbers.
--- a/test/plugins/cleanupNumericValues.03.svg.txt
+++ b/test/plugins/cleanupNumericValues.03.svg.txt
@@ -1,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox=" 0 0      150 100 "/>
+
+@@@
+
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 150 100"/>


plugins/cleanupNumericValues.js
```diff
@@ -39,6 +39,7 @@ export const fn = (_root, params) => {
         if (node.attributes.viewBox != null) {
           const nums = node.attributes.viewBox.split(/\s,?\s*|,\s*/g);
           node.attributes.viewBox = nums
+            .filter((value) => value.length != 0)
```
GreLI, line 42: It would be easier and more performant to amend regex to something like `/\s+,?\s*|,\s*/g`

SethFalco: Hmm, I don't see how that'd work.

For clarity, with your proposed regex, the split would be:

* `'0 0 100 100'` ⇒ `['0', ' ', '0', ' ', '100', ' ', '100']`
* `'0 0 100 100'` ⇒ `[' 0', ' ', '0', ' ', '100', ' ', '100 ']`
* `'0 0 100 100'` ⇒ `['  0', '  ', '0', '  ', '100', '  ', '100  ']`

I don't think this would be easy with regex alone, as it'd require a negative look behind with a non-fixed width, which isn't possible.

For example, to exclude the spaces at the end, `/(?:\s,?\s*|,\s*)(?!\s*$)/g` would work. But the same strategy can't be employed to exclude the spaces at the start.

Meanwhile, even if we were to ignore the different results, using either of the proposed regexes performs slower than a simple regex + iteration.

Something that actually would be both sensible and faster would be to `trim` the attribute instead of iterating after splitting, though.

i.e. `node.attributes.viewBox.trim().split(/\s,?\s*|,\s*/g)`

```js
const Benchmark = require('benchmark');
const suite = new Benchmark.Suite;

const data = [
  '0 0 100 100',
  ' 0 0 100 100 ',
  '  0  0  100  100  ',
]

/**
 * @param {string} viewbox
 * @return {string[]}
 */
const splitV1 = (viewbox) => {
  return viewbox.split(/\s,?\s*|,\s*/g).filter(v => v.length != 0);
};

/**
 * @param {string} viewbox
 * @return {string[]}
 */
const splitV2 = (viewbox) => {
  return viewbox.trim().split(/\s,?\s*|,\s*/g);
};

suite
  .add('v1', () => data.map((d) => splitV1(d)))
  .add('v2', () => data.map((d) => splitV2(d)))
  .on('cycle', (event) => console.log(String(event.target)))
  .run({ 'async': true });

// v1 x 2,648,671 ops/sec ±0.87% (91 runs sampled)
// v2 x 2,800,976 ops/sec ±0.53% (93 runs sampled)
```

GreLI: Number is ok to extra spaces, so all we need is to split numbers. The key factor here is to get rid of empty strings at the start and the end. So, to make it work in the middle, I've added `\b` separator (lookbehind and lookahead also would work, but `\b` is simpler). I forgot, that `split()` captures group, so non-capturing group is needed. So, the regex would be like: `/\b(?:\s+,?\s*|\s*,\s*)\b/g`.

Also, `matchAll()` can be used just to get numbers, but it will not be spec compliant — any separators would work.

SethFalco: Ahh, yeah, that makes sense. And indeed, your proposal works now!

As we have existing test cases that support `-`, we should add that too.  
So `/\b(?:\s+,?\s*|\s*,\s*)(?=[-.\d])/g`, which includes a `-` in the look ahead.

Alternatively, we could also just use `[^\s]` to exclude spaces rather than a list of items to include, which simplifies the regex a little while offering more safety in case we missed something.  
So `/\b(?:\s+,?|\s*,)\s*(?=[^\s])/g`, this also moves the duplicate matcher for `\s*` to outside of the non-capture group.

```js
const data = [
  '0 0 100 100',
  ' 0 0 100 100 ',
  '  0  0  100  100  ',
  ' 0  0  0.5  .5 ',
  '20.000001 -19.99999 17.123456 70.708090',
];

const splitV1 = (viewbox) => {
  return viewbox.split(/\s,?\s*|,\s*/g).filter(v => v.length != 0);
};
// v1 x 1,591,214 ops/sec ±1.20% (92 runs sampled)

const splitV2 = (viewbox) => {
  return viewbox.trim().split(/\s,?\s*|,\s*/g);
};
// v2 x 1,585,072 ops/sec ±1.14% (92 runs sampled)

const splitV3 = (viewbox) => {
  return viewbox.split(/\b(?:\s+,?\s*|\s*,\s*)(?=[-.\d])/g);
};
// v3 x 1,684,750 ops/sec ±0.76% (94 runs sampled)

const splitV4 = (viewbox) => {
  return viewbox.split(/\b(?:\s+,?\s*|\s*,\s*)(?=[^\s])/g);
};
// v4 x 1,698,961 ops/sec ±0.68% (94 runs sampled)

const splitV5 = (viewbox) => {
  return viewbox.split(/\b(?:\s+,?|\s*,)\s*(?=[^\s])/g);
};
// v5 x 1,736,749 ops/sec ±0.43% (93 runs sampled)
```

GreLI:
`[^\s]` is just `\S`

plugins/cleanupNumericValues.js
GreLI: `[^\s]` is just `\S`

SethFalco: I'd opened a pull request to apply the regex change, but closed it.

I've been getting more into benchmarking, and decided to finally revamp how I benchmark locally, which includes defining more rules and utilities.

One rule, is that functions must have identical input/output. Not "effectively identical", which is what I did before. This enforces a more complete benchmark, rather than only testing intermediate steps in an algorithm, where the performance of later steps may be impacted by the results of earlier steps.

For example, I treated `"100"` and `"  100"` as the same, because we know that both become `100` when called on `Number()`. However, this then doesn't account for the performance of `Number()` itself with the different inputs.

Writing a more complete benchmark has a notable impact on results:

```js
const data = [
  '0 0 120 120',
  '0 0 230 120',
  '110 0 120 120',
  '0 0 200.28423 200.28423',
  '20.000001 -19.99999 17.123456 70.708090',
  ' 0 0      150 100 ',
  '  0  0  0.5  .5  ',
];

function splitV2(viewbox) {
  return viewbox
    .trim()
    .split(/(?:\s,?|,)\s*/g)
    .map(Number);
};
// v2 x 619,413 ops/sec ±0.42% (97 runs sampled)

function splitV5(viewbox) {
  return viewbox
    .split(/\b(?:\s+,?|\s*,)\s*(?=\S)/g)
    .map(Number);
};
// v5 x 599,606 ops/sec ±0.42% (94 runs sampled)
```

The difference is minor (4-5%), but I'd argue using `trim` is simpler to maintain over a more complicated regex (I say this as someone that loves regex.) while offering on-par or better performance over the inputs we'd realistically encounter.

GreLI: But now we have `.filter()` instead of `.trim()`, which probably is worse.

SethFalco: Yeah, `.filter()` is worse! I still plan to do a PR for this to use `trim` and to move the `\s*` outside the group. I was just dropping an update already with the findings.

Before opening another PR, just wanted to finish evaluating how I benchmark locally so I can handle this better in future, and review other decisions I'd make historically based on benchmarks.

PR #2053, opened 2024-08-05
Removed call to encodeSVGDatauri() from coa.js and added test case.

### LINKED ISSUE #2048: --datauri 'unenc' add 'data:image/svg+xml,data:image/svg+xml,'
**Describe the bug**

`--datauri 'unenc'` add to ouput SVG `"data:image/svg+xml,data:image/svg+xml,"`

- SVGO Version: 3.3.2
- NodeJs Version: 18, 22
- OS: Windows

[reproduce on stackblitz](https://stackblitz.com/edit/stackblitz-starters-enayzf?file=package.json)

johnkenny54:
encodeSVGDatauri() is called within the optimize() function, but was being called a second time in coa.js when SVGO was run from the command line, giving a doubly-encoded, incorrect result.

This PR removes the call from coa.js and adds a test case.

Resolves #2048.

### DIFF
--- a/lib/svgo/coa.js
+++ b/lib/svgo/coa.js
@@ -2,7 +2,7 @@ import fs from 'fs';
 import path from 'path';
 import colors from 'picocolors';
 import { fileURLToPath } from 'url';
-import { encodeSVGDatauri, decodeSVGDatauri } from './tools.js';
+import { decodeSVGDatauri } from './tools.js';
 import { loadConfig, optimize } from '../svgo-node.js';
 import { builtin } from '../builtin.js';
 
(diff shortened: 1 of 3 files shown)


GreLI:
Actually, for current web browsers the only symbol needs to be escaped is “#”. Maybe it worth optimize data uri encoding…

johnkenny54:
Optimizing encodeSVGDatauri() would be a separate issue. This PR only addresses removing the duplicate call, which, as described in the related issue, completely breaks the feature from the command line.

GreLI:
True. Just noted here.

PR #2054, opened 2024-08-06
Ignore malformed style attributes in minifyStyles.

### LINKED ISSUE #1863: TypeError: Cannot read properties of undefined (reading '0')
Attached file issue41.svg gives this error:
```
TypeError: Cannot read properties of undefined (reading '0')
    at Object.getPropertyDescriptor [as property] (C:\Users\<username>\AppData\Roaming\npm\node_modules\svgo\node_modules\csso\node_modules\css-tree\cjs\utils\names.cjs:69:24)
```

with this command
`svgo issue41.svg
`
and also when opened here https://jakearchibald.github.io/svgomg/

image

issue41

(File is from the canvg test suite https://github.com/canvg/canvg/blob/master/test/svgs/issue41.svg )

johnkenny54:
If a style attribute contains malformed styles, csso.minifyBlock() may throw an exception. This PR changes the code to catch the exception and leave the style value unchanged.

Resolves #1863.

### DIFF
--- a/plugins/minifyStyles.js
+++ b/plugins/minifyStyles.js
@@ -129,9 +129,13 @@ export const fn = (_root, { usage, ...params }) => {
         for (const node of elementsWithStyleAttributes) {
           // style attribute
           const elemStyle = node.attributes.style;
-          node.attributes.style = csso.minifyBlock(elemStyle, {
-            ...params,
-          }).css;
+          try {
+            node.attributes.style = csso.minifyBlock(elemStyle, {
+              ...params,
+            }).css;
+          } catch {
+            /** ignore style syntax errors */
+          }
         }
       },
     },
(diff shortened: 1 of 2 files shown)


GreLI:
Maybe add an option to remove malformed style? Users may want to have garbage styles removed. But it's not clear if the styles wrong or just some new feature not recognized by SVGO, so it should be optional.

johnkenny54:
I've pretty much given up on any bugs ever getting fixed in SVGO. I've been working on a spinoff of SVGO 4 which prioritizes lossless compression - https://www.npmjs.com/package/svgo-ll - that version handles the file in issue #1863.

svgo-ll discards the malformed property name since it's not valid. If you edit the input file to correct the property name, the property is retained, but it doesn't affect how the file is rendered.

PR #2064, opened 2024-08-20
Preserve all white space within foreignObject elements.

### LINKED ISSUE #1473: Incorrect parsing of html tags within `foreignObject` tags
**Describe the bug**
Empty html tags are transformed into self-closing tags which breaks expected image.
Required whitespaces from text nodes are removed.

**To Reproduce**

```js
import SVGO from "svgo"

const svg = `
  <foreignObject x="0" y="0" width="100%" height="100%">
    <div xmlns="http://www.w3.org/1999/xhtml">
      Lorem <span>ipsum</span> dolor
    </div>
    <div class="empty-div"></div>
  </foreignObject>
</svg>`

const {data} = await SVGO.optimize(svg, {multipass:true})
console.log(data)
```

**Expected behavior**

```diff
+<svg xmlns="http://www.w3.org/2000/svg"><foreignObject x="0" y="0" width="100%" height="100%"><div xmlns="http://www.w3.org/1999/xhtml">Lorem <span>ipsum</span> dolor</div><div class="empty-div"></div></foreignObject></svg>
-<svg xmlns="http://www.w3.org/2000/svg"><foreignObject x="0" y="0" width="100%" height="100%"><div xmlns="http://www.w3.org/1999/xhtml">Lorem<span>ipsum</span>dolor</div><div class="empty-div"/></foreignObject></svg>
```
- Empty `<div>` tag should not be transformed into self closing tags (breaks html interpretation by creating "unclosed" tags)
- Spaces within html text nodes are not preserved (note that `xml:space="preserve"` or using `&nbsp;` change nothing)

 - SVGO Version: 2.3.0
 - NodeJs Version: 14.9.0
 - OS: Windows
 
**Additional context**
Relates to #1421, but add more insight about troubleshooting and reproduction
Updated https://github.com/lowlighter/metrics/issues/172#issue-822378888 to display original and optimized image

### LINKED ISSUE #1678: Non-breaking spaces in <foreignObject> get trimmed
**Describe the bug**
When parsing HTML in `<foreignObject>`, non-breaking spaces (`&nbsp;` aka `&#xA0;` aka `&#160;`) are getting removed.

**To Reproduce**
```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 400 300">
  <foreignObject width="400" height="300">
    <div xmlns="http://www.w3.org/1999/xhtml" style="width: 100%; height: 100%; background: blue; color: white; display: flex; justify-content: center; align-items: center;">
      <div>foo </div>
      <div>bar&#xA0;</div>
      <div>baz</div>
    </div>
  </foreignObject>
</svg>
```

Even with `--pretty`, this becomes `foobarbaz`.

**Expected behavior**
`foobar baz`

 - SVGO 2.8.0
 - Node.js 14.9.3
 
**Additional context**
I've narrowed it down to:

https://github.com/svg/svgo/blob/a9834efa1603198ae606765292215462ae193d0d/lib/parser.js#L228

I didn't check in detail, but presumably, `String#trim()` will remove any kind of Unicode whitespace, and there's quite a lot of variety there. If a space is non-breaking, I don't think SVGO should touch it.

Suggested fixes:
- Replace `String#trim()` with something that only trims `[ \t\r\n]`.
- Add HTML elements to `textElems` so the `else` doesn't get hit.
- Don't trim anything inside `<foreignObject>` to avoid implementing an entire HTML parser.

### LINKED ISSUE #1728: SVGO creates invalid self-closing tags inside `<foreignObject>` elements
**Describe the bug**
When an empty tag is present inside of a `<foreignObject>` element, SVGO converts it to a self-closing tag. This results in tags that are invalid HTML and are misinterpreted by the browser.

**To Reproduce**
1. Run `svgo test.svg --pretty --indent=2`, where
   `test.svg`:
``` svg
<svg xmlns="http://www.w3.org/2000/svg" width="100px" height="100px" viewBox="0 0 10 10">
  <foreignObject x="0" y="0" width="10" height="10" style="font-size:2px">
    <p style="margin-top:0">
      <span style="height:4px;display:inline-block"></span>
      Some text.
    </p>
  </foreignObject>
</svg>
```

2. Observe the result, which is:
``` svg
<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100" viewBox="0 0 10 10">
  <foreignObject width="10" height="10" x="0" y="0" style="font-size:2px">
    <p style="margin-top:0">
      <span style="height:4px;display:inline-block"/>
      Some text.
    </p>
  </foreignObject>
</svg>
```
Note that the empty `<span>` element has been converted to a self-closing tag. This results in invalid HTML, since `<span>` is not a [void element](https://html.spec.whatwg.org/multipage/syntax.html#void-elements). The browser actually interprets the self-closing tag as an opening tag, and then automatically closes the tag later on. This is essentially equivalent to:
``` svg
<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100" viewBox="0 0 10 10">
  <foreignObject width="10" height="10" x="0" y="0" style="font-size:2px">
    <p style="margin-top:0">
      <span style="height:4px;display:inline-block">
      Some text.
      </span>
    </p>
  </foreignObject>
</svg>
```
(See [here](https://stackoverflow.com/questions/3558119/are-non-void-self-closing-tags-valid-in-html5) for more information.)
When viewing the SVG in a browser, **keep in mind that `<foreignObject>` elements will not be rendered unless the SVG is inside of an HTML page**. To fix this, **add `<html>` tags and save the file as an HTML file.**

**Expected behavior**
Inside `<foreignObject>` elements, empty elements should not be converted to self-closing tags.

**Version info:**
 - SVGO Version: 3.0.2
 - NodeJs Version: v19.2.0
 - OS: Linux

johnkenny54:
This PR modifies SVGO to always preserve white space within `<foreignObject>` elements. In general there is no way to know the semantics of the `<foreignObject>` content, so all white space within the element is preserved.

The following changes were made:

- **parser.js** - keeps track of whether the parser is within a `<foreignObject>` element, and if so does not trim text or comment nodes.
- **stringifier.js** - keeps track of whether the stringifier is within a `<foreignObject>` element, and if so does not emit prettification indentation or newlines.
- **_collections.js** - removed 'pre' from `textElems` set - this was added by PR #1796, but is not necessary since all `<foreignObject>` content now preserves white space.`.
- Added a test case to confirm that CDATA and comments are preserved within a `<foreignObject>`.
- Modified expected results in several test files. In all cases, the expected result is now to preserve the contents of the `<foreignObject>` element exactly.

Resolves #1728, resolves #1678, resolves #1473.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -102,6 +102,7 @@ export const parseSvg = (data, from) => {
    * @type {XastParent[]}
    */
   const stack = [root];
+  let foreignLevel = 0;
 
   /**
    * @type {(node: XastChild) => void}
@@ -157,7 +158,7 @@ export const parseSvg = (data, from) => {
      */
     const node = {
       type: 'comment',
-      value: comment.trim(),
+      value: foreignLevel > 0 ? comment : comment.trim(),
     };
     pushToContent(node);
   };
@@ -189,12 +190,15 @@ export const parseSvg = (data, from) => {
     pushToContent(element);
     current = element;
     stack.push(element);
+    if (data.name === 'foreignObject') {
+      foreignLevel++;
+    }
   };
 
   sax.ontext = (text) => {
     if (current.type === 'element') {
       // prevent trimming of meaningful whitespace inside textual tags
-      if (textElems.has(current.name)) {
+      if (foreignLevel > 0 || textElems.has(current.name)) {
         /**
          * @type {XastText}
          */
@@ -216,9 +220,12 @@ export const parseSvg = (data, from) => {
     }
   };
 
-  sax.onclosetag = () => {
+  sax.onclosetag = (tagName) => {
     stack.pop();
     current = stack[stack.length - 1];
+    if (tagName === 'foreignObject') {
+      foreignLevel--;
+    }
   };
 
   sax.onerror = (e) => {
--- a/lib/stringifier.js
+++ b/lib/stringifier.js
@@ -14,6 +14,8 @@ import { textElems } from '../plugins/_collections.js';
  *   indent: string,
  *   textContext: ?XastElement,
  *   indentLevel: number,
+ *   foreignLevel: number,
+ *   eolLen: number
  * }} State
  * @typedef {Required<StringifyOptions>} Options
  */
@@ -81,14 +83,6 @@ export const stringifySvg = (data, userOptions = {}) => {
   } else if (typeof indent === 'string') {
     newIndent = indent;
   }
-  /**
-   * @type {State}
-   */
-  const state = {
-    indent: newIndent,
-    textContext: null,
-    indentLevel: 0,
-  };
   const eol = config.eol === 'crlf' ? '\r\n' : '\n';
   if (config.pretty) {
     config.doctypeEnd += eol;
@@ -100,6 +94,16 @@ export const stringifySvg = (data, userOptions = {}) => {
     config.tagCloseEnd += eol;
     config.textEnd += eol;
   }
+  /**
+   * @type {State}
+   */
+  const state = {
+    indent: newIndent,
+    textContext: null,
+    indentLevel: 0,
+    foreignLevel: 0,
+    eolLen: eol.length,
+  };
   let svg = stringifyNode(data, config, state);
   if (config.finalNewline && svg.length > 0 && !svg.endsWith('\n')) {
     svg += eol;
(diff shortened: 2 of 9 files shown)


rnwst:
Thanks for this PR. If all it does is to preserve white space within `foreignObject` elements, then it does **not** address #1728. #1728 is about invalid self-closing tags, not preserving whitespace.

PR #2066, opened 2024-08-22
inlineStyles plugin should not remove id attribute from target element

### LINKED ISSUE #2065: inlineStyles plugin should not remove id attribute from target element
When the inlineStyles plugin inlines a style with an id selector, it is removing the id attribute from the target element. It shouldn't do this, as it has no way of knowing whether the id is used elsewhere.

**To Reproduce**
Optimize the following with the inlineStyles plugin:

```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
  <style>
    #a {stroke:red;}
  </style>
  <path id="a" d="M 2 2 h10"/>
  <use href="#a" x="5" y="5"/>
</svg>
``` 

**Expected behavior**
The style should be inlined, but the id attribute should not be deleted.

Version 4.0.0-rc.1

johnkenny54:
Removed code that deletes id attribute from elements.

Resolves #2065.

### DIFF
--- a/plugins/inlineStyles.js
+++ b/plugins/inlineStyles.js
@@ -293,7 +293,7 @@ export const fn = (root, params) => {
           return;
         }
 
-        // clean up matched class + ID attribute values
+        // clean up matched class values
         for (const selector of sortedSelectors) {
           if (selector.matchedElements == null) {
             continue;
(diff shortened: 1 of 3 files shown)


strarsis:
Wouldn't an option be better? Subsequent plugins do not have the context to clean up the `ID` attribute, as it is removed by inline styles as being then unnecessary, as its associated styles were directly applied to the element.

johnkenny54:
Having an option isn't safe and just adds complexity to the plugin. If the id is no longer referenced, it should be removed by **cleanupIds**. Currently **cleanupIds** disables itself in the presence of `<style>` elements, which is not ideal, but at least preferable to what many other plugins do, since it doesn't break anything. But there's no reason **cleanupIds** couldn't check the selectors to see which ids are referenced by styles.

PR #2076, opened 2024-09-25
chore: export the `Output` type

### LINKED ISSUE #2074: Export the `Output` type to be directly importable
Hi! 👋 

The `Output` type (the return type of the `optimize()` function) is not currently exported. As a result, this type cannot be directly imported for reuse in scripts/packages, unlike the `Config` type, for example: 

https://github.com/svg/svgo/blob/97d7d00add2643e9b4af016cd34d761a1e044b21/lib/svgo.d.ts#L69-L89

https://github.com/svg/svgo/blob/97d7d00add2643e9b4af016cd34d761a1e044b21/lib/svgo.d.ts#L91-L93

image

If it makes sense, I would like to be able to import the `Output` type in conjunction with the `Config` type (I use SVGO mostly programmatically):

```ts
import type { Config, Output } from "svgo";
```

If I'm correct, the required change is as follows ([`svgo.d.ts` file](https://github.com/svg/svgo/blob/v4.0.0-rc.1/lib/svgo.d.ts)):

```diff
- type Output = {
+ export type Output = {
  data: string;
};
```

**Current workaround**

```ts
type Output = ReturnType<typeof optimize>
```

**Additional context**

- SVGO version: 4.0.0-rc.1

Let me know what you think and if I can open a PR. Thanks!

joaopalmeiro:
Hi! 👋 

Closes #2074

### DIFF
--- a/lib/svgo.d.ts
+++ b/lib/svgo.d.ts
@@ -88,7 +88,7 @@ export type Config = {
   datauri?: DataUri;
 };
 
-type Output = {
+export type Output = {
   data: string;
 };
 


SethFalco:
LGTM! Thank you for the contribution, I'll get a new release candidate version this weekend.

If you encounter any other problems with v4, please do let us know!

PR #2081, opened 2024-10-15
feat: export type `CustomPlugin` and `PluginConfig`

ntnyq:
This PR adds export type `CustomPlugin` and `PluginConfig`.

Which makes it easier to secondary development.


SethFalco:
LGTM! Thank you for opening the PR!

PR #2087, opened 2024-11-22
Fix typos

NathanBaulch:
Just thought I'd contribute some typo fixes that I stumbled on. Nothing controversial (hopefully).

FWIW, the typos are:
* adapated
* effecitively
* equvilent
* gurantee
* illustator
* infront
* minifys
* presedence
* quibic
* seemlessly
* wolffram


plugins/plugins-types.d.ts
```diff
@@ -96,7 +96,7 @@ type DefaultPlugins = {
     useMqs?: string[];
     /**
      * Pseudo-classes and elements to use. An empty string indicates all
-     * all non-pseudo-classes and elements.
+     * are non-pseudo-classes and elements.
```
SethFalco, line 99:
```suggestion
     * non-pseudo-classes and elements.
```

Pretty sure this one should omit the `all` rather than replace it.

plugins/removeAttrs.js
```diff
@@ -101,10 +101,10 @@ export const fn = (root, params) => {
     element: {
       enter: (node) => {
         for (let pattern of attrs) {
-          // if no element separators (:), assume it's attribute name, and apply to all elements *regardless of value*
+          // if no element separators (:), assume its attribute name, and apply to all elements *regardless of value*
```
SethFalco, line 104:
```suggestion
          // if no element separators (:), assume it's attribute name, and apply to all elements *regardless of value*
```

This one should be `it's`, short for `it is`. Though the comment in general isn't grammatically correct so I can see how this would be confusing.

Same applies to the other instance in the file.

test/plugins/reusePaths.06.svg.txt
```diff
@@ -1,5 +1,5 @@
 Don't remove and reuse the ID of the duplicate path if it's already being linked
-in an href by another node.
+in a href by another node.
```
SethFalco, line 2:
```suggestion
in an href by another node.
```

This one should be `an`! (href is pronounced aytch-ref, which starts with a vowel sound!)

SethFalco (changes requested):
Hey hey!

Thank you for looking through and correcting the typos! An embarrassing number of them were introduced by me.

I've left a few comments as I think some of the corrections are incorrect, but most of the changes look good.

test/plugins/reusePaths.06.svg.txt
NathanBaulch: Ha, in my defence I pronounce it hay-ch-ref. Also, there's another instance of "a href" [here](https://github.com/svg/svgo/blob/main/plugins/prefixIds.js#L210) so I went for consistency.

SethFalco: That's also wrong too then!

Can't speak for every country/culture, but in both US and UK, the uppercase character `H` does not start with a `h` sound.

Reference (this page includes an audio file for both UK and US pronunciations):  
https://dictionary.cambridge.org/us/pronunciation/english/h

Then, following this, if to use `a` vs `an` is based on the _sound_ of a vowel, but not the literal character used. So because an uppercase `H` starts with a vowel sound, it should be `an`. (The same reason why "It's an honor." is spelled with `an`, not `a`. `honor` does not start with a vowel character, but the way it's pronounced starts with a vowel sound.)

So href, assuming one pronounces `h` with the name of the uppercase letter, and then `ref` as in the first syllable of `reference`, is pronounced aytch-ref.

I'm not personally going to accuse you of pronouncing anything wrong, as I can see you're from Australia based on your GitHub profile, and it seems to vary there. But our docs are written in American English, so there is a single answer in our case!

SethFalco (approved):

PR #2092, opened 2025-01-27
Update links in _transforms.js

LordArugula:
Fixed link in doc comments. Previous link reaches a 404.

Old:
https://frederic-wang.fr/decomposition-of-2d-transform-matrices.html

New:
https://frederic-wang.fr/2013/12/01/decomposition-of-2d-transform-matrices/


SethFalco (approved):
Thank you very much!

Issue #2078, opened 2024-09-30
`optimize` types missing from svgo-node.d.ts in 4.0.0-rc.1


bschlenk:
**Describe the bug**
I'm using 4.0.0-rc.1 to get around a bug in v3, but types are missing for the `optimize` function. I'm currently working around it by doing

```ts
// @ts-expect-error types are bad
import { optimize } from 'svgo'
```

but it'd be nice to have types here!

**To Reproduce**

1. install `svgo@4.0.0-rc.1` in a typescript project
2. try to do `import { optimize } from 'svgo'`
3. notice typescript complains

**Expected behavior**
I'd expect types to exist for svgo's core function.

**Screenshots**

<img width="512" alt="Screenshot 2024-09-29 at 21 42 09" src="https://github.com/user-attachments/assets/d20923c7-e46b-445a-98a8-fc09f3f53e8f">

**Desktop (please complete the following information):**

- SVGO Version 4.0.0-rc.1
- NodeJs Version 20.16.0
- OS: macOS Sequoia 15.0

jovanmaric:
I have the same issue. The original [Type definitions](https://github.com/svg/svgo/blob/v3.3.2/lib/svgo.d.ts) contained `optimize` but the export from [rc1](https://github.com/svg/svgo/blob/v4.0.0-rc.1/lib/svgo-node.d.ts) doesn't seem to export them properly.

jovanmaric:
Thanks for your quick response. I've added a [PR](https://github.com/svg/svgo/pull/2099) which fixes (I think, haven't been able to run tests) the `optimize` missing error. I still get type errors for the following:

```js
// Module '"svgo"' has no exported member 'optimize'.ts(2305)
import { optimize } from 'svgo';

// Module '"svgo"' has no exported member 'PluginConfig'.ts(2305)
import type { PluginConfig } from 'svgo';
```

Not being that fluent in typescript, I'm not fully sure how to fix those.

SethFalco:
Hey hey!

I've been taking a look at this, but I've had trouble reproducing the issue. Could one of you share a minimal example? I've initialized a new Node.js project in a temp directly and added a dependency on `svgo@4.0.0-rc1`. The `optimize` function is exported according to TypeScript in all circumstances that I've tried.

**index.cjs**

```js
const a = '<svg></svg>';

import { optimize } from 'svgo/browser';
optimize(a);

import svgo from 'svgo/browser';
svgo.optimize(a);
```

**index.mjs**

[...]

PR #2099, opened 2025-04-07
fix(import): duplicate export definition for `svgo-node.d.ts`

jovanmaric:


SethFalco:
I've been unable to reproduce the issue, however this doesn't look right.
We export twice on purpose, once for default export and again for named exports.

That both of the following will work:

```js
// default export
import svgo from 'svgo';
svgo.optimize();

// named exports
import { optimize } from 'svgo';
optimize();
```

If you omit `export {…}` then we lose named exports, if you remove `export default {…}` then we lose default exports.

React doesn't do it the same way afaik, but in terms of interface, React has something similar:

```js
import React from 'react';
import { Fragment } from 'react';
React.Fragment === Fragment;
```

We're doing it this way to reduce the scope of the migration between SVGO v3 and SVGO v4, otherwise everyone that did `import … from 'svgo'` from one or the other which have to switch all instances of it.

SethFalco:
> Doesn't the react implementation simply use only named exports like here

That's why I said in terms of interface. React build their project differently from us though, so in practice it behaves like a CommonJS export.

When SVGO was CommonJS, users could use either the named or default exports to import `svgo`, and they behaved the same. You're able to try this yourself with v3.0.2. [There is no default export](https://github.com/svg/svgo/blob/ae32acf5144be318872a45f49b3ece50c0a4cb18/lib/svgo.d.ts), yet it can still be imported that way because of the implicit default export.

However, now that we've declared SVGO as a module (creating the CJS bundle as part of the build process), these are two distinct things, named exports do not yield an implicit default export, so if we want to continue supporting that, we need to add one ourselves.

> don't think it's possible to have both named and default exports in a file.

Yes you can! (At least in ESM, might make less sense for CJS.)

You've probably written or done an import like this before:

```js
export const a = …
export default const b = { …
```

Which allows the following: `const MyModule, { a } from 'my-module';`

> So only using named exports would still make the following syntax possible. import svgo from 'svgo';

If you can, feel free to share a branch where you remove the default exports but have this working. But on my end, I get this error when I try to do a default import from ESM:

```
import svgo from 'svgo';
       ^^^^
SyntaxError: The requested module 'svgo' does not provide an export named 'default'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:146:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:229:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:473:24)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:122:5)

Node.js v20.18.3
```

**index.cjs** :heavy_check_mark: 

```js
const a = '<svg></svg>';

const { optimize } = require('svgo');
optimize(a);

const svgo = require('svgo');
svgo.optimize(a);
```

**index.mjs** :x:

```js
const a = '<svg></svg>';

import { optimize } from 'svgo';
optimize(a);

import svgo from 'svgo'; // SyntaxError: The requested module 'svgo' does not provide an export named 'default'
svgo.optimize(a);
```

Removing the default exports is fine when the project is depended on from a CJS project, but will not work for when depended on from an ESM project. (Not accounting for any magic bundlers might do.)

PS: I'm working on adding some tests to the project to verify this behavior, and to test our TypeScript definitions. I'll hopefully have a PR open today or tomorrow. I'm also reviewing your comment in https://github.com/svg/svgo/issues/2078#issuecomment-2830155576 and will come back to you once I've done more testing locally.

Meanwhile, thank you for your patience and for helping investigate this issue.

jovanmaric:
> That's why I said in terms of interface. React build their project differently from us though, so in practice it behaves like a CommonJS export.

It looks like you're right. Knowing I could do both in React, after testing a bit in node I get the same error, where a default export is not implicit and should be declared for ESM to be imported that way. Looks like babel also does some import rewriting automagically, which "normalizes" these esm like imports. I'll have to look into ESM a bit more, as i'm making wrong comfy assumptions :sweat: 

Thanks a lot for the thorough explanation and taking the time to research these issues!

SethFalco:
@jovanmaric

Just noting here since I explicitly said I'd like to continue maintaining the default export too, but have gone a different direction now after further consideration, and resolving the issue we were having with types.

You can refer to this for more info: https://github.com/svg/svgo/pull/2118

Thanks again for chiming in on the issue! It was very helpful.

PR #2105, opened 2025-04-16
feat: use rolldown replace rollup to build

wChenonly:
this pr use  [rolldown](https://rolldown.rs/) replace rollup

remove 3 dependencies.

Rolldown Fast Rust-based bundler for JavaScript with Rollup-compatible API

The current package has not yet been compressed
Because the compression of rolldown is still unstable,I did a test with terser and [rollup-plugin-esbuild](https://github.com/egoist/rollup-plugin-esbuild) and can see the compressed size

use terser minify
image

use rollup-plugin-esbuild minify
image

rollup minify  size and time

image

image

rolldown no minify size and time 

image

For minify, would like to see more comments, in the meantime use terser or rollup-plugin-esbuild
After rolldown is stabilized, I will use rolldown minify.


SethFalco:
There are no plans to switch our bundler anytime soon, though I'd be happy to consider it in future.

If we were to change the bundler, we will most certainly not be using a beta or release candidate version. We'll either use something already stable, or wait for the bundler to become stable.

Thanks for the suggestion in any case! I can see you've chimed in on a few other issues too, I'm finally taking 2 weeks off work after this week, so I should be able to check out the other issues. 👍🏽

PR #2122, opened 2025-05-06
fix: fix `loadConfig` types

ntnyq:
`configFile` is optional in 4.0.0-rc.1

https://github.com/svg/svgo/blob/49954bc0d9ec3dcf166bef59657ad1e78909ff1e/lib/svgo-node.d.ts#L10-L17


SethFalco:
Thank you for spotting this, that was my bad! I'll add some tests to our `.test-d.ts` file too, and release this today.

PR #2126, opened 2025-05-08
Added a way to specify the regression test timeout as larger SVGs wou…

Lorfdail:
…ld easily go beyond the default of 30 seconds.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,10 +43,6 @@ Finally, make sure all quality assurance checks pass before making changes. This
 yarn run qa
 ```
 
-### Regression Tests
-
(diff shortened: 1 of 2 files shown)


SethFalco:
Thanks for your contributions!

I made the following amendments before merging:

* Instead of setting the timeout only on `#goto`, uses `BrowserContext#setDefaultTimeout`. This will use the specified timeout for all interactions, as screenshotting also needed its timeout extended, otherwise CI would fail.
* Increased the timeout to 10 minutes. I'm wary that 2 minutes isn't enough, so have decided that 10 minutes will be a cut off. If a file needs more than 10 minutes in CI, the performance issue must either be resolved, or that file must be excluded to run on CI until we can bring it down to < 10 minutes consistently.
* Minor rewording of the documentation, and I moved it below the implementation sections.

I will merge it despite the workflow failing for now, for two reasons:

1. One of the new SVGs in the test suite can not be efficiently optimized by SVGO at the moment. I haven't investigated in-depth yet, but I think it's the mergePaths plugin that's blocking it. The problem is, mergePaths has a check where if there are multiple siblings `path` elements under the same parent, it will check every combination of of these paths. This is normally fine, but one of the files has a parent containing 400,000~ paths. XD That's not feasible to optimize anytime soon.
2. There are some known differences caused by the latest update of Chromium/Playwright which introduced regression failures.

I will disable those SVGs in a separate PR, then they can be investigated later.

PR #2128, opened 2025-05-10
fix(applyTransform): non-scaling-stroke works when no skew

canstand:


SethFalco:
Hey hey!

Thanks for contributing to SVGO! Could you please include a test case that reproduces the problem, so we know what this is fixing?

canstand:
Hi, the modified test svg looks like this:

image

PR #2157, opened 2025-07-10
`.coords` is bad actually

KTibow:
it turns out in the past i used `item.coords` (post-instruction coordinate in ORIGINAL svg) when i should've been using `relSubpoint`; this switches to `relSubpoint` when fitting.

### DIFF
--- a/plugins/convertPathData.js
+++ b/plugins/convertPathData.js
@@ -877,10 +877,8 @@ function filters(
         (command === 'l' || command === 'h' || command === 'v')
       ) {
         if (
-          // @ts-expect-error
-          Math.abs(pathBase[0] - item.coords[0]) < error &&
-          // @ts-expect-error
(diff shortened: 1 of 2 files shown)


KTibow:
the failing test cases are things like
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 200">
<path d="M 78.301922,39.376734 L 64.971121,61.742246 L 73.061784,64.348229 L 78.512656,70.6691 L 84.285326,65.033743 L 92.451569,61.727261 L 78.301922,39.376734 z" fill="none" stroke="black" stroke-width="2"/>
</svg>
```

and the fail is because SVGO incorrectly rounds in a way that makes the start and end no longer connected

and the incorrectly coded "convert going home to z" logic was the only thing stopping the failure from surfacing

so fixing one thing shows another broken thing causes regressions... oh well i'll make a fix

PR #2161, opened 2025-07-21
added https versions for sourceforge and inkscape

juronja:


SethFalco:
Hey hey!

Thanks for opening a PR! Can you confirm for me where you saw these appear from the `https://` scheme? Usually tools use `http://`, not `https://`. I'm not aware of these tools ever using the `https://` schema. :thinking:

The `xmlns` is more of a URI (identifier) than a URL (locator/link). The fact that it looks like a URL and can often be accessed can cause confusion, but isn't stable/documented behavior.

For Inkscape, I'm 99% sure you should have `http://`, `https://` is just invalid.

> XML namespace URLs are really opaque identifiers. The fact that some namespace URLs under w3.org resolve to a page describing them is a convenience.
> …
> And notably, if you create some content with `<svg xmlns="https://www.w3.org/2000/svg">`, then browsers and other UAs will not recognize that as a real SVG `<svg>` element.
> 
> — https://github.com/w3c/svgwg/issues/738

I think these should remain in your custom config, or even better, you should amend your SVGs and replace `https://` with `http://` as that what they're supposed to be.

I'll close this for now, but happy to reopen if you have more context for us. I did look myself briefly, but yeah I only see that `http` is supported.

juronja:
@SethFalco  sure, np. Thank you for considering.

I used https://dashboardicons.com/ for a source for my project and there are some.

Jenkins SVG is one example I can remember: https://dashboardicons.com/icons/jenkins
it has: xmlns:sodipodi="https://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd" 

I can't remember for inkscape, but I can find it if that would help. 

edit:
jenkins has also inkscape

SethFalco:
Thank you for sharing the context! It's actually very useful to know the source/motivation for these changes.

If there is any evidence that this is something vector editors (or other tools that support vector exports) actually insert, and wasn't done by some miscellaneous linting tool or hand-edited in afterward, then I'd be happy to accept a PR like this.

However, atm I get the feeling someone just misconfigured their linter or maybe hand-edited these to `https://` when they shouldn't have. Since this seems to be a rare occurrence, and is _technically_ a different URI, yeah I'm leaning towards leaving it for users to handle.

But your report is insightful nonetheless! If this is something that comes up more often, we'll review the state of things again.

juronja:
oh, I see. Makes sense, the additionalNamespaces should be used for the edge cases.
Is just that I spent a few frustrating hours, thinking the svgo tool does not work, before I noticed the https protocol difference. 🙈 
Thank you again for your feedback and time.

PR #2164, opened 2025-08-09
fix(convertPathData): introducing isSafeToRemove

### LINKED ISSUE #2158: convertPathData removes valid paths
**Describe the bug**
If path contains a 0 height shape, plugin removes that part of path.

**To Reproduce**

Example script:
```js
import { optimize } from 'svgo';

const svg = optimize(
	`<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 16 16" width="16" height="16" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5">
<path d="m8 5.25v0m0 6v-3.5"/>
</svg>`,
	{
		plugins: ['convertPathData'],
	}
);

console.log(svg.data);
```

**Expected behavior**
Expected path to contain first chunk with `v0`, but actual output is:

```svg
<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 16 16" width="16" height="16" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="1.5"><path d="M8 11.25v-3.5"/></svg>
```

Note that path contains only last chunk of original path, first chunk is gone, resulting in missing dot shape in icon.

- SVGO Version: 4.0.0, 3.3.2
- Node.js Version: 20.14.0
- OS: macOS 15.5

**Additional context**
Source icon used in example: https://github.com/jaynewey/charm-icons/blob/main/icons/info.svg

### LINKED ISSUE #2163: old issue killing most content is back
Same as in https://github.com/svg/svgo/issues/890

The file http://www.jirka.org/nlin-sierpinski-mbx.svg
gets totally destroyed by svgo.  Disabling the plugin convertPathData it is not as good as in the past, but it doesn't delete content.

Last version where it worked was 3.0.4 and it starting eating the content in 3.0.5 and is broken in all subsequent versions it seems.

Tried with the newest version 4.0.0, same result: mostly empty output file (the actually sierpinski triangle is gone).

KTibow:
Removing useless commands: a tricky and inconsistent business. SVGO currently uses two approaches:
- **!maybeHasStrokeAndLinecap** for the purpose of removing useless line commands

  This works decently. It allows for removing useless lines, while not removing dots.

  Flaw: Even in a path with linecaps, if there are "dot" commands after the first move, they wouldn't render at all and should be removed. They currently aren't.
- **isSafeToUseZ** for the purpose of removing useless `z` commands

  isSafeToUseZ is defined as either having no stroke or having a all-round stroke. When that's true, a line home and a z command render the same way, and can be exchanged or removed.

  Flaw: This works great when removing a redundant `z` that comes right after another command, but what if the `z` is the first move (for the purpose of creating a dot)? Then it's incorrectly removed. This is shown in #2158 and #2163.

This PR introduces isSafeToRemove to help. It's worth reading through the code, but the TLDR is that it uses a more sane method, categorizing the path into non-stroked, stroked + first draw command, and stroked but non-first command, and using fitting behavior in each case.

Fixes #2158 and fixes #2163.

### DIFF
--- a/test/regression/lists/expect-mismatch.txt
+++ b/test/regression/lists/expect-mismatch.txt
@@ -12,7 +12,6 @@ W3C_SVG_11_TestSuite/svg/styling-css-04-f.svg
 
 # broken and pending fix
 charm-icons-0.18.0/docs/img/splash.svg
-charm-icons-0.18.0/icons/atom.svg
 oxygen-icons-5.116.0/scalable/actions/hidef/tools-rip-audio-cd.svg
 oxygen-icons-5.116.0/scalable/actions/im-ban-kick-user.svg
 oxygen-icons-5.116.0/scalable/actions/im-ban-user.svg


plugins/convertPathData.js
```diff
@@ -844,7 +852,7 @@ function filters(
       }
 
       // remove useless non-first path segments
-      if (params.removeUseless && !maybeHasStrokeAndLinecap) {
+      if (params.removeUseless && isSafeToRemove(prev.command == 'M', true)) {
```
GreLI, line 855: I think `'m'` is also possible.

GreLI:
It seems that this PR kills the optimization case when there is no stroke. Which was the goal at the first place.

KTibow:
> It seems that this PR kills the optimization case when there is no stroke.

What makes you say this?
```js
if (!maybeHasStroke) {
  return true;
}
```
literally means "if there is no stroke, it is always safe to optimize". All the previous test cases also pass.

plugins/convertPathData.js
KTibow: Ah, right - while the first M is converted to absolute and excluded from converting to relative, the others aren't.

plugins/convertPathData.js
```diff
@@ -155,13 +155,25 @@ export const fn = (root, params) => {
...
+            /** @type {boolean} */ isFirstDraw,
+            /** @type {boolean} */ safeIfNotFirstDraw,
+          ) => {
+            if (!maybeHasStroke) {
+              return true;
+            }
+            if (isFirstDraw) {
+              return !maybeHasLinecap;
+            } else {
+              return safeIfNotFirstDraw;
```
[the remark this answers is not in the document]
KTibow: I'd rather not do this - it would reduce clarity for no benefit (no statistically significant performance improvement raw, and if bundled it gets turned into the terse version automatically).

Issue #1302, opened 2020-11-16
Is there any way to distinguish monotone icon from multi-colored icon without manual string operations??


AravinthAro:
Hi I'm working on an icon system for my project with SVG. All monotone icons needs to be styled by CSS and at the same time multi-coloured icons needs to be preserved in it's original state i.e. it should not be changed by CSS at all. I've changed all styles to inline-styles in the icons, so external CSS will not overwrite the icon's styles. This works well for multi-colored icons, but monotone icons remains in black color always. I guess I need to remove all color attributes (fill, stroke) for monotone icons alone. This way I can paint those icons with external CSS. Please let me know is there any way to distinguish monotone icons from the rest with SVGO?? Or any completely different approach to solve this issue?? Thanks in advance.

Sample of icons I'm working with : 

**Monotone icon**
``` html
<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100"><defs><style>.cls-1{fill:#333;}</style></defs><title>Icons</title><polygon class="cls-1" points="34.69 65.23 34.69 32.48 58.93 25.93 58.93 76.2 25.68 71.62 58.93 85.21 78.58 78 78.58 21.5 58.93 14.79 21.42 28.87 21.42 71.45 34.69 65.23"/></svg>
```

**Multi-colored icon**
``` html
<svg xmlns="http://www.w3.org/2000/svg" width="363.191" height="263.904" viewBox="0 0 363.191 263.904">
  <g id="Group_4871" data-name="Group 4871" transform="translate(-1423 -476.129)">
[...]

Issue #1673, opened 2022-05-12
ConvertColors should not convert non-color value.


tychenjiajun:
**Describe the bug**
The value of `fill` and `stroke` is 

> \<paint\> = none | [\<color\>](https://www.w3.org/TR/css3-values/#colors) | [\<url\>](https://www.w3.org/TR/css3-values/#url-value) [none | [\<color\>](https://www.w3.org/TR/css3-values/#colors)]? | context-fill | context-stroke"

setting `currentColor` to true will convert `context-fill`, `context-stroke` and `url` to `currentColor`.

**To Reproduce**

See https://github.com/tychenjiajun/svgo/blob/fa1d6c2b9648129d5c727cfd0794354873263a63/test/plugins/convertColors.05.svg

**Expected behavior**

The above test case should be pass.

**Desktop (please complete the following information):**
 - SVGO Version latest
 - NodeJs Version 16.14.x
 - OS: IDK, I run it in Gitpod
 
**Additional context**

yimity:
also "transparent".

SethFalco:
Hey hey! Thanks for chiming in.

Would you mind sharing your use case for using the `currentColor` option? I'd like more insight on it's use before deciding how we resolve this.

Upon reading the ticket, I'd actually be tempted to close this as I don't think the option needs additional work. In fact, I think it could be worth deprecated that option. The original motivation for the PR that added is no longer relavent afaik.

> if a color is defined within an SVG you can't override the color easily in CSS
> 
> — https://github.com/svg/svgo/pull/521

All browsers that support inline SVGs also support overriding colors easily. In any situation one can use `currentColor`, one can also use CSS.

Even back in IE 9, users could override colors via CSS. 

<img width="2270" height="1334" alt="Internet Explorer 9 opened in BrowerStack. Shows the SVGO.dev website with browser devtools open. I selected a path from an SVG using a selector that targets nodes with an attribute fill set to red and made them blue. This worked, demonstrating that this is supported behavior in IE 9." src="https://github.com/user-attachments/assets/f23f49be-0c7f-4aab-ac84-04b1760f9768" />

From there using `currentColor` is a design choice, not an optimization, seems more of a job for a dedicated tool like vector editing software or actual XML manipulation libraries.

But I won't act on this now, just hoping to hear more from users so I understand what's the motivation to this.

[...]

PR #2165, opened 2025-08-17
feat(convertColors): add excludeMulticolor option to exclude multicolor icons

chip-cmyk:
## Description

When using the `convertColors` plugin with `currentColor` configuration, I found that multi-color icons were also being converted to `currentColor`, making them appear as single-color in actual use. I wanted to preserve the colors of multi-color icons while only converting single-color icons to `currentColor`. I discovered issue #1302 describing the same problem, so I tried adding parameter configuration to differentiate between multi-color and single-color icons.

## New Parameters

- **`value`**: Specifies target colors for conversion (boolean, string, or RegExp)
- **`excludeMulticolor`**: When set to `true`, excludes conversion on multi-color and gradient icons

## Usage Examples

```js
// Convert all colors, but exclude multi-color icons
{ currentColor: { excludeMulticolor: true } }

// Convert only red colors, but exclude multi-color icons
{ currentColor: { value: 'red', excludeMulticolor: true } }
```

### DIFF
--- a/docs/04-plugins/convertColors.mdx
+++ b/docs/04-plugins/convertColors.mdx
@@ -5,7 +5,7 @@ svgo:
   defaultPlugin: true
   parameters:
     currentColor:
-      description: If to convert all instances of a color to `currentColor`. This means to inherit the active foreground color, for example in HTML5 this would be the [`color`](https://developer.mozilla.org/docs/Web/CSS/color) property in CSS.
+      description: If to convert all instances of a color to `currentColor`. This means to inherit the active foreground color, for example in HTML5 this would be the [`color`](https://developer.mozilla.org/docs/Web/CSS/color) property in CSS. Can be a boolean, string, RegExp, or an object with `value` and `excludeMulticolor` options.
       default: false
     names2hex:
       description: If to convert color names to the hex equivalent.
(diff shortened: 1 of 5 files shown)


SethFalco:
Hey hey!

Thanks for contributing this! Before I review it, would you be willing to chime in on #1673?

I was actually thinking about freezing/deprecating the `currentColor` parameter on [`convertColors`](https://svgo.dev/docs/plugins/convertColors/) because this is not an optimization, but rather an editorial decision, and tends to yield unexpected results for most users.

Usually making editorial decisions in this way isn't going to yield desirable results, it's just an inefficient and less predictable way to amend the source, with longer iterations between changes.

I was learning towards deprecating instead of expanding the options to support more use cases, because so far, no one has offered more insight on why we should keep this. The reason it was added (it was hard to override colors in CSS in some clients) is no longer applicable either.

As SVGO is not an SVG editing tool, I think I'd rather keep things simple, and direct users to either hand-edit their SVG, use an actual editor, use an XML library, or a custom SVGO plugin. I wouldn't mind even writing blog-esk guides for this on our website. But I don't think SVGO should support scenarios like this.

I'd love more insight though, if you believe that using this SVGO plugin is actually better than doing it in your vector editor or by hand-editing it, feel free to share!

TL;DR: Imo, rather than adding more maintenance/complexity to the `currentColor` option, users should be encouraged to _fix_ their SVG. Not make editorial decisions via an SVG optimization tool. Especially as the only way to define the config you've proposed, is by reading the SVG code already, it would've been faster and better to edit the source, right?

chip-cmyk:
Thank you for the detailed explanation! I've read #1673 regarding the potential issues with `currentColor`conversion, and I fully understand your intention to keep SVGO's core functionality clean and simple.

In our project, this feature has been very helpful in solving a specific problem. We're working with many Figma-exported icons that need to adapt to different themes through CSS color control, similar to text styling. At first, I found that setting `fill="currentColor"`via CSS wasn't taking effect because the exported SVGs contain their own fill and stroke attributes. As developers with only view permissions on the design files ,manually editing all these Figma-exported icons would be difficult and time-consuming.

I've tried other approaches like using `removeAttrs`to remove fill/stroke attributes or Figma SVG export plugins, but they don't handle special cases like `fill="none"`or mask elements well. The `currentColor`conversion has worked better for our needs.

If you still prefer to remove this feature, I completely understand. Could you share your thoughts on alternative approach for similar requirements? Before submitting this PR, I actually implemented similar functionality through a custom SVGO plugin. However, I found it difficult to access some internal methods like `includesUrlReference`. Making more of these utilities available could be helpful for developers creating custom plugins.

image

PR #2169, opened 2025-08-26
Add removing data-figma-skip-parse attrs

Connormiha:
I added plugin to remove `data-figma-skip-parse` attribute, which added Figma.

### DIFF
--- a/docs/04-plugins/removeFigmaAttrs.mdx
+++ b/docs/04-plugins/removeFigmaAttrs.mdx
@@ -1,0 +1,8 @@
+---
+title: removeFigmaAttrs
+svgo:
+  pluginId: removeFigmaAttrs
+  defaultPlugin: true
+---
+
+Remove Figma attributes from elements in the document.
--- a/lib/builtin.js
+++ b/lib/builtin.js
@@ -33,6 +33,7 @@ import * as removeElementsByAttr from '../plugins/removeElementsByAttr.js';
 import * as removeEmptyAttrs from '../plugins/removeEmptyAttrs.js';
 import * as removeEmptyContainers from '../plugins/removeEmptyContainers.js';
 import * as removeEmptyText from '../plugins/removeEmptyText.js';
+import * as removeFigmaAttrs from '../plugins/removeFigmaAttrs.js';
 import * as removeHiddenElems from '../plugins/removeHiddenElems.js';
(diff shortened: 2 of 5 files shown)


GreLI:
You can already just set `keepDataAttrs` attrubute to `false` in [removeUnknownsAndDefaults](https://svgo.dev/docs/plugins/removeUnknownsAndDefaults). It will remove all data-* attributes. It's default value is `true`.

Connormiha:
> You can already just set `keepDataAttrs` attrubute to `false` in [removeUnknownsAndDefaults](https://svgo.dev/docs/plugins/removeUnknownsAndDefaults). It will remove all data-* attributes. It's default value is `true`.

Oh. I didn't know about this option. I usually use the CLI with svgo *.svg. But today, while looking through the files, I noticed these data attributes.

SethFalco:
Hey hey! Thanks for contributing to SVGO!

Indeed, this shouldn't be necessary as GreLI is rightly pointed out, this use case is already supported.

An SVGO config for you would look like this (admittedly a little clunky):

**`svgo.config.mjs`**
```mjs
/** @type {import('svgo').Config} */
export default {
  plugins: [
    {
      name: 'preset-default',
      params: {
        overrides: {
          removeUnknownsAndDefaults: {
            keepDataAttrs: false
          }
        }
      }
    }
  ]
}
```

Explanation:

* `/** @type {import('svgo').Config} */` just adds type checking.
* `name: 'preset-default',` specifies the default preset, which you've been using already.
* `keepDataAttrs: false` prevents preserving data attributes.

You can either put this file in the same directory used to invoke SVGO, or if you want to use this as a default configuration (unless overridden otherwise), you can put it in your home directory. SVGO looks up the parent directories recursively until it encounters a config, so that's fine if you want to use the same config across all your workspaces.

I'll close this since it should already be resolved. :+1: 

---

On the side, feedback is always welcome. If you have time, I'd welcome if you'd share with us how this may have been overlooked.

Did you do anything between noticing the `data-figma-skip-parse` and submitting this PR, that you feel should've revealed to you the existence of this feature already?

If you jumped straight to doing a PR, that's fine! But for the sake of better discovery, I'm just wondering did you search the docs or try anything else first where we failed to convey the existence of this feature/parameter to you? (If you searched online, search terms would be useful too!)

No pressure to answer ofc. :heart:

PR #2173, opened 2025-09-21
Handle transparent colors in removeUselessStrokeAndFill

viralcodex:
- Added isTransparent utility to detect transparent and zero-alpha color values for fill and stroke attributes.
- Updated removeUselessStrokeAndFill plugin to treat these values as 'none', improving SVG cleanup.
- New tests to verify correct removal of elements with transparent or zero fill/stroke.

Hi @SethFalco,

I initially thought of adding the code changes with removeHiddenElems, but it was getting tricky with the previous checks and to integrate with them the checks for the transparency and color alpha 0.
I felt it would be better to have the enhancements to the removeUselessStrokeAndFill, which already handled most of it from what I understood (can be wrong ofc).

Can be reviewed and let me know if any changes are needed or we can do it in some other way also.

Let me know what u think.

Thanks and have a great day!

### DIFF
--- a/lib/svgo/tools.js
+++ b/lib/svgo/tools.js
@@ -1,4 +1,5 @@
 import { attrsGroups, referencesProps } from '../../plugins/_collections.js';
+import * as csstree from 'css-tree';
 
 /**
  * @typedef CleanupOutDataParams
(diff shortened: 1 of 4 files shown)


viralcodex:
Hi @SethFalco, can we have a review on this? Would love to know if it is needed or not/enhance it in anyway. thanks 😄

Issue #2154, opened 2025-07-07
convertStyleToAttrs not enabled by default, or missing "(disabled by default)" in svgo --show-plugins


qubodup:
Looking at `svgo --show-plugins` disabled plugins are marked as such.

```bash
$ svgo --show-plugins | grep tyle
 [ convertStyleToAttrs ] converts style to attributes
 [ inlineStyles ] inline styles (additional options)
 [ mergeStyles ] merge multiple style elements into one
 [ minifyStyles ] minifies styles and removes unused styles
 [ removeStyleElement ] removes <style> element (disabled by default)
```

It looks like `convertStyleToAttrs` is enabled by default, but it is not.

**To Reproduce**
1. Prepare `circle.svg`:
```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20"><circle cx="10" cy="10" r="10" style="fill:red"/></svg>
```
2. Run `svgo circle.svg -o circleout1.svg`, resulting in no change.

3. create `svgo.config.mjs`:
```js
export default {
  plugins: [
    {
      name: 'preset-default',
    },
    'convertStyleToAttrs'
  ],
};
```
4. Run `svgo circle.svg -o circleout2.svg`, resulting in `0.112 KiB - 5.2% = 0.106 KiB` and the following file:
```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20"><circle cx="10" cy="10" r="10" fill="red"/></svg>
```

**Expected behavior**
convertStyleToAttrs should be enabled by default or
`--show-plugins` should state "(disabled by default)"

**Screenshots**
<img width="20" height="20" alt="Image" src="https://github.com/user-attachments/assets/6abaf763-7d76-4054-b3e7-3a785b4b93f8" />

**Desktop:**

- SVGO Version 4.0.0
- Node.js Version v22.16.0
- OS: Windows 10 10.0.19045.5965

zk2k2:
Hi, can you please assign me this? I would like to contribute, thanks!

SethFalco:
Hey @zk2k2!

Thanks for your willingness to contribute! There's no need to request assignment. Anyone is welcome to pick up whatever they want!

I'll share some guidance on how you could do this. Feel free to ask questions if you get stuck!

## Quick Fix

All plugins can have the details about if they're enabled or disabled by default removed.

Example output:

```
$ svgo --show-plugins | grep tyle
[...]

PR #2174, opened 2025-09-26
Updated plugin descriptions and added logic to set preset to the plugin description.

viralcodex:
Closes: #2154
- Removed '(disabled by default)' from several plugin descriptions to prevent any discrepancy.
- added logic to display which presets each plugin belongs to in the plugin list output.

### DIFF
--- a/lib/svgo/coa.js
+++ b/lib/svgo/coa.js
@@ -524,10 +524,33 @@ function checkWriteFileError(input, output, data, error) {
   }
 }
 
-/** Show list of available plugins with short description. */
+/**
+ * Show list of available plugins with short description with the presets it belongs to.
+ */
 function showAvailablePlugins() {
+  const pluginToPresetsMap = new Map();
+
+  for (const plugin of builtinPlugins) {
+    if (plugin.isPreset) {
+      for (const presetPlugins of plugin.plugins) {
+        if (!pluginToPresetsMap.has(presetPlugins.name)) {
+          pluginToPresetsMap.set(presetPlugins.name, [plugin.name]);
+        } else {
+          pluginToPresetsMap.get(presetPlugins.name).push(plugin.name);
+        }
+      }
+    }
+  }
   const list = builtinPlugins
-    .map((plugin) => ` [ ${colors.green(plugin.name)} ] ${plugin.description}`)
+    .map((plugin) => {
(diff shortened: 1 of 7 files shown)


viralcodex:
@SethFalco and team,
Please review and let me know if any other changes are needed.
Thanks

CURRENT OUTPUT AFTER THE CHANGES:

```bash
svgo % node bin/svgo.js --show-plugins | grep tyle
[ convertStyleToAttrs ]  converts style to attributes
[ inlineStyles ]  inline styles (additional options) (preset-default)
[ mergeStyles ]  merge multiple style elements into one (preset-default)
[ minifyStyles ]  minifies styles and removes unused styles (preset-default)
[ removeStyleElement ]  removes <style> element
```

viralcodex:
Hi, Made the requested changes, please have a look 😄 

One Thought regarding this from my side that the preset text can have a distinct formatting, maybe color, so that it is easily visible in the list of show-plugins?

@SethFalco , what do you think?

SethFalco (approved):
LGTM! :tada: 

I did rebase for you. For future reference, I think it can be better to rebase a PR with upstream/main rather than merge upstream/main into your branch.

This keeps the git history a little cleaner, which is especially helpful during reviews, so I won't see changes unrelated to your pull request. Not everyone loves to rebase though, so up to you if you use it!

In any case, I'm happy with the implementation, and appreciate you taking the time to work on this fix.

I did make one tweak to it during my rebase, but have otherwise kept it as-is. In the original implementation, each item in the list was prefixed with a whitespace. Your PR removed this space, so I added it back.

In future, I'd like to review the formatting of the CLI, but we'll do this as a separate ticket. For this fix, I'd like to maintain as much of existing interface as possible outside of the obvious bugs.

Thank you very much for your work and patience!

viralcodex:
Hi @SethFalco,

Thanks for the support through the process. This was my first PR on a project I used in my own project and really happy to be able to contribute, will continue to do more in the future.

Yeah, I did the rebase from upstream but I think I did it in the checked out branch instead of first doing it in the main branch of my fork. Sometimes it happens with me with git (in work too 🥲).

For the future formatting for the CLI, would like to work on that too and learn and contribute more.

Overall, a great thing for me as well.

Thanks to the team of SVGO.

Issue #2073, opened 2024-09-08
prefixIds does not apply prefix in non-trivial begin/end expressions of animate tag


alex-kowalczyk:
**Describe the bug**
Seems that https://github.com/svg/svgo/issues/848 is fixed only partially, for simple expressions - [the fixing commit](https://github.com/svg/svgo/commit/fafcbd7bd00de03d93ab55429e5841bc70681a14#diff-c28ffd9bde80c0a95d54215e9636676384bababddd916b4786156a58b9598546) splits by '; ' and checks for .end presents via endsWith() 

For the following animation of rain drops, coming from https://github.com/basmilius/weather-icons (partly-cloudy-day-rain.svg):
```
                <animate id="y1" attributeName="opacity" begin="0s; y1.end+.33s" dur=".67s" keyTimes="0; .25; 1" values="0; 1; 0"/>
```
the result is:
```
                <animate id="partly-cloudy-day-rain_svg__j" attributeName="opacity" begin="0s; j.end+.33s" dur=".67s" keyTimes="0; .25; 1" values="0; 1; 0"></animate>
```

**To Reproduce**

**Expected behavior**
```
                <animate id="partly-cloudy-day-rain_svg__j" attributeName="opacity" begin="0s; partly-cloudy-day-rain_svg__j.end+.33s" dur=".67s" keyTimes="0; .25; 1" values="0; 1; 0"></animate>
```

**Desktop (please complete the following information):**

- SVGO Version: 3.3.2
- NodeJs Version: 20
- OS: Linux

valtism:
Also ran into this issue using SVGR. Basic `id`s and `id.end` work, but having `id.end + 0.06s` does not get prefixed and breaks

alex-kowalczyk:
If anyone interested, I fixed (or at least improved) that in a forked branch https://github.com/alex-kowalczyk/svgo/commits/prefixes-fix/ , namely https://github.com/alex-kowalczyk/svgo/commit/f797003a684b6e125749443dd07f0441f32a1752  (although not enough time & capabilities to prepare full-fledged PR).

GreLI:
@alex-kowalczyk _id_ (the part before “.start” or “.end”) in XML is defined as [NameStartChar](https://www.w3.org/TR/xml/#NT-NameStartChar) ([NameChar](https://www.w3.org/TR/xml/#NT-NameChar))*, where

```
[...]

PR #2208, opened 2026-03-19
fix(prefixIds): does not handle SMIL timing offsets

yangxu52:
This PR **improves** handling of `begin` and `end` attributes in the `prefixIds` plugin.

**Problem**

The current implementation only prefixes values ending with `.end` or `.start`, but valid SMIL syntax also includes offsets such as:

- `a.end-0.5s`
- `b.begin+0.1s`

These cases are currently ignored, leading to broken animation references after ID prefixing.

**Solution**

Replace the suffix check with a regex that correctly parses:
`<id>.(begin|end)(+/-offset)?`
This ensures all valid SMIL timing references are properly prefixed.

**Impact**

- Fixes broken animation chaining
- Maintains backward compatibility
- No effect on non-SMIL attributes

**Example**

Before:

```xml
begin="a.end-0.5s"
```

After

```xml
begin="prefix-a.end-0.5s"
```

**Test**

1. without offsets, like `x.end`, OK
2. with offsets, like `b.end-0.5s`, OK
3. mix and multiple, like `x.begin;x.end-0.5s`, OK

Test code e.g:

```js
it('should prefix SMIL begin/end with offsets', () => {
  const svg = `
    <svg>
      <animate id="a" begin="b.end-0.5s" />
      <animate id="b" begin="a.begin+0.1s" />
    </svg>
  `

  const result = optimize(svg, {
    plugins: [
      'cleanupIds',
      {
        name: 'prefixIds',
        params: { prefix: 'p' },
      },
    ],
  })
  expect(result.data).toContain('p-b.end-0.5s')
  expect(result.data).toContain('p-a.begin+0.1s')
})
```

image

**Notice**
This PR intentionally keeps a minimal regex-based approach consistent with existing implementation, while extending coverage to valid SMIL offset syntax instead of all.

**Linked Issue**
Fixes: #2207 #2073

### DIFF
--- a/plugins/prefixIds.js
+++ b/plugins/prefixIds.js
@@ -257,12 +257,13 @@ export const fn = (_root, params, info) => {
             node.attributes[name] != null &&
             node.attributes[name].length !== 0
           ) {
-            const parts = node.attributes[name].split(/\s*;\s+/).map((val) => {
-              if (val.endsWith('.end') || val.endsWith('.start')) {
-                const [id, postfix] = val.split('.');
-                return `${prefixId(prefixGenerator, id)}.${postfix}`;
-              }
-              return val;
+            const parts = node.attributes[name].split(/\s*;\s*/).map((val) => {
+              return val.replace(
+                /^([a-zA-Z0-9_-]+)\.(begin|end)([+-].+)?$/,
+                (_, id, type, offset = '') => {
+                  return `${prefixId(prefixGenerator, id)}.${type}${offset}`;
+                },
+              );
             });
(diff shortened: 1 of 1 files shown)


yangxu52:
Hi, @SethFalco 

This PR has been submitted for some time, and both CI tests have passed. 
It is also associated with Issue #2207 #2073 
I wonder if it would be convenient for you to help review it?
If there are any necessary modifications/additions, I will follow up at any time.

Thanks!

yangxu52:
relate: #2288

PR #2211, opened 2026-03-19
feat: open importing to anything in lib

### LINKED ISSUE #1612: svgo/lib/svgo/svg2js is no longer available - is there a replacement?
**Describe the bug**
Since 2.7.0 the file `svgo/lib/svgo/svg2js` no longer exists... it is used in https://github.com/Ashung/svg2vectordrawable

Is there a "public" API with the capabilities of svg2js?

**To Reproduce**
1. `mkdir test; cd test`
2. `npm i svg2vectordrawable svgo`
3. `$ npx s2v -h`
4. 
   ```
   node:internal/modules/cjs/loader:936
     throw err;
     ^

   Error: Cannot find module 'svgo/lib/svgo/svg2js'
   Require stack:
   // ...
   ```

**Expected behavior**
`s2v` to run

 - SVGO Version 2.7.x+
 - NodeJs Version v16.11.1
 - OS: Mac
 
**Additional context**

I search through some issues and it seems svg2js is considered a private API? if so how about releasing a new major version with an "export" map then those apis will become actually private (e.g. users will no longer be able to import anything that is NOT in the export map)

https://nodejs.org/api/packages.html#exports

KTibow:
should make building custom plugins easier

motivation: https://github.com/svg/svgo/pull/2165#issuecomment-3194388419

should close #1612

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,10 @@
     "./browser": {
       "import": "./dist/svgo.browser.js",
       "types": "./types/lib/svgo.d.ts"
+    },
+    "./lib/*": {
+      "import": "./lib/*.js",
+      "types": "./types/lib/*.d.ts"
     }
   },
   "files": [


KTibow:
At some point I want to execute this sprint (although not at a sprint's pace):
- Support external `lib` imports (this PR)
- Write a plugin with the sole purpose of doing what `convertColors`'s `currentColor` currently does and document it widely (code, docs, release changelog)
- Deprecate `currentColor` (reasoning I've heard: it's not an optimization so it doesn't belong in core, and it lacks many community requested features)
- Release a new version
- Close [issues and PRs](https://github.com/svg/svgo/issues?q=state%3Aopen%20convertcolors%20currentcolor) regarding `currentColor`

SethFalco (commented):
I'm not going to block this PR, but I will weigh in with an opinion!

I don't think it's a great idea to expose the `lib/` directory directly, as this adds too much cruft to our supported interface. I especially do not want us to expose our directory structure in the API.

Would you mind creating an `lib/index.js` file and just exporting all the functions that make sense to you in it? (Even if that's all of them.)

Then people can do this for example:

```js
import { stringifyPathData, computeStyle } from 'svgo/lib';
// stringifyPathData from lib/path.js
// computeStyle from lib/style.js
```

This way you can achieve what you want, which I appreciate—perhaps I was a bit overzealous when sealing up our supported interface.

If we do this, perhaps consider deprecating anything I had exported at `svgo` that will now overlap with the exports to `svgo/lib`. For example, we export some helpers like `querySelectorAll` already, but if we'll officially export helpers for plugins, then they can all just reside in the `svgo/lib` export.

PR #2224, opened 2026-06-19
feat: add maxEntityCount param

### LINKED ISSUE #2218: Add maxEntityCount parameter to sax parser configuration
## Problem

SVGO 3.3.3 replaced `@trysound/sax` with `sax ^1.5.0`. When installing via npm, version 1.6.0 is installed, which introduced XXE protection with a default `maxEntityCount = 512` limit.

This causes `SvgoParserError: Parsed entity count exceeds max entity count` when processing SVG files with more than 512 entities, even though these files are valid and not malicious.

## Root Cause

The sax parser in SVGO's `parser.js` creates a parser without passing the `maxEntityCount` parameter:

```javascript
const sax = SAX.parser(config.strict, config);
```

In sax 1.6.0, this defaults to 512:
```javascript
parser.opt.maxEntityCount = parser.opt.maxEntityCount || 512
```

## Solution

Make `maxEntityCount` configurable via the SVGO API:

1. Add `maxEntityCount` parameter to the sax parser configuration in `lib/parser.js`:

```javascript
const config = {
  strict: true,
  trim: false,
  normalize: false,
  lowercase: true,
  xmlns: true,
  position: true,
  unparsedEntities: true,
  maxEntityCount: options.maxEntityCount || 512,
};

const sax = SAX.parser(config.strict, config);
sax.opt.maxEntityCount = config.maxEntityCount;
```

2. Expose `maxEntityCount` in the `optimize()` function options so users can configure it:

```javascript
optimize(svg, {
  maxEntityCount: 512, // configurable parameter
  // ... other options
});
```

## Impact

This affects any project that:
- Uses SVGO 3.3.3 or later
- Processes SVG files with many entities (complex diagrams, icons with many paths, etc.)
- Builds CLI binaries with pkg that bundle SVGO

## References

- sax 1.6.0 release: https://github.com/isaacs/sax-js/releases/tag/v1.6.0
- XXE protection in sax: https://github.com/isaacs/sax-js/pull/158

goldserg:
fixes #2218

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -75,14 +75,17 @@ const config = {
  *
  * @param {string} data
  * @param {string=} from
- * @param {number=} maxEntityCount Maximum number of XML entities the parser may
- *   expand, forwarded to sax as its XXE protection limit. When omitted, sax
- *   falls back to its own default (512 in sax >= 1.6.0).
+ * @param {number=} maxEntityCount Maximum number of XML entities the parser is
+ *   allowed to expand before throwing. When omitted, the underlying parser
+ *   falls back to its own default (512 in recent versions).
  * @returns {import('./types.js').XastRoot}
  */
 export const parseSvg = (data, from, maxEntityCount) => {
+  // Build a fresh options object per call: sax mutates the options it receives
+  // (e.g. it stamps a default `maxEntityCount`), so the shared `config` must
+  // not be passed directly.
   const parserConfig =
-    maxEntityCount == null ? config : { ...config, maxEntityCount };
+    maxEntityCount == null ? { ...config } : { ...config, maxEntityCount };
   const sax = SAX.parser(parserConfig.strict, parserConfig);
   /** @type {import('./types.js').XastRoot} */
   const root = { type: 'root', children: [] };
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -75,10 +75,15 @@ const config = {
  *
  * @param {string} data
  * @param {string=} from
+ * @param {number=} maxEntityCount Maximum number of XML entities the parser may
+ *   expand, forwarded to sax as its XXE protection limit. When omitted, sax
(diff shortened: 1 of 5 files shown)


goldserg:
@mozzie @KTibow @SethFalco

goldserg:
Tests on node 24 will fail due to incompatibility with playwright, and raising playwright leads to incompatibility in node 16 and breaks the regression test. Please ask the maintainer to fix the current node 24 tests, as no new commit will pass them.

goldserg:
@mozzie @KTibow @SethFalco will we do something to fix the playwright and nodejs version incompatibility versions?

goldserg:
@mozzie @KTibow @SethFalco

SethFalco:
Hey @goldserg!

Sorry for not acknowledging your PR sooner! All the maintainers are really busy. As you may have seen from our last release, the time we did have available went towards another security advisory that was reported.

We'll try to get through PRs/issues soon. 💪(•ᴗ•💪)

I can confirm already that the issue is valid, and we will accept a pull request to add the options. It was an oversight on my part to not make a way to disable those checks.

Meanwhile, could you please _stop_ pinging maintainers. We get notified of open issues already, pinging us is not going to get your code reviewed any sooner. The bottleneck is availability, and spamming us with notifications just takes up more of the already scarce availability we have! ono

> Tests on node 24 will fail due to incompatibility with playwright, and raising playwright leads to incompatibility in node 16 and breaks the regression test.

PS: If you look at recent PRs including this one, all of our tests are passing on both Node.js v24 and v16 just fine. This doesn't sound correct, could you elaborate? :thinking:

goldserg:
> PS: If you look at recent PRs including this one, all of our tests are passing on both Node.js v24 and v16 just fine. This doesn't sound correct, could you elaborate? 🤔

I rolled back the changes that were breaking the tests, and now they're all passing

PR #2228, opened 2026-07-06
Avoid removing dimensions on nested SVG elements

### LINKED ISSUE #2217: removeDimensions removes `width` and `height` from nested SVG elements
**Describe the bug**
The `removeDimensions` option removes `width` and `height` attributes from nested `<svg>` elements, even when it is unsafe to do so.

**To Reproduce**
Take this example SVG:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10">
	<path fill="pink" d="M0 0h10v10H0Z"/>
	<svg x="1" y="1" width="8" height="8" viewBox="0 0 33 33">
		<path fill="red" d="M1 16a1 1 0 0 0 31 1 1 1 0 0 0-31-1"/>
	</svg>
</svg>
```

One can use [SVGOMG](https://jakearchibald.github.io/svgomg/) to quickly see that toggling the "Prefer viewBox to width/height" shows a significant change in the rendered SVG.

**Expected behavior**
It is expected that this option only affects the outermost `<svg>` element, since the `width` and `height` on nested `<svg>` elements are used to correctly size and position the nested SVG.

**Screenshots**

Original SVG (rendered as PNG):

Image

Compressed SVG (as PNG):

Image

- SVGO 4.0.1
- Node v22.22.1
- OS: MacOS Tahoe 26.2

vrugtehagel:
Fixes https://github.com/svg/svgo/issues/2217

This aligns the behavior with the description in [the docs](https://svgo.dev/docs/plugins/removeDimensions/):

> Removes the width and height attribute from the top-most <svg> element
> if specified, and replaces it with the viewBox attribute if it's missing.

### DIFF
--- a/plugins/removeDimensions.js
+++ b/plugins/removeDimensions.js
@@ -17,8 +17,8 @@ export const description =
 export const fn = () => {
   return {
     element: {
-      enter: (node) => {
-        if (node.name === 'svg') {
+      enter: (node, parentNode) => {
+        if (node.name === 'svg' && parentNode.type === 'root') {
           if (node.attributes.viewBox != null) {
             delete node.attributes.width;
             delete node.attributes.height;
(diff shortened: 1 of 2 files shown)


vrugtehagel:
Never mind, hadn't seen https://github.com/svg/svgo/pull/2225 - closing in favor of that (the change is identical).

Issue #1815, opened 2023-11-01
`--multipass` optimization produces artifacts


nschloe:
When running

```
svgo --multipass a.svg  
```
on the SVG

![a](https://github.com/svg/svgo/assets/181628/9101f5f9-fa62-4a76-8f04-3df3e5f1bc16)

it will produce extra artifacts:

![b](https://github.com/svg/svgo/assets/181628/87f24791-11d3-4579-8aa6-cbca85609ae8)

(See, e.g., the axis labels.)

The problem does not occur without `--multipass`.

This is with svgo 3.0.2, node 21.0.0, on Arch Linux.

SethFalco:
The issue was that the original SVG had some instance of `<use>` that referred to an ID to a node that didn't exist. If the offending node referenced `#a`… but later the `cleanupIds` plugin renames another node ID to `a`, this makes those bad references valid and unintentionally points them to a different node. This causes previously malformed nodes to effectively duplicate an existing node.

See: https://github.com/svg/svgo/pull/1817

### [MRE](https://stackoverflow.com/help/minimal-reproducible-example)

**`minimal.svg`**
```svg
<svg xmlns="http://www.w3.org/2000/svg">
  <defs>
[...]

PR #2238, opened 2026-08-07
fix(cleanupIds): don't skip IDs that were already reassigned

### LINKED ISSUE #2237: `cleanupIds`: minified IDs are not idempotent and can collide
**Describe the bug**

When minifying, `cleanupIds` skips a generated ID if that ID belongs to an element it has already renamed. This has two symptoms:

1. IDs come out non-consecutive and change again on a second pass.
2. With `remove: false`, a generated ID can collide with a kept unreferenced ID, producing duplicate IDs and silently repointing references at the wrong element.

**To Reproduce**

1. Optimize with default params:

```xml
    <defs>
        <path id="b" d="M0 0"/>
        <path id="a" d="M1 1"/>
    </defs>
    <use href="#b"/>
    <use href="#a"/>
</svg>
```

Pass 1 gives `a` and `c`; pass 2 turns `c` into `b`.

2. Optimize with `{ remove: false }`:

```xml
    <defs>
        <path id="x" d="M0 0"/>
        <path id="a" d="M1 1"/>
    </defs>
    <use href="#x"/>
</svg>
```

Both paths come out as `id="a"`, and `<use href="#a">` now resolves to whichever the UA picks.

**Expected behavior**

Pass 1 should produce consecutive IDs (`a`, `b`) that survive a second pass, and should never emit duplicate IDs.

**Desktop**
- SVGO Version: 4.0.2
- Node.js Version: 22+
- OS: Linux / macOS / Windows

**Additional context**

### Root Cause

In `plugins/cleanupIds.js`, the ID generator avoids capturing dangling external references (#1815) with:

```javascript
while (
  isIdPreserved(currentIdString) ||
  (referencesById.has(currentIdString) && nodeById.get(currentIdString) == null)
)
```

Renaming an element runs `nodeById.delete(id)`, so any ID already reassigned during the pass satisfies that test as well and is skipped. The check also doesn't cover unreferenced IDs, which are kept when `remove` is disabled.

### Suggested Fix

Collect external references before any renaming, and guard separately against IDs that will be kept:

```diff
--- a/plugins/cleanupIds.js
+++ b/plugins/cleanupIds.js
@@ -214,6 +214,13 @@ export const fn = (_root, params) => {
          */
         const isIdPreserved = (id) =>
           preserveIds.has(id) || hasStringPrefix(id, preserveIdPrefixes);
+        /** @type {Set<string>} */
+        const externalIds = new Set();
+        for (const id of referencesById.keys()) {
+          if (!nodeById.has(id)) {
+            externalIds.add(id);
+          }
+        }
         /** @type {?number[]} */
         let currentId = null;
         for (const [id, refs] of referencesById) {
@@ -228,8 +235,10 @@ export const fn = (_root, params) => {
                 currentIdString = getIdString(currentId);
               } while (
                 isIdPreserved(currentIdString) ||
-                (referencesById.has(currentIdString) &&
-                  nodeById.get(currentIdString) == null)
+                externalIds.has(currentIdString) ||
+                (!remove &&
+                  nodeById.has(currentIdString) &&
+                  !referencesById.has(currentIdString))
               );
               node.attributes.id = currentIdString;
```

victorvianna:
The generator skips a candidate ID when it looks like a dangling external reference (#1815). Renaming an element deletes its old ID from nodeById, so IDs already reassigned during the pass matched that test too and were skipped, yielding non-consecutive IDs that changed again on a second pass.

Collect external references before any renaming, and guard separately against unreferenced IDs, which are kept when remove is disabled and otherwise led to duplicate IDs and references resolving to the wrong element.

Fixes #2237

### DIFF
--- a/plugins/cleanupIds.js
+++ b/plugins/cleanupIds.js
@@ -214,6 +214,13 @@ export const fn = (_root, params) => {
          */
         const isIdPreserved = (id) =>
           preserveIds.has(id) || hasStringPrefix(id, preserveIdPrefixes);
+        /** @type {Set<string>} */
+        const externalIds = new Set();
+        for (const id of referencesById.keys()) {
+          if (!nodeById.has(id)) {
+            externalIds.add(id);
+          }
+        }
         /** @type {?number[]} */
         let currentId = null;
         for (const [id, refs] of referencesById) {
@@ -228,8 +235,10 @@ export const fn = (_root, params) => {
                 currentIdString = getIdString(currentId);
               } while (
                 isIdPreserved(currentIdString) ||
-                (referencesById.has(currentIdString) &&
-                  nodeById.get(currentIdString) == null)
+                externalIds.has(currentIdString) ||
+                (!remove &&
+                  nodeById.has(currentIdString) &&
+                  !referencesById.has(currentIdString))
               );
               node.attributes.id = currentIdString;
               for (const { element, name } of refs) {
--- a/test/plugins/cleanupIds.27.svg.txt
+++ b/test/plugins/cleanupIds.27.svg.txt
@@ -1,0 +1,27 @@
+Minified IDs must be consecutive and stable in a single pass, even when IDs are
+referenced out of alphabetical order. An ID that has already been reassigned
+must not be mistaken for an external reference.
+
+See: https://github.com/svg/svgo/issues/2237
+
+===
+
+<svg xmlns="http://www.w3.org/2000/svg">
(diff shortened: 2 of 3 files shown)


victorvianna:
@SethFalco PTAL

victorvianna:
@TrySound Could you please approve the workflow runs when you have a chance? Happy to address any feedback.

victorvianna:
Thanks. The failure is unrelated to the CL

PR #2241, opened 2026-08-10
fix(removeUnknownsAndDefaults): don't strip inherited attrs across ID boundaries

victorvianna:
Two independent uselessOverrides bugs, both from comparing an attribute against the parent's full computed style.

1. Inheritance was resolved through the whole ancestor chain, but an element with an `id` (or inside `<defs>`/`<symbol>`) may be instantiated by `<use>` somewhere else entirely, where the ancestors above it do not apply. Stop the walk at that boundary, keeping the boundary element itself since it is cloned along with the subtree.

2. The parent's own presentation attributes were all treated as inheritable. `presentationNonInheritableGroupAttrs` only covers eight of them, so `overflow`, `vector-effect`, `stop-color`, `stop-opacity`, `flood-color`, `clip`, `transform-origin` and friends leaked through and were dropped from children that legitimately carried them — e.g. `overflow` on a nested `<svg>`, which controls its clipping. Filter on `inheritableAttrs` instead.

Both views of the cascade are collected in a single ancestor walk. Each level re-matches the whole stylesheet, so walking twice doubled the cost of the plugin on deep documents.

### DIFF
--- a/lib/style.js
+++ b/lib/style.js
@@ -124,7 +124,7 @@ const parseStyleDeclarations = (css) => {
  * @param {Map<import('./types.js').XastNode, import('./types.js').XastParent>=} parents
  * @returns {import('./types.js').ComputedStyles}
  */
-const computeOwnStyle = (stylesheet, node, parents) => {
+export const computeOwnStyle = (stylesheet, node, parents) => {
   /** @type {import('./types.js').ComputedStyles} */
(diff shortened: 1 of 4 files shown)


victorvianna:
@TrySound Could you please approve the workflow runs when you have a chance? Happy to address any feedback.

PR #2247, opened 2026-08-11
perf: use odiff for regression comparisons

TrySound:
## Summary

- replace Pixelmatch and PNGJS with native `odiff-bin` buffer comparisons
- preserve the existing 3/4-pixel aliasing allowance without decoding PNGs in JavaScript
- retain optional diff artifacts and add bounded odiff/browser cleanup
- add native integration coverage for exact counts, layout differences, artifacts, and thresholds

## Motivation

PR #2246 showed that comparison dominates regression wall time after optimization. ODiff avoids PNGJS RGBA allocations and uses a SIMD-native comparator.

## Testing

- `pnpm test` (17 suites, 497 passed, 3 skipped)
- `pnpm test:types`
- `pnpm lint`
- native odiff integration tests: 6 passed

## Benchmark Status

The current 8 GiB development host is too memory-constrained for a full browser comparison: Chromium is kernel-OOM-killed with both Pixelmatch and odiff, even at one reused page. The 16 GiB regression CI job is therefore the authoritative full-corpus benchmark and classification-parity gate. This PR should only be accepted if all 4,905 matches and 139 expected mismatches remain unchanged and runtime or memory materially improves.


TrySound:
Full-corpus CI result: this experiment does not meet its acceptance criteria.

- odiff comparison phase: approximately 10m36s
- Pixelmatch comparison phase: approximately 10m28s
- expected matches: 4,892 / 4,905 (13 regressions)
- expected mismatches: 134 / 139 (5 regressions)
- full regression job: 24m27s

The classification changes occur in both directions, so a global threshold adjustment cannot restore parity. Chromium rendering remains the dominant cost; replacing the pixel comparator does not materially improve wall time.

PR #2248, opened 2026-08-12
Match CSS attribute selector values before retaining attributes

### LINKED ISSUE #2145: includesAttrSelector: support different equality checks in attribute selectors too
We have a helper function named `includesAttrSelector`. We use this to check if an attribute has been referenced in the stylesheet. If so, the plugin may have different behavior. Usually, by not removing that attribute because it may break styles otherwise.

However, our logic takes a very safe approach, of never deleting attributes if it's been referenced in CSS, even if the operator/check wouldn't match.

For example, suppose an element in an SVG had `preserveAspectRatio="xMidYMid meet"`, with a CSS selector of `[preserveAspectRatio^="y"]`.

This would not match, yet to be on the safe side, we will not remove the attribute. We do not actually check the operator and value, but only the key. Let's improve the logic to actually perform these checks so that we do remove the attribute when it's truly unnecessary.

Neallin-917:
Fixes #2145

`includesAttrSelector` now evaluates CSS attribute selector operators against the current attribute value. The callers in `inlineStyles` and `removeUnknownsAndDefaults` now provide that value, so selectors that cannot match no longer keep a redundant attribute.

Added unit coverage for supported operators and regressions for both affected plugins.

Validation:
- full Jest suite: 494 passed, 3 skipped
- ESLint and Prettier
- type checks, `tsd`, and bundle build
- bundled-browser test

AI assistance: this patch was AI-assisted and the listed validations were run locally.

### DIFF
--- a/lib/style.js
+++ b/lib/style.js
@@ -335,7 +335,43 @@ export const includesAttrSelector = (
         return false;
       }
 
-      return value == null ? true : segment.value === value;
+      if (value == null || segment.action === 'exists') {
+        return true;
+      }
+
+      const selectorValue =
+        segment.ignoreCase === true
+          ? segment.value.toLowerCase()
+          : segment.value;
+      const attributeValue =
+        segment.ignoreCase === true ? value.toLowerCase() : value;
+
+      switch (segment.action) {
+        case 'equals':
+          return attributeValue === selectorValue;
+        case 'hyphen':
+          return (
+            attributeValue === selectorValue ||
+            attributeValue.startsWith(selectorValue + '-')
+          );
+        case 'element':
+          return attributeValue.split(/\s+/).includes(selectorValue);
+        case 'start':
+          return (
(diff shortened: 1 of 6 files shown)


Neallin-917:
Friendly follow-up: this is a focused fix for CSS attribute-selector values being matched before attributes are retained, with a regression fixture covering quoted and escaped values. It was validated with the project's QA suite. I would appreciate a maintainer review when convenient, and I am happy to adjust the approach if needed. Thank you.

Neallin-917:
I rebased this focused change onto current main after confirming the earlier CI failure came from the prior base’s regression-test dependency state, outside this PR’s diff. Local `pnpm run qa` now passes (523 tests passed, 3 skipped). The new GitHub Actions runs are currently marked `action_required` before any jobs start; could a maintainer approve them when convenient? Thank you.

PR #2249, opened 2026-08-12
chore: separate regression rendering and comparison

TrySound:
## Summary

- render original and optimized SVG screenshots to temporary files with at most two reusable Chromium pages
- close Chromium before decoding PNGs and comparing them in a bounded worker-thread pool
- preserve Pixelmatch thresholds, diff behavior, report classifications, and cleanup temporary screenshots on success or failure

## Motivation

The regression comparison currently overlaps Chromium renderer memory with PNGJS decode buffers and runs synchronous Pixelmatch work on the main thread. This separates those phases and introduces a path-based comparison boundary that can support a future odiff experiment.

## Verification

- focused regression tests: 15 passed
- ESLint passed for changed regression files
- Prettier passed for changed regression files
- `git diff --check` passed
- representative Playwright -> filesystem -> worker-thread smoke test passed

## CI experiment

The full corpus could not complete on the 8 GiB VPS because the existing optimization stage was OOM-killed before comparison while processing its five largest fixtures. CI should validate:

- 4,905 expected matches and 139 expected mismatches remain unchanged
- comparison wall time versus current main
- total process-tree peak memory
- temporary screenshot disk usage

The repository-wide unit suite is currently red on main in this checkout due to existing `css-select` plugin failures unrelated to these regression files.

### DIFF
--- a/test/regression/compare-worker.js
+++ b/test/regression/compare-worker.js
@@ -1,0 +1,93 @@
+import fs from 'node:fs/promises';
+import path from 'node:path';
+import { isMainThread, parentPort } from 'node:worker_threads';
+import pixelmatch from 'pixelmatch';
+import { PNG } from 'pngjs';
+
+/**
+ * @typedef CompareOptions
+ * @property {string} name
+ * @property {string} originalPath
+ * @property {string} optimizedPath
+ * @property {string | null} diffPath
+ * @property {typeof fs.rm=} remove
+ */
+
+/**
+ * @param {CompareOptions} options
+ * @returns {Promise<{ name: string, matched: number, width: number }>}
+ */
+export async function compareImages(options) {
+  const {
+    name,
+    originalPath,
+    optimizedPath,
+    diffPath,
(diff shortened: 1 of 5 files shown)


TrySound:
Profiling update:

- previous CI optimization phase: 12m58s
- previous CI render + compare phase: approximately 20m
- local disk-backed Chromium rendering: 25m43s for 5,119 pairs before a large SVG exhausted the VPS
- local Pixelmatch worker phase processed thousands of pairs in seconds before encountering a stale partial-run pair

This identifies Chromium rendering at two-page concurrency as the dominant runtime cost. The latest commit adds explicit render and compare durations to CI output and fixes the TypeScript/JSDoc errors from the first run.

TrySound:
Restored the original render concurrency (`os.cpus().length * 2`) for an apples-to-apples benchmark. Disk staging, strict browser/matcher phase separation, two Pixelmatch workers, and phase timing output remain unchanged.

PR #2251, opened 2026-08-12
ci: set up pnpm without Corepack

TrySound:
## Summary
- replace Corepack setup with `pnpm/action-setup@v4` in every GitHub Actions job
- infer pnpm 10.34.5 from the `packageManager` field
- update contributor setup instructions to install the same pnpm version directly

## Verification
- `pnpm lint`
- confirmed 10 `pnpm/action-setup@v4` steps
- confirmed zero Corepack references in the repository
- `git diff --check`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,10 +31,10 @@ Clone the repository with Git.
 git clone https://github.com/svg/svgo.git
 ```
 
-As this is a Node.js project and uses pnpm for package management, enable Corepack and install the dependencies.
+As this is a Node.js project and uses pnpm for package management, install pnpm and the dependencies.
 
 ```sh
-corepack enable
+npm install --global pnpm@10.34.5
 pnpm install
 ```
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
-      - run: corepack enable
+      - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v6
         with: &setup-node-config
           node-version: 24
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
-      - run: corepack enable
+      - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v6
         with: *setup-node-config
       - run: pnpm install --frozen-lockfile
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -31,10 +31,10 @@ Clone the repository with Git.
 git clone https://github.com/svg/svgo.git
 ```
 
-As this is a Node.js project and uses pnpm for package management, enable Corepack and install the dependencies.
+As this is a Node.js project and uses pnpm for package management, install pnpm and the dependencies.
 
 ```sh
-corepack enable
```
SethFalco, line 37: @TrySound 

I understand using `pnpm/action-setup` in CI and dropping corepack there, but is there any reason to remove `corepack enable` from the contributor's side?

Personally I think corepack is quite nice. It is also helpful when the package manager or packager manager version has has changed, since then contributors will see that right away. (As opposed to finding out another way, such as patches not resolving, or the lockfile having excessive changes because of different lockfile versions.)

KTibow: corepack is being phased out everywhere from what I can tell - Node and package managers don't include or enable it by default, so many contributors are unlikely to even have it installed. I believe at least some package managers already warn you if the configured packageManager is different from what you have.

TrySound: Yep, corepack is becoming less relevant and pnpm itself actually does something similar and loads version specified in packageManager field which should guarantee correct behavior.

SethFalco: Huh… looks like I missed the memo. :thinking: 

Obviously you two know what's up, but just being elaborate here for my own reference, and for potential onlookers. I can see that:

- [Node.js indeed no longer bundles it from Node.js v25.](https://github.com/nodejs/TSC/pull/1697#issuecomment-2737093616)
- pnpm stripped most (or all) references to Corepack out of the docs.
  - https://github.com/pnpm/pnpm.io/pull/862. 
- On npm they document a standard/stable [`package.json#devEngines.packageManager`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#devengines) property which looks it'll be the established way to define the package manager rather than the experimental root `packageManager`. (Both pnpm and even Corepack itself already uses it.)

Yeah, the change makes sense to me now. Fair enough! 💪(•ᴗ•💪)

TrySound: Btw good catch, I missed devEngines intro in pnpm 11. As far as I understand both .packageManager and .devEngines.packageManager fields are supported but the latter is recommended one. Though we still on pnpm 10. To use pnpm 11 we would need to drop node 20. But I would leave it for later.

PR #2254, opened 2026-08-13
perf: render regression SVGs with resvg

TrySound:
## Summary

- install a pinned, checksum-verified resvg 0.48.1 CLI for regression jobs
- replace Chromium regression rasterization with bounded resvg child processes
- migrate regression workflows and establish a reviewed resvg expectation baseline
- retain Playwright Chromium for the browser bundle test

## Performance

On the available host, resvg rendered the full fixture corpus in 104 seconds. Chromium required reduced two-page concurrency for stability and rendered it in 1,407 seconds, so this is an end-to-end improvement rather than a like-for-like microbenchmark.

## Verification

- `pnpm lint`
- `pnpm test` (535 passed, 3 skipped)
- `pnpm test:bundles`
- `pnpm test:regression` (4,977 expected matches, 67 expected mismatches)
- targeted rerendering of all 40 new resvg-specific expectations under three font configurations


TrySound:
Superseded by the rebased Vitest-compatible PR from `perf/resvg-regression-rendering-v2`.

PR #2270, opened 2026-08-24
ci: publish v3 with npm trusted publishing

TrySound:
## Summary

- switch the v3 publish workflow from Yarn's publisher to npm 11.5.1 so npm trusted publishing can use GitHub OIDC
- build the package with Yarn before invoking npm with lifecycle scripts disabled
- always publish this release line under the `v3` dist-tag
- use Node.js 24 for publishing

This fixes the authentication failure observed while publishing v3.3.5 (`YN0033: No authentication configured for request`).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,25 +10,14 @@ permissions:
 jobs:
   publish-npm:
     runs-on: ubuntu-latest
-    env:
-      GH_TOKEN: ${{ github.token }}
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-node@v6
         with:
-          node-version: 20
+          node-version: 24
           registry-url: https://registry.npmjs.org
+      - run: corepack enable
+      - run: npm install --global npm@11.5.1
       - run: yarn install
-      - run: |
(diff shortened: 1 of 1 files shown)


SethFalco:
Thank you for fixing that! I had some trouble with it before as well back with Yarn Berry, I thought it was simply because our Yarn version was too old, so I was hoping migrating to pnpm would've solved it. :/

On the side, I've noticed you've backported to v3, but not to v2. I know it was a bit overkill of me anyway to have backported anything to v2 anyway. :sweat_smile: 

Do you have thoughts on that?

Shall we consider SVGO v2 completely dead and deprecated, or shall we continue to backport the security fixes?

If you think we shouldn't give it security fixes either, I vote we deprecate v2 in npm too, similarly to what we (well, someone in the org!) already did for SVGO v1.

If we do deprecate it, the message could include a link to the [SVGO v2 to v3 Migration Guide](https://svgo.dev/docs/migrations/migration-from-v2-to-v3/).

TrySound:
Already running an agent to backport to v2 as well though yeah we may deprecate it later. Maybe once v5 is ready.

PR #2277, opened 2026-08-24
feat: add browser conditional export

ludofischer:
Hello, for some time, I had been struggling to bundle postcss-plugin-svgo for the browser without overriding the svgo dependency in the bundler config. This change should hopefully allow to run modules that depend on svgo on the browser and on node without overrides. One issue could be if someone wants to force the Node build for a  browser build. In that case with this change they probably would need to change their bundler configuration. So for bundler users it could be seen as either a fix or a breaking change. But I don't think the Node.js functionality would ever work in the browser, so if they were bundling the Node.js versions it was likely a mistake.

Expose the browser bundle from the root `svgo` entry when a resolver enables the `browser` condition, retaining Node-specific `import` and `require` entries under `node` and `default`.

Document resolution, the `svgo/browser` fallback, and TypeScript `customConditions: ["browser"]`.

Add runtime export coverage and a browser-specific typecheck, run by the existing test and typecheck package.json scripts.


TrySound:
I've been thinking to unify entry points and stop bundling browser bundle. It would reduce install size quite a bit. Might do this by the next major bump.

gameroman:
going esm-only would be great 👀

test/exports.js
```diff
@@ -0,0 +1,34 @@
...
+
+const require = createRequire(import.meta.url);
+
+const nodeSvgo = require('svgo');
+assert.equal(typeof nodeSvgo.optimize, 'function');
+assert.equal(typeof nodeSvgo.loadConfig, 'function');
+
+// Add the browser condition to this import so Node resolves the browser branch
+// of SVGO's root conditional export, just as a browser-targeting bundler would.
+const hooks = registerHooks({
```
ludofischer, line 12: Not very convinced about this test, since it is assuming something about how the bundler works internally, rather than testing a real bundler.

ludofischer:
> I've been thinking to unify entry points and stop bundling browser bundle. 

I did not notice both entry points were bundles. I think the change in this PR should be backward compatible, but I understand you might want to wait for a major before changing entry points, with all the build setups in the wild you never know when something might break.

> It would reduce install size quite a bit. 

One thing I have been wondering for some time is if it would be possible for users to use their own CSS minifier instead of CSSO (using an optional dependency somehow?).

PR #2278, opened 2026-08-24
chore: secure actions

ludofischer:
As some months ago there have been a few incidents with attackers taking hijacking GitHub actions, I have applied some of the recommendations recommended by the the [zizmor](https://docs.zizmor.sh/) tool: 

 - Pin actions versions to commit hash
 - Do not persist checkout credentials
 - Disable package manager cache in the publish workflow
- add dependabot cooldown delay
- handle variables more safely

I think the last step is in some cases redundant, as the attacker is unlikely to get control of `github.*` input variables, but I have defined everything as env parameters for consistency, since there should be no disadvantage for doing so.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,5 @@ updates:
     directory: '/'
     schedule:
       interval: monthly
+    cooldown:
+      default-days: 7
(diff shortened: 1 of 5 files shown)


TrySound:
This is great, thanks!

PR #2279, opened 2026-08-24
perf: compile stylesheet selectors once per document

Cherry:
We saw some minor slowdowns when benchmarking our SVG icons workflow with complex SVGs. There are two hot paths recompile static patterns once per element today, and have been fixed here. Pure performance changes: no API change, no config change, byte-identical output.

**1. Stylesheet selectors.** `computeOwnStyle` tests every rule against every element via `matches()`, and css-select parses and compiles the selector string on every call, building a fresh `RegExp` per attribute selector. A document with R rules and E elements pays R × E compilations, multiplied again by tree depth as `computeStyle` walks ancestors. `stylesheet.parents` is already stable, so the compiled matcher is now cached per stylesheet.

**2. `removeAttrs` patterns.** The element/name/value regexes were built inside the element visitor, recompiling every configured pattern for every element, though the patterns come from static config. Moved to plugin init.

## Benchmarks

Synthetic Illustrator-style documents (one class per shape, one stylesheet), Node 24, best of 5:

| Document | before | after | |
| --- | --- | --- | --- |
| 10 rules × 100 shapes | 35.0 ms | 11.9 ms | 2.9× |
| 50 rules × 300 shapes | 383.4 ms | 59.3 ms | 6.5× |
| 100 rules × 600 shapes | 1670.2 ms | 106.8 ms | 15.6× |
| `removeAttrs`, 4 patterns, 600 elements | 7.2 ms | 3.7 ms | 1.9× |

The ratio widens with document size because the old behaviour was quadratic in rules × elements. On a real set of 330 Illustrator-exported icons, a production pipeline (~30 plugins, `multipass`) goes from 835 ms to 552 ms (1.51×); with stock `preset-default`, 1.11×.

<details>
<summary>Benchmark script (repo root, <code>node bench-pr.mjs</code>)</summary>

```js
import { optimize } from './lib/svgo.js';

const makeSvg = (rules, shapes) => {
	const css = Array.from({ length: rules }, (_, i) => `.st${i}{fill:#${(i * 7919 % 0xffffff).toString(16).padStart(6, '0')};}`).join('');
	const body = Array.from({ length: shapes }, (_, i) => `<path class="st${i % rules}" d="M${i % 32} ${Math.floor(i / 32) % 32}h4v4h-4z"/>`).join('');
	return `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32"><style>${css}</style>${body}</svg>`;
};

const time = (label, input, config, iters) => {
	for (let i = 0; i < 3; i++) optimize(input, config);
	let best = Infinity;
	for (let r = 0; r < 5; r++) {
		const t = performance.now();
		for (let i = 0; i < iters; i++) optimize(input, config);
		best = Math.min(best, (performance.now() - t) / iters);
	}
	console.log(`${label.padEnd(40)} ${best.toFixed(1).padStart(8)} ms`);
};

for (const [rules, shapes] of [[10, 100], [50, 300], [100, 600]]) {
	time(`${rules} rules x ${shapes} shapes`, makeSvg(rules, shapes), { multipass: true }, 3);
}
time('removeAttrs, 4 patterns', makeSvg(1, 600).replace(/<style>.*<\/style>/, ''), {
	plugins: [{ name: 'removeAttrs', params: { attrs: ['fill', 'stroke', 'class', 'style'] } }],
}, 20);
```

### DIFF
--- a/lib/style.js
+++ b/lib/style.js
@@ -1,7 +1,7 @@
 import * as csstree from 'css-tree';
 import * as csswhat from 'css-what';
 import { syntax } from 'csso';
-import { matches } from './xast.js';
+import { compileSelector } from './xast.js';
 import { visit } from './util/visit.js';
 import {
   attrsGroups,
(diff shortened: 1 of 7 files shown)


TrySound:
Hey, I would split each improvement to allow testing in isolation

Cherry:
sure can do! Opened https://github.com/svg/svgo/pull/2283 and https://github.com/svg/svgo/pull/2284 instead

PR #2282, opened 2026-08-25
fix: preserve inheritable defaults in referenced subtrees

### LINKED ISSUE #2239: removeUnknownsAndDefaults drops default values that block inheritance into `<use>`
**Describe the bug**

`removeUnknownsAndDefaults` removes a presentation attribute whose value equals the SVG default when no ancestor sets that property. Inside a subtree that `<use>` can instantiate, that attribute is the only thing stopping the property from being inherited at the instantiation site, so removing it changes rendering.

**To Reproduce**

1. Save as `bug.svg`:

   ```svg
     <g id="a"><rect width="10" height="10" fill-opacity="1"/></g>
     <g fill-opacity=".3"><use href="#a"/></g>
   </svg>
   ```

2. Run `npx svgo bug.svg -o -`
3. Output:

   ```svg
   <svg xmlns="http://www.w3.org/2000/svg"><path id="a" d="M0 0h10v10H0z"/><use fill-opacity=".3" href="#a"/></svg>
   ```

**Expected behavior**

`fill-opacity="1"` is retained. In the input the `<use>` instance is fully opaque because the explicit `1` overrides the `.3` above it; in the output it inherits `.3` and renders semi-transparent.

**Desktop**

- SVGO Version 4.0.2
- Node.js Version 22.22.1
- OS: Linux

**Additional context**

Any inheritable property with a default reproduces this — `fill-opacity`, `stroke-opacity`, `stroke-width`, `fill-rule`, `stroke-linejoin`, etc.

The `uselessOverrides` path of this plugin already stops resolving inheritance at the first ancestor with an `id` (or `<defs>`/`<symbol>`) for exactly this reason. The `defaultAttrs` path still walks the full chain. The fix is the same boundary with the opposite polarity: inside a referenceable subtree, keep inheritable attributes that equal their default. Non-inheritable defaults are unaffected.

Neallin-917:
Fixes #2239.

When an element with an `id` is instantiated through `<use>`, an inheritable default inside its subtree can prevent the instance from taking an unintended inherited value from the `<use>` context. Preserve only inheritable defaults in those subtrees; ordinary default attributes remain removable.

Adds a focused regression fixture for `fill-opacity="1"` beneath an `id` target used from a `fill-opacity=".3"` context.

Validation:
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test` (518 passed, 3 skipped)
- `pnpm test:bundles`
- local regression suite completed optimization and comparison with 4,905 expected matches; the local renderer reported 32 `shouldHaveMismatched` differences and no `shouldHaveMatched` differences in unrelated Oxygen/W3C fixtures.

Disclosure: AI assistance was used to investigate and implement this contribution.

### DIFF
--- a/plugins/removeUnknownsAndDefaults.js
+++ b/plugins/removeUnknownsAndDefaults.js
@@ -3,6 +3,7 @@ import {
   attrsGroupsDefaults,
   elems,
   elemsGroups,
+  inheritableAttrs,
   presentationNonInheritableGroupAttrs,
 } from './_collections.js';
 import { detachNodeFromParent } from '../lib/xast.js';
@@ -114,6 +115,28 @@ export const fn = (root, params) => {
   } = params;
   const stylesheet = collectStylesheet(root);
 
+  /**
+   * A referenced subtree may be instantiated by a `use` element, whose
+   * inherited styles must not replace defaults declared inside the subtree.
+   *
+   * @param {import('../lib/types.js').XastNode} node
+   * @returns {boolean}
+   */
+  const hasReferenceableAncestor = (node) => {
+    let ancestor = node;
+    while (ancestor.type === 'element') {
+      if (ancestor.attributes.id != null) {
+        return true;
+      }
+      const parent = stylesheet.parents.get(ancestor);
+      if (parent == null) {
+        return false;
+      }
+      ancestor = parent;
+    }
+    return false;
+  };
+
   return {
     instruction: {
       enter: (node) => {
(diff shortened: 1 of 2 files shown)


Neallin-917:
Friendly follow-up: this focused fix preserves inheritable defaults only for `<use>`-referenced subtrees and includes a regression fixture; its full local validation completed successfully. If the approach no longer fits the project direction, I would appreciate guidance; otherwise, a maintainer review when convenient would be very welcome. Thank you.

PR #2284, opened 2026-08-25
perf: compile removeAttrs patterns once per run

Cherry:
We saw some minor slowdowns when benchmarking our SVG icons workflow with complex SVGs. This is one of two hot paths that recompile static patterns per element; split out from #2279 so each can be reviewed and tested in isolation. The other half is https://github.com/svg/svgo/pull/2283.

Pure performance change: no API change, no config change, byte-identical output.

The element, name and value regexes in `removeAttrs` were built inside the element visitor, recompiling every configured pattern for every element in the document, though the patterns come from static config. Moved to plugin init.

Adds two fixtures covering the paths the hoisting touches: multi-pattern element/name/value matching, and the `elemSeparator` form.

## Benchmarks

600-element document where every element carries all four target attributes, so the name and value matching paths actually run. Node 24, best of 15.

| Case | before | after | |
| --- | --- | --- | --- |
| `removeAttrs`, 600 elems × 4 patterns | 9.32–9.72 ms | 5.65–6.08 ms | ~1.6× |

Range is across three alternating before/after rounds on the same machine.

<details>
<summary>Benchmark script (repo root, <code>node bench-pr.mjs</code>)</summary>

```js
import { optimize } from './lib/svgo.js';

const svg =
      `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32">` +
      Array.from(
              { length: 600 },
              (_, i) =>
                      `<path fill="red" stroke="blue" class="st0" style="opacity:1" d="M${i % 32} ${Math.floor(i / 32) % 32}h4v4h-4z"/>`,
      ).join('') +
      `</svg>`;

const config = {
      plugins: [
              {
                      name: 'removeAttrs',
                      params: { attrs: ['fill', 'stroke', 'class', 'style'] },
              },
      ],
};

for (let i = 0; i < 50; i++) optimize(svg, config);

let best = Infinity;
for (let r = 0; r < 15; r++) {
      const t = performance.now();
      for (let i = 0; i < 40; i++) optimize(svg, config);
      best = Math.min(best, (performance.now() - t) / 40);
}
console.log(`removeAttrs 600 elems x 4 patterns: ${best.toFixed(2)} ms`);
```

</details>

### DIFF
--- a/plugins/removeAttrs.js
+++ b/plugins/removeAttrs.js
@@ -105,27 +105,31 @@ export const fn = (root, params) => {
       : false;
   const attrs = Array.isArray(params.attrs) ? params.attrs : [params.attrs];
 
+  // patterns are static, so compile them once instead of on every element
+  const patterns = attrs.map((attr) => {
+    let pattern = attr;
+    // if no element separators (:), assume it's attribute name, and apply to all elements *regardless of value*
+    if (!pattern.includes(elemSeparator)) {
+      pattern = ['.*', pattern, '.*'].join(elemSeparator);
+      // if only 1 separator, assume it's element and attribute name, and apply regardless of attribute value
+    } else if (pattern.split(elemSeparator).length < 3) {
+      pattern = [pattern, '.*'].join(elemSeparator);
+    }
+
+    // create regexps for element, attribute name, and attribute value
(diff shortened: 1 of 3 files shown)


TrySound:
Thanks!

PR #2288, opened 2026-09-03
fix: prefixIds handles SMIL begin/end offsets and space-optional semicolons

si-kui-a:
Fixes #2207.

## Problem

Two related bugs in the "prefix begin/end attribute value" block of `plugins/prefixIds.js`:

1. The suffix check `val.endsWith('.end') || val.endsWith('.start')` only matched a bare syncbase-value like `b.end`, not one with a clock-value offset like `b.end-0.5s` or `a.begin+0.1s` — both valid SMIL timing syntax. Any offset suffix made the whole check fail, silently leaving the id reference unprefixed and breaking the animation chain it pointed to. (Side note: this repo's only existing fixture for this code path, `prefixIds.10.svg.txt`, only exercises the `.end` half of the check — `.start` isn't valid SMIL syncbase syntax at all per [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/begin), which documents only `id.begin`/`id.end`, so that branch was untested, un-triggerable dead code from the start.)

2. Separately — and this one only surfaced by actually running the issue's own reproduction end to end, not by reading the diff for fix 1 — the sibling split, `attr.split(/\s*;\s+/)` for the semicolon-separated list of timing values, requires at least one whitespace character *after* the semicolon to split at all. Valid SMIL timing lists don't require that whitespace (`0;b.end-0.5s` is valid, and is literally the reporter's own reproduction case). Without a space, the whole multi-value string never gets split, so it would still fail to prefix correctly even with fix 1 alone.

## Fix

1. Replaced the suffix check with a regex, `/^([^.]+)\.(begin|end)([+-].+)?$/`, that captures the id, the `begin`/`end` keyword, and an optional offset separately — the offset is preserved verbatim, only the id gets prefixed.
2. Changed the split delimiter's trailing `\s+` to `\s*`, making whitespace optional on both sides of the semicolon.

## Testing

Added `test/plugins/prefixIds.14.svg.txt`, reproducing the exact scenario from #2207 (including the unspaced semicolon).

Ran the actual suite locally (not just hand-traced): `pnpm vitest run` — 521 passed, 3 skipped (includes multipass idempotence checking built into the fixture-test harness). Confirmed the new fixture genuinely fails against the pre-fix code by temporarily stashing the fix and re-running (it fails as expected, with the exact broken output described in the issue), then restored the fix and reconfirmed passing. `pnpm lint` (eslint + prettier) and `pnpm typecheck` (tsc) both clean.

Co-authored with Claude Code (Anthropic); reviewed by @si-kui-a before submission.

### DIFF
--- a/plugins/prefixIds.js
+++ b/plugins/prefixIds.js
@@ -258,10 +258,20 @@ export const fn = (_root, params, info) => {
             node.attributes[name] != null &&
             node.attributes[name].length !== 0
           ) {
-            const parts = node.attributes[name].split(/\s*;\s+/).map((val) => {
-              if (val.endsWith('.end') || val.endsWith('.start')) {
-                const [id, postfix] = val.split('.');
-                return `${prefixId(prefixGenerator, id)}.${postfix}`;
+            // Semicolon-separated SMIL timing values don't require
+            // whitespace after the semicolon (e.g. "0;b.end-0.5s" is valid),
+            // so whitespace on either side must be optional here, not
+            // required after.
+            const parts = node.attributes[name].split(/\s*;\s*/).map((val) => {
+              // A syncbase-value is "id.begin" or "id.end", optionally
+              // followed by a "+"/"-" clock-value offset (e.g. "b.end-0.5s",
+              // "a.begin+0.1s") -- a plain suffix check misses any value
+              // with an offset, since it doesn't end with just ".begin" or
+              // ".end" anymore.
+              const match = /^([^.]+)\.(begin|end)([+-].+)?$/.exec(val);
+              if (match) {
+                const [, id, postfix, offset = ''] = match;
+                return `${prefixId(prefixGenerator, id)}.${postfix}${offset}`;
               }
               return val;
             });
--- a/test/plugins/prefixIds.14.svg.txt
+++ b/test/plugins/prefixIds.14.svg.txt
@@ -1,0 +1,30 @@
+Prefix IDs should prefix begin/end syncbase references that carry a clock-value
+offset (e.g. "b.end-0.5s"), not just a bare "id.begin"/"id.end" -- and should
+split semicolon-separated values without requiring whitespace after the
+semicolon.
+
+===
+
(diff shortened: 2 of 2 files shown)


SethFalco:
Hey @si-kui-a!

I'll let off these 2 PRs that you've opened, but going forward, could you please refrain from using an LLM to contribute to this project autonomously or communicate to on your behalf.

For better or for worse, none of the currently active maintainers afaik will raise a problem with someone using an LLM, including myself. _However_, to use an LLM to generate code that I'm not convinced you've even read, and then to generate communications is just disrespectful.

Especially for smaller PRs like this, reviewing your own code and doing a brief write up is the least you could do. Collaboration is a core pillar of OSS, so we welcome your contribution, but we expect to be collaborating with a human, perhaps even a human with a robot assistant, but not with the robot itself.

si-kui-a:
Hi @SethFalco !
Thank you very much for your feedback, and I sincerely appreciate your understanding and for merging those two PRs. 

I want to be open about this: 
I relied on an LLM to help phrase my messages because I lack confidence in my English and wanted to avoid misunderstandings. 

However, you are absolutely right—over-relying on it while neglecting code review defeats the purpose of collaboration and goes against the spirit of open source. 

I am deeply sorry if I made you feel disrespected; that was certainly never my intention. 

Moving forward, I will ensure that I personally and rigorously review every line of code. 

I will also do my best to express myself in my own words during communication, using such tools only for basic translation assistance. 

Thank you again for your patience and for pointing this out to me.

codeCraft-Ritik (commented):
Good spot on removing .start since SMIL syncbase timing explicitly uses id.begin and id.end.

As a sanity check: does SVGO also handle event-base values with IDs (e.g., button.click or foo.activate) elsewhere in this plugin, or is the intention here strictly scoped to syncbase references (begin/end)? Keeping it strictly scoped to begin and end makes sense to prevent accidentally prefixing non-ID event names.

